PR #223, opened 2023-04-14
Add function to explore design space of a simulator

luispcunha:
This ports the simple function to explore the design space that @IvanPombo had in the notebook to the API. A limitation is that it only works for simulators that take `input_dir` and `sim_config_filename` as parameters, and only allows for exploring one parameter.


IvanPombo (approved):
Simple and clean. LGTM!
I would just add the function to the __init__ so that we do not need to do 'design.explore_design_space()'.

begaiym-k (approved):
Th code looks very clean!

I am not sure if this is how a template file is created for SWASH normally? Because if it's not there should be a quick note on how to create a template file, like: replace the parameter you want to iterate over with `{{ param_name }}`, but that can be done in another PR, since it's a storytelling part.

luispcunha:
> Th code looks very clean!
> 
> I am not sure if this is how a template file is created for SWASH normally? Because if it's not there should be a quick note on how to create a template file, like: replace the parameter you want to iterate over with `{{ param_name }}`, but that can be done in another PR, since it's a storytelling part.

The notebook is outdated. The only thing I added to the notebook was the correct usage of this function.

PR #224, opened 2023-04-17
Add fluid tank notebook

fabiocruz:
This PR adds a demo notebook with the fluid tank use case, and a hint of how Inductiva API can be used to perform design space parameter exploration/optimization.


fabiocruz:
See [this](https://github.com/inductiva/inductiva/blob/fc-tank-tutorial/demos/tutorials/6_fluid_tank.ipynb) for a proper render of the notebook.

demos/tutorials/6_fluid_tank.ipynb
```diff
@@ -0,0 +1,280 @@
...
+    },
+    "language_info": {
+      "name": "python"
+    }
+  },
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "source": [
+        "# Introduction\n",
```
IvanPombo, line 20: This should provide a more descriptive name of what we will do:
Suggestion: "Optimal Geometry Design of a Fluid Tank"

IvanPombo: I would also not focus so much on optimization, since that points output to find the "best" solution from all feasible solutions.

demos/tutorials/6_fluid_tank.ipynb
```diff
@@ -0,0 +1,280 @@
...
+        "\n",
+        "This tutorial demonstrates how simulations performed via Inductiva API can be used in an optimization task.\n",
```
IvanPombo, line 22: "Inductiva API can effortlessly use simulation to determine the performance of various parameters in the design space." 

"In a previous notebook, we have demonstrated the exploration with a low-level call of the simulator. Here, we explain how the same can be done for a specific scenario and how we can post-process the data to achieve a desirable optimal design."

"We make use of ...."

demos/tutorials/6_fluid_tank.ipynb
```diff
@@ -0,0 +1,280 @@
...
+        "\n",
+        "The fluid container is a cylinder of radius $R_\\mathrm{tank}$ and height $h_\\mathrm{tank}$. The container is initially partially filled. New fluid is injected with velocity $v_\\mathrm{inlet}$ into the tank from a circular inlet at the top of the container with arbitrary coordinates $(x_\\mathrm{inlet}, y_\\mathrm{inlet})$. The fluid exits the tank from a cylindrical outlet at the bottom of the container.\n",
```
IvanPombo, line 28: "Our exploration will focus in the particular case of a cylindrical fluid tank that is defined by a base radius $R_\\mathrm{tank}$ and height $h_\\mathrm{tank}."

We should fix the word tank and use it throughout instead of container.

IvanPombo: Should it be *cylindrical inlet?

demos/tutorials/6_fluid_tank.ipynb
```diff
@@ -0,0 +1,280 @@
...
+        "\n",
+        "In this tutorial, we perform simulations of this scenario for varying $x_\\mathrm{inlet}$, and show that it has an important impact on the average fluid velocity inside the tank. We thus illustrate that Inductiva API can be used for streamlined experimentation with simulations, including in optimization pipelines."
```
IvanPombo, line 30: But isn't our goal an optimal design? Hence, our focus should be immediately maximizing the average fluid velocity and verifying its impact in terms the of the inlet.

Maybe we should point out that it is enough to only vary the x position of the inlet due to the rotational symmetry.

IvanPombo (commented):
Again Fabio this looks really cool!!

I would just like to streamline the storyline a bit more, since sometimes we talk about optimization, others about design exploration. In my view we aren't just exploring the space, we already want to make "decisions" with the data provided by the simulation, so we can also view this as optimal design exploration.

It should be more focused on a single message and not many. This notebook introduces space exploration with post-processing, so I think the storyline is already optimal design, and not just simulations for optimization pipelines or so. WDYT?

demos/tutorials/6_fluid_tank.ipynb
fabiocruz: > Should it be *cylindrical inlet?

Nop. The inlet has no height. It's a circle.

demos/tutorials/6_fluid_tank.ipynb
fabiocruz: > But isn't our goal an optimal design? Hence, our focus should be immediately maximizing the average fluid velocity and verifying its impact in terms the of the inlet.

Not sure what you mean here. Can you elaborate, please?

demos/tutorials/6_fluid_tank.ipynb
IvanPombo: Alright!

Ok, but the outlet is cylindrical?

demos/tutorials/6_fluid_tank.ipynb
fabiocruz: > Maybe we should point out that it is enough to only vary the x position of the inlet due to the rotational symmetry.

I don't want to get into these technical details. I added a sentence to explain that our goal is to illustrate how a single example parameter affects a metric of interest, though. In other words, we chose one parameter to vary but could have chosen a number of others.

demos/tutorials/6_fluid_tank.ipynb
fabiocruz: Yep.

demos/tutorials/6_fluid_tank.ipynb
IvanPombo: This is related with the story we want to tell in the notebook. If it is more about experimental studying something then I think your sentence here is perfect, but if it is more about optimizing/optimal design that the goal is already to maximize the avg. fluid velocity (since it may influence other interesting properties) then the sentence should be reverted, because we are choosing to do it just with the inlet position. We could also do it with the outlet at the same time.

IvanPombo: > I don't want to get into these technical details. I added a sentence to explain that our goal is to illustrate how a single example parameter affects a metric of interest, though. In other words, we chose one parameter to vary but could have chosen a number of others.

Fair enough, my mind just immediately went to why just use the x-position instead of both. So I though of clarifying it.

fabiocruz: Yep, that's a great point. We were missing a sentence that said "this is just an example parameter exploration, we could have picked any other(s)".

demos/tutorials/6_fluid_tank.ipynb
```diff
@@ -0,0 +1,296 @@
...
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "ODULhRDcnDqv"
+      },
+      "source": [
+        "## Introduction\n",
+        "\n",
+        "This tutorial demonstrates how simulations performed via Inductiva API can be used in an optimal design task.\n",
+        "\n",
+        "In a previous notebook, we have demonstrated the exploration with a low-level call of a simulator. Here, we explain how the same can be done for a specific scenario and how to post-process simulation data to determine optimal design conditions.\n",
```
IvanPombo, line 22: *design space exploration

IvanPombo (approved):
LGTM! Minor correction.

demos/tutorials/6_fluid_tank.ipynb
```diff
@@ -0,0 +1,296 @@
...
+    },
+    {
+      "cell_type": "code",
+      "execution_count": 3,
+      "metadata": {
+        "id": "MoYtO0bYOlru"
+      },
+      "outputs": [],
+      "source": [
+        "# Set the API key to be able to access our hardware.\n",
```
begaiym-k, line 84: Can you add this link "https://forms.gle/MkcvWA8BuUFmGVVu6" and mention that the key can be requested here.

demos/tutorials/6_fluid_tank.ipynb
```diff
@@ -0,0 +1,296 @@
...
+        "\n",
+        "The fluid tank is a cylinder of radius $R_\\mathrm{tank}$ and height $h_\\mathrm{tank}$, initially filled with a height $h_\\mathrm{fluid}$ of fluid. New fluid is injected with velocity $v_\\mathrm{inlet}$ into the tank from a circular inlet at the top of the container with arbitrary coordinates $(x_\\mathrm{inlet}, y_\\mathrm{inlet})$. The fluid exits the tank from a cylindrical outlet at the bottom of the container.\n",
```
begaiym-k, line 28: A tank and a container are the same thing right?

begaiym-k (approved):
This looks great! Just left a few comments.

demos/tutorials/6_fluid_tank.ipynb
fabiocruz: Yep. We use these terms interchangeably (mainly to avoid repeating the same word multiple times in a sentence).

PR #226, opened 2023-04-18
Fix file format

IvanPombo:
In this PR, I fix the file format on the explore_design_space function, that was fixed for ".sws" by mistake. As such, we read the file format from the template file and add it at the end of the new input file.

### DIFF
--- a/inductiva/design.py
+++ b/inductiva/design.py
@@ -30,13 +30,14 @@ def explore_design_space(simulator: Simulator,
         extra_sim_kwargs: Extra keyword arguments to pass to the simulator.
     """
     input_dir = pathlib.Path(input_dir)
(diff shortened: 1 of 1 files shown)


inductiva/design.py
```diff
@@ -30,13 +30,14 @@ def explore_design_space(simulator: Simulator,
         extra_sim_kwargs: Extra keyword arguments to pass to the simulator.
     """
     input_dir = pathlib.Path(input_dir)
+    file_format = os.path.splitext(template_filename)[1]
```
luispcunha, line 33: Since `pathlib` is being used here, I think it's better to use it instead of `os.path`. `pathlib.Path(template_filename).suffix` can be used to extract the extension from the path

PR #227, opened 2023-04-19
Fix explore design input dir

luispcunha:


inductiva/design.py
```diff
@@ -47,11 +52,6 @@ def explore_design_space(simulator: Simulator,
             input_file,
         )
 
-        if output_dir is None:
```
IvanPombo, line 50: Why remove this? Does files resolve path takes care of it?

luispcunha: I just moved this to outside the loop.

IvanPombo: Ah sorry! Did not notice that at the start.

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = inductiva
-version = 0.0.8
+version = 0.0.14
```
begaiym-k, line 3: Why did change the version? Are we updating versions every time we push something?

IvanPombo: No, only everytime it goes to PyPi.

Issue #230, opened 2023-04-20
Ability to run simulations Async


IvanPombo:
We have been discussing this aspect for a while, and finally there was a user request for it: "Is it possible to receive the results even if my PC is off...". Therefore, I think it is time to start implementing it.

The priority for now seems to be for the low-level simulation calls and from what I have been discussing with @luispcunha it will take a bit more extra work for the scenarios. Since the low-level calls are the backbone of the scenarios, we can start by implementing an async method for them. Moreover, we can provide the option to run sync and async simulations.

The steps:
- [x] Add personal tokens and simulation identification tokens on the server side. This will allow users to check their job list, statistics and identify the simulations they may want to download to their side.
- [x] Implement methods to send a simulation and to fetch a simulation via simulation identification token. The idea is to split the invoke api functions into smaller steps. There is already various functions to manage tasks that will allow to treat this with simplicity.
- [x] Implement methods for users to check their job list. In the near future, this can also include statistics of time and hardware used, which is good for both parties.
- [x] Maybe, by default, the current invoke_api method should be able to keep running a simulation in case we do not cancel it, but there is some event that causes the collapse of the computer or IDE. For example, a simple event is internet connection drop.
- [ ] Attach a simulation that is still running to check the simulation logs. This goes in hand with the item above or with the idea of just checking in if everything is ok, and detaching again. The idea is very similar to the `screen` method on terminal, but we are providing this feature in the client side.

There might still be intermediate steps in each of the ones above. This issue serves to open the discussion and implementation, since users are already asking for it. Feel free to split each task into more focused issues.

fabiocruz:
Possibly related to #81.

PR #231, opened 2023-04-28
Introducing async simulation calls and fetch data

IvanPombo:
This commit introduces the ability to run async simulations, check the job status and fetch the simulation data in case the job as finished.

This is a first working step into async simulation. Right now, I made available a SPlisHSPlasH low-level async simulation available.

The methodology to run async is as before, but now you obtain a task_id.
```python
sim = sim = inductiva.fluids.simulators.SPlisHSPlasH()

task_id = sim.run_async(input_dir="splish_input",
                      sim_config_filename="input_file.json")
```
To check its status in a python script or Colab: 
```python
inductiva.jobs.check_status(task_id)
```

To fetch the simulation data:
```python
inductiva.jobs.fetch_task(task_id, "splish_out")
```

There are a few things that I would like to keep in mind for the next steps:
- The check status and fetch task are weird functions to just have in a python script. On HPC clusters, users can do this via terminal. In our case, we need to think how we want to provide this functionality, as well as the more general view of all the jobs the user as run.

This ticks one of the boxes in #230.


inductiva/api/methods.py
```diff
@@ -132,6 +132,18 @@ def kill_task(api_instance: TasksApi, task_id: str):
     logging.info("Task terminated.")
 
 
+def check_status(api_instance: TasksApi, task_id: str):
```
fabiocruz, line 135: Should we rename this to `check_task_status` for consistency with the naming used in the rest of the file (and in the API in general)?

fabiocruz: Or, actually, `get_task_status`, since we are getting the status and returning it?

inductiva/api/methods.py
```diff
@@ -348,13 +360,92 @@ def invoke_api(method_name: str,
     return unpack_output(result_list, output_dir, return_type)
 
 
+def invoke_async_api(method_name: str, params, type_annotations: Dict[Any,
+                                                                      Type]):
```
fabiocruz, line 364: All the logic in this function is contained in `invoke_api`. Can we rewrite this to avoid repeating code?

inductiva/api/methods.py
```diff
@@ -379,6 +470,34 @@ def run_simulation(
     return result
 
 
+def run_async_simulation(
```
fabiocruz, line 473: Again, most logic is in `run_simulation`. Can we reuse it? Either move the common logic to a separate function or have a single `run_simulation` method with an `async` parameter?

inductiva/jobs/methods.py
```diff
@@ -0,0 +1,78 @@
+"""Base class for low-level simulators."""
```
fabiocruz, line 1: This seems to be an incorrect description of this file.

inductiva/jobs/methods.py
fabiocruz, line 1: Should the submodule be called `tasks` for consistency?

inductiva/fluids/simulators/splishsplash.py
```diff
@@ -35,3 +35,14 @@ def run(
             device=device,
             input_filename=sim_config_filename,
         )
+
+    def run_async(
```
fabiocruz, line 39: Add a one-liner docstring (at least).

inductiva/jobs/methods.py
```diff
@@ -0,0 +1,78 @@
...
+
+import inductiva
+from inductiva import api
+from inductiva.types import Path
+from inductiva.client import ApiClient, Configuration
+from inductiva.client.apis.tags.tasks_api import TasksApi
+from inductiva.utils.data import (extract_output, unpack_output)
+
+
+def check_status(task_id: str):
```
fabiocruz, line 16: The same comment about `inductiva.api.methods.check_status` applies here.

inductiva/jobs/methods.py
```diff
@@ -0,0 +1,78 @@
...
+
+        status = api.methods.check_status(api_instance=api_instance,
+                                          task_id=task_id)
+
+    return status
+
+
+def fetch_task(task_id: str,
+               output_dir: Optional[Path] = None,
+               return_type: Type = pathlib.Path) -> pathlib.Path:
```
fabiocruz, line 40: `fetch_task_output`, to be more explicit?

inductiva/simulation/simulator.py
```diff
@@ -57,3 +57,29 @@ def run(
...
             log_remote_execution=track_logs,
         )
+
+    def run_async(
+        self,
+        input_dir: types.Path,
+        *_args,
+        **kwargs,
+    ) -> str:
+        """Run the simulation.
```
fabiocruz, line 67: Add "asynchronously".

fabiocruz (changes requested):
Thanks @IvanPombo. I spotted a few places where we are repeating code, hence my request for changes.

inductiva/jobs/methods.py
IvanPombo: I feel that the terminology "jobs" is more consistent with HPC community. WDYT?

fabiocruz: It is, but we are already using the term “tasks” consistently here and on the web API side. I’d stick to it for now for consistency and later update everything if we want to.

inductiva/api/methods.py
IvanPombo: I will keep this as it is for now. There is not enough to a different function. Me and @hpenedones discussed that since the functions `run_simulation` and `run_async_simulation` have different outputs (`pathlib.Path` and string, respectively) we would start with this approach and latter discuss this again.

inductiva/api/methods.py
```diff
@@ -348,13 +360,92 @@ def invoke_api(method_name: str,
...
+
+            upload_input(
+                api_instance=api_instance,
+                original_params=params,
+                task_id=task_id,
+                type_annotations=type_annotations,
+            )
+
+            logging.info("Request submitted.")
+            block_until_status_is(
```
hpenedones, line 430: it's a little counter-intuitive that this async mode has blocking logic inside it (apparently waiting for the task to be started). Is there a strong reason for this? Is it because before that it does not have a task id?

inductiva/api/methods.py
```diff
@@ -327,25 +363,63 @@ def invoke_api(method_name: str,
...
+        1. Transform request params into the params used to
+        validate permission to execute the request.
+        2. Submit task via the "POST task/submit" endpoint.
+            Note: HTTP status code 400 informs the requested method is invalid,
+                and HTTP status code 403 informs that the user is not authorized
+                to post such request.
+        3. If the status returned by the previous HTTP request is
+            "pending-input", ZIP inputs and send them via
+            "POST task/{task_id}/input".
+        4. Return task_id and leaves the simulation running on the server.
```
fabiocruz, line 392: The simulation is not necessarily left running on the server. It can be waiting in the queue until resources become available.

inductiva/tasks/methods.py
```diff
@@ -0,0 +1,48 @@
...
+        api_instance = TasksApi(client)
+
+        status = api.get_task_status(api_instance, task_id)
+
+    return status
+
+
+def fetch_task_output(task_id: str,
+                      output_dir: Optional[Path] = None,
+                      return_type: Type = pathlib.Path) -> pathlib.Path:
```
fabiocruz, line 30: In the cases of the `if-else` block of lines 41-44, the return is not a path, but rather `None`. Adjust the type annotation.

fabiocruz (approved):
LGTM. Left minor comments only this time.

luispcunha (commented):
I took a look at this PR to get up to speed, everything looks fine. I left a small comment. Nice work @IvanPombo!

PR #232, opened 2023-05-03
Update async runs for all simulators

IvanPombo:
With the initial update of async runs done only for SPlisHSPlasH, this serves now to add the async runs to all of our available simulators.

The only thing needed is a function defining the async_run on each simulator that is based on the function in the Simulator class.


inductiva/fluids/simulators/openfoam.py
```diff
@@ -38,3 +38,21 @@ def run(
...
+        
+        Args:
+            sim_config_filename: Name of the simulation configuration file.
+            device: Device in which to run the simulation.
+            other arguments: See the documentation of the base class.
+            """
+
+        return super().run_async(input_dir,
+                                 openfoam_solver=openfoam_solver,
+                                 n_cores=n_cores)
```
fabiocruz, line 58: Doesn't this simulator take an `input_filename`? Its this expected?

inductiva/fluids/simulators/openfoam.py
```diff
@@ -38,3 +38,21 @@ def run(
...
+    def run_async(
+        self,
+        input_dir: types.Path,
+        openfoam_solver: str = "interFoam",
+        n_cores: int = 1,
+    ) -> str:
+        """Run the simulation asynchronously.
+        
+        Args:
+            sim_config_filename: Name of the simulation configuration file.
```
fabiocruz, line 51: This doesn't seem to be in the list of parameters.

inductiva/molecules/simulators/gromacs.py
```diff
@@ -40,3 +40,24 @@ def run(
...
+        input_dir: types.Path,
+        sim_config_filename: str,
+        protein_filename: str,
+        topology_filename: str,
+    ) -> str:
+        """Run the simulation asynchronously.
+        
+        Args:
+            sim_config_filename: Name of the simulation configuration file.
+            device: Device in which to run the simulation.
```
fabiocruz, line 56: This doesn't seem to be in the list of parameters.

fabiocruz (approved):
LGTM. Left minor questions only.

inductiva/fluids/simulators/openfoam.py
IvanPombo: No, openFOAM does not use a single input file to prepare the simulation. In fact it has to do multiple steps to obtain a simulation, which are done on the server side.

PR #238, opened 2023-05-08
Add batch template for multiple files

IvanPombo:
As a requirement for further scenarios we may need to iterate with multiple template files to start a simulation.

In this sense, to not use `replace_params_in_template` multiple times when defining the scenario functions, this function brings this capability directly in the `templates.py`.

### DIFF
--- a/inductiva/utils/templates.py
+++ b/inductiva/utils/templates.py
@@ -1,6 +1,6 @@
 """Utils related to template files."""
 
-from typing import Dict
+from typing import Dict, List
 
 from jinja2 import Environment, FileSystemLoader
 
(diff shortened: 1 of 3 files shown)


inductiva/utils/templates.py
```diff
@@ -17,3 +17,30 @@ def replace_params_in_template(
     template = environment.get_template(template_filename)
     stream = template.stream(**params)
     stream.dump(output_file_path)
+
+
+def batch_replace_params_in_template(
+    templates_dir: str,
+    template_filename_paths: List[str],
+    params: Dict,
+    output_filename_paths: List[str],
```
luispcunha, line 26: Maybe we could add a default for this, wdyt?

IvanPombo: But a default would need to consider the number of `template_filename_paths`. Given that there is an output_filename_path for each template_filename_path.

luispcunha: Sorry. I didn't mean a default value in the function definition, but rather a default value that we assign dynamically in case the function argument is None. Couldn't it be something as simple as removing the `.jinja` part of each filename? Wdyt?

IvanPombo: The template filenames are a bit more explicit, e.g., "0/include/initialConditions_template.openfoam.jinja" representing the fact that it is a template and uses openfoam. 

So I think it is not so simple to provide a default value to output_filename_paths.

fabiocruz (approved):

PR #247, opened 2023-05-10
Adapt WindTunnel scenario to find the new templates directory

IvanPombo:
This commit fits the new templates directory, that is outside of the WindTunnel scenario folder, in the scenario description.


inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -14,7 +14,9 @@
 from inductiva.utils.templates import batch_replace_params_in_template
 from inductiva.utils.files import remove_files_with_tag
 
-OPENFOAM_TEMPLATE_INPUT_DIR = "openfoam_template"
+SCENARIO_TEMPLATE_DIR = os.path.join(os.path.dirname(__file__), "..",
```
luispcunha, line 17: The path to the root of the templates directory should be defined in a dedicated place (e.g. in the templating utils module) for all scenarios to use, otherwise, we will have some duplication. Having to find the templates directory with relative paths everywhere looks a bit error-prone. Wdyt?

IvanPombo: Seems good!

IvanPombo:
Following @luispcunha recommendation, I defined a function to simplify find the path to a package (e.g., the templates folder) in the users local inductiva library. 

However, to achieve this in an appropriate way, I noticed that it is more adequate to have a general folder with all template files, since in the near future we can have many more scenarios in different areas of expertises, like biology or structural mechanics. As it is now, this would require a templates folder inside each new module. 

For this reason, I decided to move the templates folder to the outside. In this way, it serves as a database for all template files. This is still in the scope of this PR, since the initial goal was to appropriately fit the template files for this scenario. Again, the rest of the template files will be moved accordingly, but this will be super easy to do now with this new function.

luispcunha (approved):

PR #251, opened 2023-05-10
Add visualization capabilities to the WindTunnel scenario

IvanPombo:
This PR implements some visualizations capabilities for the WindTunnel scenario. 

Recall, OpenFOAM already provides post-processing features and the respective outputs are provided with the rest of the simulation results. In particular, we are providing: streamLines, cutting planes and force coefficients.

With this in mind, we use `pyvista` to implement visualization capabilities that allow users to understand the flow over the object in the WindTunnel. We point three main visualizations:
- scalar fields over the object surface;
- streamlines with scalar field coloring;
- cutting plane with scalar field coloring.

These visualizations aren't neither specific of the WindTunnel scenario or general enough for all OpenFOAM possible simulations. Due to this, the  implementation was done via the class `OpenFOAMSimulationOutput` analogously to what was done with SPH scenarios. 
In this way, after receiving the outputs with `output = scenario.simulate(...)` users can immediately render their results with `output.render_streamlines(time_step=10, scalars="p")`. 

We may want to refactor this methodology in the near future, but I leave it like this to fit the WindTunnel scenario.

### DIFF
--- a/inductiva/fluids/scenarios/_openfoam_post_processing.py
+++ b/inductiva/fluids/scenarios/_openfoam_post_processing.py
@@ -1,4 +1,10 @@
-"""Post process OpenFOAM simulation outputs."""
+"""Post process OpenFOAM simulation outputs.
+
+Visualization Features:
+- Scalar fields over object.
+- Streamlines with scalar field coloring
+- Cutting plane with scalar field coloring.
+"""
 import os
 import pyvista as pv
 
@@ -58,6 +64,8 @@ class OpenFOAMSimulationOutput:
                            time_step: int,
                            scalars: str = "p"):
         """Render streamlines flowing by object with scalar color map.
+
+        The streamlines are immediately obtained as outputs of 
         
         Args:
             time_step: The time frame for which we want to plot
--- a/inductiva/fluids/scenarios/_openfoam_post_processing.py
+++ b/inductiva/fluids/scenarios/_openfoam_post_processing.py
@@ -12,6 +12,7 @@ from inductiva.types import Path
 
 
 class OpenFOAMSimulationOutput:
+    """Post process WindTunnel simulation outputs."""
 
     def __init__(self, sim_output_path: Path):
         """Initializes a `OpenFOAMSimulationOutput` object.
--- a/inductiva/fluids/scenarios/_openfoam_post_processing.py
+++ b/inductiva/fluids/scenarios/_openfoam_post_processing.py
@@ -1,0 +1,119 @@
+"""Post process OpenFOAM simulation outputs."""
+import os
+import pyvista as pv
+
+from inductiva.types import Path
+
+
+class OpenFOAMSimulationOutput:
+
+    def __init__(self, sim_output_path: Path):
+        """Initializes a `OpenFOAMSimulationOutput` object.
+
+        Args:
+            sim_output_path: Path to simulation output files.
+        
+        Attributes:
+            sim_output_dir: path to the simulation directory
+            object_path: path to the object insert in the WindTunnel
+            post_processing_path: path to the post-processing data.
+        """
+
+        self.sim_output_dir = sim_output_path
+        self.object_path = os.path.join(
+            sim_output_path, "constant",
+            "triSurface", "object.obj")
+        self.post_processing_path = os.path.join(
+            sim_output_path, "postProcessing")
+
+    def render_scalar_field(self,
+                            time_step: int,
+                            scalars: str ="p"):
+        """Render a scalar field over an object from OpenFOAM data.
+        
+        Args:
+            time_step: The time frame for which we want to plot
+                the scalar field over the object.
+            scalars: The scalar field to plot. Meaningfull ones are
+                the pressure "p" and velocity magnitude "U".
+        """
+
+        reading_file = os.path.join(self.sim_output_dir,
+                                    "foam.foam")
+
+        # Create the reading file for the WindTunnel scenarios
+        open(reading_file, "w", encoding="utf-8").close()
+
+        # Initialize reader and define reading time-step
+        reader = pv.POpenFOAMReader(reading_file)
+        reader.set_active_time_value(time_step)
+
+        # Read mesh and mesh properties
+        mesh = reader.read()
+        object_mesh = mesh[1][5:]
+
+        object_mesh.plot(scalars=scalars)
+
+    def render_streamlines(self,
+                           time_step: int,
+                           scalars: str = "p"):
+        """Render streamlines flowing by object with scalar color map.
+        
+        Args:
+            time_step: The time frame for which we want to plot
+                the scalar field over the object.
+            scalars: The scalar field used to color the streamlines.
+                Meaningfull ones are the pressure "p"
+                and velocity magnitude "U".
+        """
+
+        object_mesh = pv.read(self.object_path)
+
+        streamlines_path = os.path.join(
+            self.post_processing_path, "sets", "streamLines",
+            str(time_step))
+
+        streamlines_file = "track0_" + scalars + ".vtk"
+        streamlines_mesh = pv.read(os.path.join(streamlines_path,
+                                           streamlines_file))
+
+        plot = pv.Plotter()
+        plot.add_mesh(streamlines_mesh, scalars=scalars)
+        plot.add_mesh(object_mesh)
+
+        plot.show()
+
+    def render_cutting_plane(self,
+                             time_step: int,
+                             scalars: str = "p"):
+        """Render the cutting plane obtain from the simulation.
+        
+        The cutting plane contains the scalar field properties over
+        a plane over the domain. It is easier to see some of the
+        properties of the flow.
+
+        For now the only cutting plane being computed is orthogonal to
+        the y-axis.
+
+        Args:
+            time_step: The time frame for which we want to plot
+                the scalar field over the object.
+            scalars: The scalar field used to color the streamlines.
+                Meaningfull ones are the pressure "p"
+                and velocity magnitude "U".
+        """
+        object_mesh = pv.read(self.object_path)
+
+        cutting_plane_path = os.path.join(
+            self.post_processing_path, "cuttingPlane",
+            str(time_step))
+
+        cutting_plane_file = scalars + "_yNormal.vtk"
+        cutting_plane_mesh = pv.read(os.path.join(
+            cutting_plane_path, cutting_plane_file))
+
+        plot = pv.Plotter()
+        plot.add_mesh(cutting_plane_mesh, scalars=scalars)
+        plot.add_mesh(object_mesh)
+
+        plot.show()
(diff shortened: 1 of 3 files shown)


sarmento:
This is very cool! 

I just have a few suggestions regarding usability of the WindTunnel scenario itself. For example, it is very difficult for a user to remember that for getting the pressure field you need to set some argument to whatever = "p", especially because this seems very much dependent on the fact that we are using OpenFOAM. Ideally, you would just to 

```
pressure_field = whatever_result.get_pressure_field()
pressure_field.render()
```

Having these higher level methods makes everything more intuitive and easier to remember, and it also decouples the scenario from the idiosyncrasies of each specific simulator.

WDYT @hpenedones ?

hpenedones:
I agree, I think we can improve those calls... scalars="p" is quite cryptic... If this was related with the low-level call to OpenFOAM, then we could argue that the user needs to know that the simulator might use a particular naming convention. But here, on a scenario, we can make the choices and it should be more clear.

IvanPombo:
Update: I have refactored the visualisation of OpenFOAM with the recommendations you gave above.

Now, one can obtain an object `sim_data = inductiva.fluids.scenarios.WindTunnelSimulationOutput(output_path)` 
and render some visualizations via:

```
pressure_field = sim_data.pressure_field
pressure_field.render()
```

I am still working on the cutting plane and streamlines, but I think some feedback on the approach taken is welcome!

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -0,0 +1,86 @@
...
+"""
+from functools import singledispatchmethod
+
+import os
+import pyvista as pv
+
+from inductiva.simulation import Simulator
+from inductiva.fluids.simulators import OpenFOAM
+from inductiva.types import Path
+from inductiva.utils.visualization import Object
```
sarmento, line 19: Object is super super generic. What is an Object? Is this a python Object?

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class WindTunnelSimulationOutput:
```
sarmento, line 22: This class seems to be just a data holder (which is actually fine). I don't see any visualization methods here, are they? Or are they being read from data? 
This is not clear...

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -0,0 +1,86 @@
...
+    @singledispatchmethod
+    def get_pressure_field(self, simulator: Simulator): # pylint: disable=unused-argument
+        return ValueError(
+            f"Simulator not supported for `{self.__class__.__name__}` scenario."
+        )
+
+
+@WindTunnelSimulationOutput.get_object_data.register
+def _(self, simulator: OpenFOAM, time_step: int): # pylint: disable=unused-argument
+    """Get data over object for OpenFOAM."""
```
sarmento, line 62: ?

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -0,0 +1,86 @@
...
+            object_path: path to the object insert in the WindTunnel
+            post_processing_path: path to the post-processing data.
+        """
+
+        self.sim_output_dir = sim_output_path
+        self.object_path = self.get_object
+        self.object_data = self.get_object_data(simulator, time_step)
+        self.pressure_field = self.get_pressure_field(simulator)
+
+    @singledispatchmethod
```
sarmento, line 47: I am confused about why you need this...

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -0,0 +1,86 @@
...
+from inductiva.utils.visualization import Object
+
+
+class WindTunnelSimulationOutput:
+    """Post process WindTunnel simulation outputs."""
+
+    def __init__(self,
+                 sim_output_path: Path,
+                 time_step: int,
+                 simulator: Simulator = OpenFOAM()):
```
sarmento, line 28: It seems you are trying to make things overly general now, and you are getting stuck in very complex class imports and method overloadings...

inductiva/utils/visualization.py
```diff
@@ -424,3 +424,24 @@ def create_color_plot_movie(
         create_movie_from_frames(frames_dir=tmp_dir,
                                  movie_path=movie_path,
                                  fps=movie_fps)
+
+
+class Object():
```
sarmento, line 429: Please avoid this name. Object can mean anything in python programmnig :D

inductiva/utils/visualization.py
```diff
@@ -424,3 +424,24 @@ def create_color_plot_movie(
...
+    """Object class containing data and render properties.
+    
+    This class allows the manipulation of data over objects
+    and the rendering of this data. It serves a niche, but the goal
+    is to integrate this in a more general object file.
+    """
+
+    def __init__(self, object_data, scalar):
+        """Initialize a `Scalar` type object."""
```
sarmento, line 438: scalar? What do you mean, a number like "3.453"?

sarmento (commented):
I think the goal is good, but I am having trouble understanding some naming options and some of the overloading going on here. Are you trying to make things overly generic at this point? I suggest, you make them less generic for now, and more specific for this use case, and then we think about how to develop a generic approach to this. 

There are other visualization cases we want to integrate (e.g. SWASH, FluidTank) and I think we should have a discussion on this.

It's probably ok to open a google document / slides where you start proposing a generalization but I would keep it specific for now.

sarmento (commented):
I think the goal is good, but I am having trouble understanding some naming options and some of the overloading going on here. Are you trying to make things overly generic at this point? I suggest, you make them less generic for now, and more specific for this use case, and then we think about how to develop a generic approach to this. 

There are other visualization cases we want to integrate (e.g. SWASH, FluidTank) and I think we should have a discussion on this.

It's probably ok to open a google document / slides where you start proposing a generalization but I would keep it specific for now.

inductiva/fluids/scenarios/__init__.py
```diff
@@ -2,6 +2,7 @@
 from .dam_break import DamBreak
 from .fluid_block import FluidBlock
 from .wind_tunnel import WindTunnel
+from .wind_tunnel import WindTunnelSimulationOutput
```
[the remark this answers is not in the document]
IvanPombo: My goal here is to able to just use `inductiva.fluids.WindTunnelSimulationOutput`. Maybe the name can be more appropriate or add these features to the WindTunnel itself.

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
IvanPombo: The goal is to be able to overload these functions for different simulators. Check out FluidBlock that already does this for SPlisHSPlasH and DualSPHysics. Before this was done, we had multiple if...else...

sarmento: You have the answer in your question :)

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -40,47 +39,99 @@ def __init__(self,
...
+            A MeshData object that allow to manipulate data over a mesh
+            and to render it.
+        """
+        pressure_field = MeshData(self.object_data, "p")
+
+        return pressure_field
+
+    def get_streamlines(self,
+                        time_step: int,
+                        physical_property: str = "pressure"):
```
begaiym-k: If there are only two options for physical property, maybe it is better to create a list of options here, and let the user insert one.

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -40,47 +39,99 @@ def __init__(self,
...
+
+    def get_flow_plane(self,
+                        time_step: int,
+                        physical_property: str = "pressure"):
+        """Get flow properties in a plane of the domain in WindTunnel."""
+        cutting_plane_path = os.path.join(
+            self.sim_output_dir, "postProcessing",
+            "cuttingPlane", str(time_step))
+
+        if physical_property.lower() == "pressure":
```
begaiym-k: I see a code repetition here, you could wrap this in a separate function.

IvanPombo:
The last commit addresses @sarmento  comments and simplifies the `WindTunnelSimulationOutput`. At this moment, we have a structure that is concise, easy for the user to understand and clean (I think) to extend for other simulators.

I want to note that indeed we need to think more deeply about the structure of the scenarios and visualisations themselves, so that it is easy for users to understand and for developers to integrate in their own scenarios. With this in mind, I will create a doc to organize thoughts and discuss these aspects with the team!

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -40,47 +39,99 @@ def __init__(self,
...
+        return cutting_plane_mesh
+
+
+    def render_flow(self, flow_property_mesh,
+                    physical_property: str = "pressure"):
+        """Render flow property over the object we insert in the WindTunnel."""
+
+        plotter = pv.Plotter()
+        plotter.add_mesh(self.object_data)
+        if physical_property.lower() == "pressure":
```
begaiym-k: Will there be more properties that we will add? Because for now if statements seem fine, but if we have more properties using something like `enumerate` will be better.

begaiym-k (commented):
Great work @IvanPombo! 
I just left some comments and questions.

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
IvanPombo: This was done right when you were reviewing :)

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -40,47 +39,99 @@ def __init__(self,
...
 
-    pressure_field = Object(self.object_data, "p")
+        return object_data
 
-    return pressure_field
+    def get_pressure_field(self): # pylint: disable=unused-argument
+        """Get pressure field over mesh points for a certain time_step.
+        
+        Returns:
+            A MeshData object that allow to manipulate data over a mesh
```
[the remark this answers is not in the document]
IvanPombo: The class MeshData provides you a mesh of the object inside the WindTunnel and physical properties over it. From this you have a render function. But you can also use this data to train ML models :)

inductiva/utils/visualization.py
```diff
@@ -426,22 +426,53 @@ def create_color_plot_movie(
...
-        self.scalar_field = object_data[scalar]
-        self.scalar = scalar
+    def __init__(self, mesh_data, scalar_name: str = None):
+        """Initialize a `MeshData` type object.
+        
+        Args:
+            mesh_data: pyvista mesh (PolyData or Unstructured) which
+                contains all of the simulated outputs over the mesh.
+            scalar_name: string that defines the scalar we want to
+                manipulate and render. This names depend on the mesh_data
```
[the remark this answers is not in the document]
IvanPombo: I have an example on the documentation of the class.

begaiym-k (approved):
LGTM!

PR #257, opened 2023-05-16
Gromacs user interface update to work with new generic gromacs exec

david451lima:
Gromacs now allows for any number of arbitrary user flags


inductiva/molecules/simulators/gromacs.py
```diff
@@ -16,50 +16,40 @@ def api_method_name(self) -> str:
...
+                           user_flags=gromacs_flags)
 
     def run_async(
         self,
         input_dir: types.Path,
-        sim_config_filename: str,
-        protein_filename: str,
-        topology_filename: str,
+        method_name: str,
+        **gromacs_flags: Optional[types.Args],
```
sarmento, line 44: Just a quick note: it seems a bit inconsistent to have one type called Args but then the name of the argument is gromacs_FLAGS. Maybe the type should be called Args (too generic?) but should be called CommandLineFlags.
But I am not sure about this. It just feels weird...

david451lima: Good catch, that was a type, should be str instead of types.Args

inductiva/molecules/simulators/gromacs.py
```diff
@@ -16,50 +16,40 @@ def api_method_name(self) -> str:
 
     def run(
         self,
-        input_dir: types.Path,
-        sim_config_filename: str,
-        protein_filename: str,
-        topology_filename: str,
-        output_dir: Optional[types.Path] = None,
+        input_directory: types.Path,
```
IvanPombo, line 19: Change the docstring on the class, since now you are able to perform any GROMACS comand.

IvanPombo (approved):
LGTM!

PR #269, opened 2023-05-25
Add methods to fetch tasks info

luispcunha:
This PR adds a few functions to fetch task info from the API.

### DIFF
--- a/inductiva/admin/tasks.py
+++ b/inductiva/admin/tasks.py
@@ -1,0 +1,43 @@
+"""Admin functions to manage tasks."""
+from typing import Dict, List, Optional
+
+import inductiva
+from inductiva import api
+from inductiva.client import ApiClient, ApiException
+from inductiva.client.apis.tags.admin_api import AdminApi
+from inductiva.client.model.task_status_code import TaskStatusCode
+
+
+def get_tasks(statuses: Optional[List[str]] = None,
+              usernames: Optional[List[str]] = None,
+              methods: Optional[List[str]] = None,
+              page=1,
+              per_page=10) -> List[Dict]:
+    api_config = api.validate_api_key(inductiva.api_key)
+
+    with ApiClient(api_config) as client:
+        api_instance = AdminApi(client)
+
+        query_params = {
+            "page": page,
+            "per_page": per_page,
+        }
+        if statuses:
+            query_params["status"] = [
+                TaskStatusCode(status) for status in statuses
+            ]
+
+        if usernames:
+            query_params["username"] = usernames
+
+        if methods:
+            query_params["method"] = methods
+
+        try:
+            # Get User Tasks
+            api_response = api_instance.get_tasks(query_params=query_params,)
+
+            return [{**task} for task in api_response.body]
+
+        except ApiException as e:
+            raise e
--- a/inductiva/admin/tasks.py
+++ b/inductiva/admin/tasks.py
@@ -1,4 +1,4 @@
-"""Admin functions to manage tasks."""
+"""Admin functions to query tasks."""
 from typing import Dict, List, Optional
 
 import inductiva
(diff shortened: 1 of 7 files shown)


inductiva/admin/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+              usernames: Optional[List[str]] = None,
+              methods: Optional[List[str]] = None,
+              page=1,
+              per_page=10) -> List[Dict]:
+    api_config = api.validate_api_key(inductiva.api_key)
+
+    with ApiClient(api_config) as client:
+        api_instance = AdminApi(client)
+
+        query_params = {
```
hpenedones, line 21: Potentially this could be a namedtuple, no? 
I would just like to avoid accessing fields using strings as dictionary keys, which is more error prone and the IDE can not provide validation for...

inductiva/tasks/methods.py
```diff
@@ -49,3 +50,45 @@ def fetch_task_output(task_id: str,
...
+
+    with ApiClient(api_config) as client:
+        api_instance = TasksApi(client)
+
+        try:
+            # Get User Tasks
+            api_response = api_instance.get_task(
+                path_params={"task_id": task_id})
+
+            return {**api_response.body}
```
hpenedones, line 66: This is a bit of "advanced" Python, can you leave here a note on what the type is the api_response.body and why you are using the double-asterisk operator?

Typically, it would be used to expand a dictionary while you are inserting the contents into another dict: (see example from https://bas.codes/posts/python-asterisks)

```python
social_media_details = {
    'twitter': 'bascodes'
}

contact_details = {
    'email': 'blog@bascodes.example.com'
}

user_dict = {'username': 'bas', **social_media_details, **contact_details}
# {'email': 'blog@bascodes.example.com', 'twitter': 'bascodes', 'username': 'bas'}
```

But  in your case, if the "body" was already a dict you could return it directly or make a (deep?) copy...
Can you maybe leave a comment clarifying the type of "body" and if your intent is just to create a dict with a copy of its contents?

inductiva/tasks/methods.py
```diff
@@ -49,3 +50,45 @@ def fetch_task_output(task_id: str,
...
+            raise e
+
+
+def get_tasks(status: Optional[str] = None, page=1, per_page=10) -> List[Dict]:
+    api_config = api.validate_api_key(inductiva.api_key)
+
+    with ApiClient(api_config) as client:
+        api_instance = TasksApi(client)
+
+        query_params = {
```
hpenedones, line 78: try to avoid code duplication regarding query_params, perhaps using that namedtuple I suggested... not sure..

inductiva/tasks/methods.py
luispcunha: The response body is a `DynamicSchema` (some data type from the generated client), and I am doing this to convert it to a regular dictionary

inductiva/admin/tasks.py
luispcunha: Because that's the input that the generated client expects... this is how the docs generated by the client do it

luispcunha: I am not sure how the namedtuple here would apply

inductiva/admin/tasks.py
```diff
@@ -0,0 +1,43 @@
+"""Admin functions to manage tasks."""
```
IvanPombo, line 1: manage or query? This only provides the ability to check data and not add? Or in the future we will have that capability here?

inductiva/tasks/methods.py
```diff
@@ -49,3 +50,45 @@ def fetch_task_output(task_id: str,
             result_list = api.download_output(api_instance, task_id, output_dir)
 
             return unpack_output(result_list, output_dir, return_type)
+
+
+def get_task(task_id) -> Dict:
```
IvanPombo, line 55: Let's iterate a bit on the naming to be more specific and to clarify the reason to have just a `get_task_status`. Suggestion: `get_task_info`?

inductiva/tasks/methods.py
```diff
@@ -49,3 +50,45 @@ def fetch_task_output(task_id: str,
...
+            api_response = api_instance.get_task(
+                path_params={"task_id": task_id})
+
+            return {**api_response.body}
+
+        except ApiException as e:
+            raise e
+
+
+def get_tasks(status: Optional[str] = None, page=1, per_page=10) -> List[Dict]:
```
IvanPombo, line 72: Same here: `get_tasks_info`?

IvanPombo (approved):
I would like just to have a one-liner for each new function, but besides that it seems good to go!

PR #270, opened 2023-05-29
Async Sim for WindTunnel Scenario

IvanPombo:
This PR introduces the async runs for the WindTunnel scenario. The implementation follows the one for the low-level, where two separate functions where established for sync and async simulation. The main reason for this is that the inputs and outputs differ slightly in nature.

Moreover, this PR also changes the WindTunnel definition, which came to light during the development of the async run. Now, the object is only established on the simulation part and not on the definition of the scenario. The reason is clear, since the object is the central piece of this scenario we want to change it between runs, even though we keep the domain fixed.


begaiym-k (approved):
It looks fine for now, but I think we need to find a better way to consolidate the code to not have so much repeated code, like in simulate() and simulate_async().

PR #277, opened 2023-06-07
Addition of WindTunnel Executer

IvanPombo:
This PR implements connects the WindTunnel scenario simulation to its own executer on the backend. In practice, we just need to change on the client the name of the api_method to call on the backend, which is changed inside the scenario without users configuration. In this sense, no interference occurs on the user side!

Previously, the backend for OpenFOAM was hard-coded to support simulations of the WindTunnel. However, this implies that no other CFD simulations can be run with OpenFOAM. The next step is to add the low-level implementation of OpenFOAM to run all possible/available commands (like GROMACs) through the use of the Commands&Flags methodology.

A PR on the backend will follow!

Disclaimer: The perfect solution we are working towards is a Pipe of commands run always under the same input folder. However, this will still take some time and practical discussion. Therefore, in the meantime we do not compromise both the low-level and the WindTunnel.

### DIFF
--- a/inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
+++ b/inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
@@ -48,7 +48,7 @@ class WindTunnel(Scenario):
             self.domain = domain
 
     def simulate(self,
-                 simulator: Simulator = OpenFOAM(),
+                 simulator: Simulator = OpenFOAM("windtunnel.openfoam.run_simulation"),
                  output_dir: Optional[Path] = None,
                  object_path: Optional[Path] = None,
                  simulation_time: float = 100,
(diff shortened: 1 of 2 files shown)


hpenedones:
Ok, thanks Ivan.

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -48,7 +48,8 @@ def __init__(self,
             self.domain = domain
 
     def simulate(self,
-                 simulator: Simulator = OpenFOAM(),
+                 simulator: Simulator = OpenFOAM(
+                     "windtunnel.openfoam.run_simulation"),
```
begaiym-k, line 52: Will the user have to provide this string when defining the scenario?

begaiym-k (approved):

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
IvanPombo: Only if he would pass another simulator (like in the case of DualSPH and SplishSplash for FluidBlock). 

Since we only have OpenFOAM and this is not a permanent solution I went with this simple structure.

PR #279, opened 2023-06-13
Protein Solvation Scenario

cristianacarpinteiro:
In this PR I've created the Protein Solvation Scenario. 
This scenario does not inherit from the Scenario class already implemented because more than one simulator.run call are needed to execute this scenario and the input and output directory are the same.  
I've also done it working only for GROMACS because the syntax and order of commands is specific to this simulator. 
Feel free to suggest modifications to comply with the code already implemented, as this is my first PR in this project :)


inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -0,0 +1,166 @@
...
+        Args:
+            protein_pdb: The protein pdb file.
+            working_dir: The working directory where the simulation 
+            will be executed.
+            simulation_time: The simulation time in ns.
+            temperature: The temperature in K.
+            integrator: The integrator to use for the simulation.
+            nsteps_minim: The number of steps to use for the energy minization.
+        """
+        self.nsteps = int(simulation_time * 1e6 / 2)
```
david451lima, line 51: Why this magic number?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -0,0 +1,166 @@
...
+        self.working_dir = self.setup_working_dir(working_dir, protein_pdb)
+        self.gen_config()
+
+        self.simulator.run(input_dir=self.working_dir,
+                           output_directory=self.working_dir,
+                           method_name="pdb2gmx",
+                           f=protein_pdb,
+                           o="protein.gro",
+                           water="tip3p",
+                           user_input="6")
```
david451lima, line 65: I would suggest to comment the command that is being run before the code

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -0,0 +1,166 @@
...
+        shutil.copy(protein_pdb,
+                    os.path.join(working_dir, os.path.basename(protein_pdb)))
+
+        # Copy template files to working_dir
+        for file in os.listdir(self.template_dir):
+            shutil.copy(os.path.join(self.template_dir, file),
+                        os.path.join(working_dir, file))
+
+        # Remove all files that have .jinja in the working_dir
+        remove_files_with_tag(working_dir, ".jinja")
```
david451lima, line 145: Why is this necessary?

david451lima (approved):
LGTM, just some minor questions

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
cristianacarpinteiro: This removes the .jinja files from the working dir. This files are necessary to create the mdp files filled out with the parameters defined by the user.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -0,0 +1,170 @@
...
+        working_dir = resolve_path(working_dir)
+        if not os.path.exists(working_dir):
+            os.makedirs(working_dir)
+
+        # Copy protein pdb to working_dir
+        shutil.copy(protein_pdb,
+                    os.path.join(working_dir, os.path.basename(protein_pdb)))
+
+        # Copy template files to working_dir
+        for file in os.listdir(self.template_dir):
```
luispcunha, line 144: is there any reason you need to iterate on the contents of the directory instead of copying the full directory with shutil.copytree?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -0,0 +1,170 @@
...
+        GROMACS simulator.
+        The three main steps of this scenario are solvation, 
+        energy minimization and simulation. The user can control 
+        the number of steps used to perform the energy minimization step and 
+        the duration, temperature and integrator used to perform 
+        the simulation.
+        """
+        self.template_dir = os.path.join(SCENARIO_TEMPLATE_DIR,
+                                         GROMACS_TEMPLATE_INPUT_DIR)
+        self.simulator = GROMACS()
```
luispcunha, line 31: I think the idea with the scenarios is that at some point they will be supported by multiple simulators, so I don't think simulator should be a property of the class but an argument of the simulate method.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def simulate(
+            self,
+            protein_pdb: str,
+            working_dir: Optional[Path] = None,
+            simulation_time: float = 10,  # ns
+            temperature: float = 300,
```
luispcunha, line 38: Shouldn't temperature and the protein be part of the scenario instead of the simulation? I'm asking because e.g. in the FluidBlock scenario, stuff that is related with the physical properties of the scenario are defined in the constructor of the class (kinematic viscosity, ...) and only properties related with the simulation are specified in the simulate method.

luispcunha (changes requested):
Overall looks good to me. The issues I see are that the scenario is completely tied to gromacs and that this doesn't use the scenario base class. For consistency with the other scenarios, I think you could make some changes to what goes in the constructor and what goes in the simulate method. Related to using the scenario base class, this scenario doesn't "fit the mould", and while we could probably make the base class more general such that it supports simulators that run several commands, I don't think it is worth it to do it now because it will likely change if we implement something related to this in the backend.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
hpenedones: Yes, also the protein can be part of the constructor of the scenario.....

hpenedones:
Thanks Cristiana, can you please open an issue saying that we need to refactor it to integrate well with generic Scenario? And start the discussion there for possible directions on how to implement that: adding your initial thoughts if you have some. :)

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
cristianacarpinteiro: you're right, I will change it to copytree.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
cristianacarpinteiro: I was undecided about what parameters put on the constructor and what to put on the simulate. For example when I am constructing the dataset I define the temperature and simulation steps and then simulate this scenario for different proteins. That's why i put the protein in the simulate. But do you think that the scenario should be specific to the protein? I think that temperature should stay in the simulate method, since you can have the same protein simulated at different temperatures.

luispcunha (approved):
lgtm!

Issue #280, opened 2023-06-14
Need to refactor scenarios.py 


cristianacarpinteiro:
The current implementation of the Scenario class lacks the desired level of abstraction. It is limited to fluid scenarios and does not support cases where multiple commands need to be executed.
Example (taken from the protein solvation scenario):
```  
    simulator.run(input_dir=self.working_dir,
                  output_directory=self.working_dir,
                  method_name="pdb2gmx",
                  f=self.protein_pdb,
                  o="protein.gro",
                  water="tip3p",
                  user_input="6")
    simulator.run(input_dir=self.working_dir,
                  output_directory=self.working_dir,
                  method_name="editconf",
                  f="protein.gro",
                  o="protein_box.gro",
                  c="yes",
                  d="1.0",
                  bt="cubic")
``` 
The current implementation of scenarios only allows for one simulator.run call inside the simulate [method](https://github.com/inductiva/inductiva/blob/f6b3177e49fc57177322d7976bb5e0a7a14b343a/inductiva/scenarios.py#L40).
One possible solution could be to allow for the simulator.run to accept a list of commands. 
For example the code provided could be like this: 
```
    simulator.run(input_dir=self.working_dir,
                  output_directory=self.working_dir, pipeline= [Commands(method='pdb2gmx', f=protein_pdb, ...), Commands(method=editconf, fi=protein.gro, ..)]) 
```
And when we have the pipeline as input we ignore the other arguments. 
But maybe it would be cleaner to have a separate method called simulator.run_pipeline and when we have just one command it pipeline is just a list with one element. Like this: 
```
    simulator.run_pipeline(input_dir=self.working_dir,
                  output_directory=self.working_dir, pipeline= [Commands(method='pdb2gmx', f=protein_pdb, ...), Commands(method=editconf, fi=protein.gro, ..)]) 
```

PR #282, opened 2023-06-15
Refactor scenarios and pipeline implementation 

cristianacarpinteiro:
In this pull request, I am addressing issue #280.
I have opened this PR to receive your feedback on whether this is the best approach to solve the issue, as it will require changes to the inductiva-web-api as well.
I have made the following modifications:

1. Created a Commands class that should be inherited by each simulator. For example, I implemented the GROMACSCommands class, where specific arguments of the GROMACS commands were implemented (method_name).
2. Implemented the simulator.run_pipeline() method, which takes a list of Commands and the working directory as inputs to run the simulation.
3. Changed the api.run() argument parameters to accept a list of arguments instead of a dictionary. This way, we can use the api.run to run a single command or a list of them. 
4.  Added a method called simulate_pipeline to the Base class scenario. 
5.  Refactored the  protein solvation scenario to  inherit from the new scenario base class. 

To complete this change and make it functional, I need to make alterations to the BaseExecuter on the backend side.
I still need to test this solution, but I wanted to get your feedback first.
Thank you :)


cristianacarpinteiro:
It is getting too big, I will stop making changes and wait for your feedback :)

inductiva/scenarios.py
```diff
@@ -2,48 +2,68 @@
...
                 *args,
                 output_dir=output_dir,
                 **kwargs,
             )
+        return output_path
 
+    def simulate_pipeline(
+        self,
+        simulator: Simulator,
+        pipeline: List[Command],
```
hpenedones, line 57: I don't understand this part... The Scenario is something high-level, that should not request from the user any low-level commands from a simulator... But here, the "pipeline" argument is a list of commands.... I think the pipeline of commands needs to be associated with the class simulator... and then a specific scenario (that is aware of which low-level commands need to be run) will tell the simulator to run that pipeline of commands, but the base class "Scenario" wouldn't expose that in a method (i.e. there should not be a simulate_pipeline in Scenario)... No?

inductiva/scenarios.py
```diff
@@ -2,48 +2,68 @@
...
             output_dir = get_timestamped_path(f"{output_dir_prefix}-output")
-
         output_dir = resolve_path(output_dir)
+        return output_dir
 
-        with tempfile.TemporaryDirectory() as input_dir:
-
-            self.gen_aux_files(simulator, input_dir)
-            self.gen_config(simulator, input_dir)
+    def setup_config(self, simulator: Simulator, input_dir: Path):
```
hpenedones, line 25: Also here: you can't ask the user of a Scenario to setup configuration files of simulators... that needs to be hidden....

Unless you want to have some private methods (in Python nothing is really private, but you can use the standard of prepending underscore to the method name to say this an internal method to the class, see : https://google.github.io/styleguide/pyguide.html#3162-naming-conventions) and the derived classes for specific Scenarios would use those methods for some auxiliriary thing, but the user of the scenario never needs to use this setup_config methods or simulate_pipeline... that's too low-level to be exposed in a Scenario...

inductiva/scenarios.py
cristianacarpinteiro: @hpenedones I agree, but I was having difficulties because I wanted to keep the other existing scenarios working and didnt want to change the input arguments to the simulate command. I've altered the code. Is this more inline to what you had in mind?

IvanPombo (commented):
Thanks @cristianacarpinteiro for the great effort you have done and for start developing functionalities to the API out of your comfort zone!!

As a general comment, your implementation seems fine, but there are details that are bit unfriendly, like the Commands class. I think these can be discussed in more detail.

I think right now, and before we start establishing the client side, we need to start thinking how things will work on the backend. This will clear how we can do it on the front-end.
Do you already have any ideas?

inductiva/simulation/command.py
```diff
@@ -0,0 +1,19 @@
+"""Classe for construction of simulator commands."""
+
+
+class Command():
+    """Base class for construction of simulator commands."""
+
+    def __init__(self, **kwargs):
```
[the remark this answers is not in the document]
cristianacarpinteiro: Yes, i did like this so to be a template classe that can be overwritten by other classes for more specific commands, for example the GROMACSCOmmands where the method_name is always necessary. Do you think we can do it any other way?

inductiva/simulation/simulator.py
```diff
@@ -81,5 +87,28 @@ def run_async(
...
+        Args:
+            working_dir: Path to the directory where the pipeline will be 
+            executed.
+            pipeline: List of commands to be executed in the simulation API.
+        """
+        input_dir, output_dir = self._setup_dirs(input_dir, output_dir)
+
+        params = [command.get_args() for command in pipeline]
+        return api.run_simulation(
+            self.api_method_name,
```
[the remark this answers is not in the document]
cristianacarpinteiro: I didn't think about this parameter. But I agree that I need to test this to see where I will get into problems. Because I would assume it is that for gromacs but I need to check.

inductiva/simulation/simulator.py
```diff
@@ -81,5 +87,28 @@ def run_async(
...
+                     input_dir: types.Path,
+                     *_args,
+                     output_dir: Optional[types.Path] = None,
+                     pipeline: List[Command],
+                     track_logs: bool = False) -> pathlib.Path:
+        """Run a pipeline of commands.
+        Args:
+            working_dir: Path to the directory where the pipeline will be 
+            executed.
+            pipeline: List of commands to be executed in the simulation API.
```
cristianacarpinteiro, line 103: I did like this instead @IvanPombo, in response to your comment above

cristianacarpinteiro: I did like this instead @IvanPombo, in response to your comment above

inductiva/simulation/command.py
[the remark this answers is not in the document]
IvanPombo: I think we will only use this when we have several `method_name`'s. Otherwise, I won't need a pipeline.

In general, I will assume that this allows me to run the simulator commands without requiring the "gmx" which is standard for gromacs. For openfoam, I will just assume this are the commands that the binaries allow.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -43,16 +43,16 @@ def __init__(
...
+            output_dir: Optional[Path] = None,
             simulation_time: float = 10,  # ns
             integrator: str = "md",
             nsteps_minim: int = 5000):
         """Simulate the solvation of a protein.
         Args:
-            working_dir: The working directory where the simulation 
-            will be executed.
+            output_dir: The  output directory for the simulation.
             simulation_time: The simulation time in ns. Default is 10 ns.
```
fabiocruz, line 53: No need to state the default value. That's given in the argument list. The same applies to other arguments.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -43,16 +43,16 @@ def __init__(
...
             integrator: The integrator to use for the simulation. Default is md.
+            Other options for integrator include sd, steep, cg, l-bfgs.
```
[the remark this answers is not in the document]
fabiocruz: Also, consider pointing the user to further information about these integrators, e.g. GROMACS documentation?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -97,80 +70,87 @@ def gen_config(self, simulator: Simulator):
...
-                  output_directory=self.working_dir,
-                  method_name="mdrun",
-                  deffnm="solvated_protein",
-                  v="yes")
+    pipeline.append(
+        GROMACSCommand(method_name="mdrun", deffnm="solvated_protein", v="yes"))
+    return pipeline
 
 
 @ProteinSolvation.gen_config.register
```
fabiocruz, line 156: We should consider renaming this to `gen_config_files`, since it can in general be used to generate an arbitrary number of config files. Should we open an issue to address it in a future PR?

inductiva/simulation/command.py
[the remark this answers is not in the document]
fabiocruz: I think `method_name` should be part of the attributes of the base class. A command only exists if it has a name.

inductiva/simulation/command.py
```diff
@@ -0,0 +1,19 @@
...
+        """Command constructor.
+        Args:
+            method_name: The name of the method to be executed in 
+            the simulation API.
```
fabiocruz, line 11: I think you should keep `method_name` in this class, but if you don't please remember to remove this.

fabiocruz: I also think it should be named `name` and not `method_name`. Introducing the `method` prefix is confusing. This is the name of the command, after all...

inductiva/simulation/command.py
```diff
@@ -0,0 +1,19 @@
...
+            **kwargs: Additional keyword arguments to be passed to
+            the simulation API method."""
```
fabiocruz, line 13: This needs an update. I would like to see an example here.

inductiva/simulation/command.py
```diff
@@ -0,0 +1,19 @@
...
+
+        self.kwargs = kwargs
+
+    def get_args(self):
+        """Return the command as a dictionary."""
```
fabiocruz, line 18: I think this is returning the command *flags* as a dictionary. Also, if `self.kwargs` stores the flags, why not call it explicitly `self.flags`?

inductiva/simulation/simulator.py
```diff
@@ -16,10 +16,27 @@ class Simulator(ABC):
     def api_method_name(self) -> str:
         pass
 
+    def _setup_dirs(self, input_dir: types.Path,
```
fabiocruz, line 19: Call this `_setup_io_dirs` or `_setup_input_output_dirs` to make it explicit that the directories being created are for input/output.

inductiva/simulation/simulator.py
```diff
@@ -16,10 +16,27 @@ class Simulator(ABC):
...
+                    output_dir: types.Path) -> pathlib.Path:
+        """Setup the scenario input and output directories."""
+        input_dir = self._setup_input_dir(input_dir)
```
fabiocruz, line 22: This method does not seem to exist.

inductiva/simulation/simulator.py
```diff
@@ -39,29 +57,25 @@ def run(
...
-        if not input_dir.is_dir():
-            raise ValueError(
-                f"The provided path (\"{input_dir}\") is not a directory.")
-
-        if output_dir is None:
-            output_dir = input_dir.with_name(f"{input_dir.name}-output")
-            output_dir = files.get_timestamped_path(output_dir)
+        input_dir, output_dir = self._setup_dirs(input_dir, output_dir)
+        if pipeline is None:
+            params = [kwargs]
```
fabiocruz, line 62: Why wrapping `kwargs` with `[]`?

inductiva/simulation/simulator.py
```diff
@@ -39,29 +57,25 @@ def run(
...
         else:
-            output_dir = files.resolve_path(output_dir)
+            params = [command.get_args() for command in pipeline]
```
fabiocruz, line 64: I think this way of defining `params` can be problematic...

First, the web API will have to be aware whether to extract commands or parameters from the `params` it receives.

Also, there doesn't seem to be a way to run a simulation by setting both a pipeline of commands and extra parameters.

I think this would be much cleaner if `api.run_simulation` accepted an extra parameter `pipeline` **or** if we passed the pipeline as a parameter, with `params["pipeline"] = pipeline`.

Let me know what you think. cc @luispcunha @IvanPombo

fabiocruz, line 64: How is the command name passed? Shouldn't we pass `command.build()` for each command in the pipeline, where `build()` would concatenate the command name and its flags?

fabiocruz (changes requested):
Good work @cristianacarpinteiro, this isn't easy to get clean and you made great progress! I have a few questions, in particular regarding to how we pass the pipeline to the web API.

inductiva/simulation/simulator.py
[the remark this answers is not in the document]
IvanPombo: I'm sorry, my mistake, this parameter is not referring to the simulation at all, just with the simulator we are using. So in principle, you are assuming command contains everything.

inductiva/molecules/simulators/gromacs.py
```diff
@@ -53,3 +53,18 @@ def run_async(
...
+
+    def __init__(self, method_name: str, **gromacs_flags: Optional[str]):
+        """Command constructor.
+
+        Args:
+            method_name: The name of the method to be executed in GROMAS.
+            **gromacs_flags: Additional keyword arguments to be passed 
+            to the simulation API method.
+        """
+        all_args = {"method_name": method_name, **gromacs_flags}
```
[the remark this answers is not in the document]
IvanPombo: In principle, you always have a `method_name` and may or may not have some flags. But the method_name is fixed, since it is calling a specific command in the executer side.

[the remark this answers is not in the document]
IvanPombo: Technically, we didn't need to also add a GromacsCommand on the backend ;)

inductiva/simulation/simulator.py
IvanPombo: I did not get it... `working_dir` is just on the comment, not on the actual arguments, that is why I was confused.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
IvanPombo: Yes!

inductiva/simulation/simulator.py
IvanPombo: This is done in the backend. I think here, we just want to let users define the method and the flags to run.

inductiva/molecules/simulators/gromacs.py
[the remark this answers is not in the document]
cristianacarpinteiro: Ok, so your suggestion is to just keep a commands class, right?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
cristianacarpinteiro: Agreed :)

inductiva/simulation/simulator.py
```diff
@@ -88,7 +88,7 @@ def run_async(
             **kwargs: Additional keyword arguments to be passed to the
                 simulation API method.
         """
-        input_dir, _ = self._setup_dirs(input_dir, None)
+        input_dir, _ = self._setup_io_dirs(input_dir, None)
```
IvanPombo, line 91: In principle this is fine, but notice that you are performing the operation of creating an output_dir name with this.

inductiva/molecules/simulators/gromacs.py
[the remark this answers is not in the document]
IvanPombo: For now, yes! But I think we need to discuss if it is even necessary.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -97,80 +71,87 @@ def gen_config(self, simulator: Simulator):
         )
 
     @singledispatchmethod
-    def run_simulation(self, simulator: Simulator):
+    def gen_pipeline(self, simulator: Simulator):
+        pass
+
+    @singledispatchmethod
+    def gen_aux_files(self, simulator: Simulator, input_dir: str):
```
hpenedones, line 78: Why raising this error here ?

I do think we need to throw this exception if the `simulator` argument in the `run` method is not GROMACS (i.e. someone tries to run this scenario passing in a another simulator)

hpenedones:
Thanks a lot @cristianacarpinteiro, we are getting there! :)

inductiva/molecules/simulators/gromacs.py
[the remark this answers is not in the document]
cristianacarpinteiro: I was having the same doubt! Because it is just a dictionary, but having the command classe will make the user do the proper syntax of the command and define a name for the command and other parameters we may want to pass as arguments.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
fabiocruz: The exception throw can be removed from here, but the function has to exist, such that the registration below (see line 84) works.

For reference, this was introduced when we added `singledispatchmethods` to handle the methods `gen_aux_files`, in order to avoid the solution presented [here](https://github.com/inductiva/inductiva/pull/154). See [this](https://github.com/inductiva/inductiva/pull/203). It's a weird (and ugly!) solution, so it would be great to come up with an alternative...

inductiva/simulation/simulator.py
cristianacarpinteiro: Because our idea was to pass the params to the api as a list of arguments, where each member of the list is a dictionary with parameters for the specific command. Then the API would iterate over the different commands. But this isn't yet implemented in the backend.

inductiva/simulation/simulator.py
cristianacarpinteiro: Yes, but I wanted to avoid creating an additional function because for the setup of the output_dir name also needs the input_dir name.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -43,52 +43,26 @@ def __init__(
     def simulate(
             self,
             simulator: Simulator = GROMACS(),
-            working_dir: Optional[Path] = None,
+            output_dir: Optional[Path] = None,
             simulation_time: float = 10,  # ns
-            integrator: str = "md",
+            integrator: Literal["md", "sd", "bd"] = "md",
```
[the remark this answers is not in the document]
cristianacarpinteiro: You're right and even the pipeline that I am creating is very specific to GROMACS. I wanted to follow the structure of the other scenarios, that's why I did it like this, but should I simplify and do it just for GROMACS?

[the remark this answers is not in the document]
IvanPombo: A future exploration to keep a few things in mind: 
- Does this integrator is specific to GROMACs or other simulators use it under a different name? In case it is not specific, you can follow a structure similar to the link above by providing more descriptive names to the integrators and then connecting them to each specific simulator.
- Is the choice of integrator specific for the scenario we are using here. That is, will the user really need to select a different one or can we keep it fixed for now?

I think this are important to explore and will keep the scenario more agnostic without to much changes. However, one can keep it "specific" for GROMACs for now and when adding a new simulator to molecular dynamics and this scenario change it accordingly.

sarmento:
Just a general comment on this thread: it seems that this is taking too long to converge and is raising a very large number of issues and comments. Next time let's make sure we break changes in smaller incremental pieces that can be merged subsequently. Also, if think get complicated, it may make sense to organize a design meeting to discuss things at high level before start implementing. 

In any case: this is a good piece of work (it's just not organized in a more reasonable way)

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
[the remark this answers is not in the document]
cristianacarpinteiro: I think that other simulators will have a similar parameter to specify the integrator, however I haven't explored enough to say for sure. The choice of the integrator is not specific to this scenario, I just left it because I wanted to explore what is the best one.

cristianacarpinteiro:
Following today's discussion on the API meeting, I've decided to close this PR and open a new one, where I will implement the scenario using a JSON file template to list the commands necessary to run the simulation. I've opened issue #286, where I mention the need to implement a pipeline where the user can specify the commands he wants to run on the API.

PR #292, opened 2023-06-21
Refactor protein solvation scenario

cristianacarpinteiro:
Now, the `ProteinSolvationScenario` class extends the `Scenario` abstract class, providing a specific implementation for simulation of protein solvation in water. 

In this implementation, the API commands to be executed sequentially are defined in a `commands.json` file located in the templates folder. In the backend, this file is copied to the artifacts directory, where the GROMACS executer will process them in order, without requiring the repetitive download of files from the input folder.

To trigger the execution of the command list, `method_name` parameter should be `None` when calling `gromacs.run(input_dir, method_name=None)`. This assumes that you have a `commands.json` file.

Alternatively, you can run a specific command, such as `pdb2gmx`, by providing the `method_name` parameter as `"pdb2gmx"`, along with any additional `gromacs_flags`, like `gromacs.run(input_dir, method_name="pdb2gmx", **gromacs_flags) `.

It's important to note that this solution is currently designed for GROMACS only, as the modifications have been made solely to the GROMACS executer and not the base executer. Please refer to the associated [pull request](https://github.com/inductiva/inductiva-web-api/pull/146)  in the `inductiva-web-api` repository for further details and changes made.

### DIFF
--- a/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
@@ -77,6 +77,17 @@ class ProteinSolvation(Scenario):
             f"Simulator not supported for `{self.__class__.__name__}` scenario."
         )
 
+    @singledispatchmethod
+    @classmethod
+    def get_config_filename(cls, simulator: Simulator):  # pylint: disable=unused-argument
+        raise ValueError(
+            f"Simulator not supported for `{cls.__name__}` scenario.")
+
+
+@ProteinSolvation.get_config_filename.register
+def _(self, simulator: GROMACS):  # pylint: disable=unused-argument
+    pass
+
 
 @ProteinSolvation.gen_aux_files.register
 def _(self, simulator: GROMACS, input_dir):  # pylint: disable=unused-argument
--- a/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
@@ -79,10 +79,10 @@ class ProteinSolvation(Scenario):
 
 
 @ProteinSolvation.gen_aux_files.register
-def _(self, simulator: GROMACS, input_dir, protein_pdb):  # pylint: disable=unused-argument
+def _(self, simulator: GROMACS, input_dir):  # pylint: disable=unused-argument
     """Setup the working directory for the simulation."""
     # Copy protein pdb to working_dir
-    shutil.copy(protein_pdb, os.path.join(input_dir, "protein.pdb"))
+    shutil.copy(self.protein_pdb, os.path.join(input_dir, "protein.pdb"))
     # Copy template files to working_dir
     shutil.copytree(os.path.join(self.template_dir),
                     os.path.join(input_dir),
@@ -92,7 +92,7 @@ def _(self, simulator: GROMACS, input_dir, protein_pdb):  # pylint: disable=unus
 
 
 @ProteinSolvation.gen_config.register
-def _(self, simulator: GROMACS):  # pylint: disable=unused-argument
+def _(self, simulator: GROMACS, input_dir):  # pylint: disable=unused-argument
     """Generate the mdp configuration files for the simulation."""
     batch_replace_params_in_template(
         templates_dir=self.template_dir,
@@ -107,6 +107,6 @@ def _(self, simulator: GROMACS):  # pylint: disable=unused-argument
             "nsteps_minim": self.nsteps_minim,
         },
         output_filename_paths=[
-            os.path.join(self.working_dir, "simulation.mdp"),
-            os.path.join(self.working_dir, "energy_minimization.mdp"),
+            os.path.join(input_dir, "simulation.mdp"),
+            os.path.join(input_dir, "energy_minimization.mdp"),
         ])
--- a/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
@@ -49,7 +49,8 @@ class ProteinSolvation(Scenario):
             nsteps_minim: int = 5000):
         """Simulate the solvation of a protein.
         Args:
-            output_dir: The  output directory for the simulation.
+            output_dir: The  output directory to save the results of the
+            simulation. 
             simulation_time: The simulation time in ns. 
             integrator: The integrator to use for the simulation. 
             For more information about the integrator used in the simulation, 
@@ -79,8 +80,7 @@ class ProteinSolvation(Scenario):
 
 @ProteinSolvation.gen_aux_files.register
 def _(self, simulator: GROMACS, input_dir, protein_pdb):  # pylint: disable=unused-argument
-    """Setup the working directory for the simulation. 
-    For this scenario, the input and output directories are the same."""
+    """Setup the working directory for the simulation."""
     # Copy protein pdb to working_dir
     shutil.copy(protein_pdb, os.path.join(input_dir, "protein.pdb"))
     # Copy template files to working_dir
--- a/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
@@ -61,7 +61,8 @@ class ProteinSolvation(Scenario):
                 - "bd" (Brownian Dynamics): Euler integrator for Brownian or 
                 position Langevin dynamics. 
                 
-            For more details on the integrators, refer to the GROMACS documentation at
+            For more details on the integrators, refer to the GROMACS 
+            documentation at 
             https://manual.gromacs.org/current/user-guide/mdp-options.html.
 
             nsteps_minim: Number of steps for energy minimization.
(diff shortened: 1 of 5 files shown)


inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -43,52 +43,27 @@ def __init__(
     def simulate(
             self,
             simulator: Simulator = GROMACS(),
-            working_dir: Optional[Path] = None,
+            output_dir: Optional[Path] = None,
             simulation_time: float = 10,  # ns
-            integrator: str = "md",
+            integrator: Literal["md", "sd", "bd"] = "md",
```
IvanPombo, line 48: I am still in favour of this being a bit more descriptive, since md/sd/bd don't bring anything to mind, unless you are a gromacs expert.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -43,52 +43,27 @@ def __init__(
...
-        shutil.copy(protein_pdb,
-                    os.path.join(working_dir, os.path.basename(protein_pdb)))
-        # Copy template files to working_dir
-        shutil.copytree(os.path.join(self.template_dir),
-                        os.path.join(working_dir),
-                        dirs_exist_ok=True)
-        # Remove all files that have .jinja in the working_dir
-        remove_files_with_tag(working_dir, ".jinja")
-        return working_dir
+        return super().simulate(simulator, output_dir)
```
IvanPombo, line 66: Can we decide the type of hardware to run? Or GROMACs will consume everything it has available?

Just questioning, because David showed that not always is appropriate to use a lot of cores.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -97,80 +72,23 @@ def gen_config(self, simulator: Simulator):
...
-                  o="solvated_protein.tpr")
-    simulator.run(input_dir=self.working_dir,
-                  output_directory=self.working_dir,
-                  method_name="mdrun",
-                  deffnm="solvated_protein",
-                  v="yes")
+@ProteinSolvation.gen_aux_files.register
+def _(self, simulator: GROMACS, input_dir, protein_pdb):  # pylint: disable=unused-argument
+    """Setup the working directory for the simulation."""
+    # Copy protein pdb to working_dir
```
IvanPombo, line 84: I think these comments are unnecessary, it is clear what you are doing. What you can do is explain the reason to establish a unique name to the `protein.pdb`.

inductiva/simulation/simulator.py
```diff
@@ -16,6 +15,22 @@ class Simulator(ABC):
     def api_method_name(self) -> str:
         pass
 
+    def _setup_io_dirs(self, input_dir: types.Path,
```
IvanPombo, line 18: Maybe to unify with the scenarios we can also split this into two functions.

IvanPombo (approved):
This indeed looks great and now one can create scenarios much simpler. To be honest here, it evens seems that the scenario is defined by the commands.json which is cool!

The only comments I left are suggestions regarding the overall scenario! I will check the backend, but here it LGTM, great work @cristianacarpinteiro!

inductiva/scenarios.py
```diff
@@ -13,39 +12,56 @@
...
+    def gen_aux_files(self, simulator: Simulator, input_dir: Path):
+        """Generate the auxiliary files for the scenario. To be implemented
+        in subclasses."""
+        pass
+
+    @abstractmethod
+    def gen_config(self, simulator: Simulator, input_dir: Path):
+        """Generate the configuration files for the scenario. To be implemented
+        in subclasses."""
+        pass
```
fabiocruz, line 45: The docstrings here should be one-liners. Maybe remove the bit "To be implemented in subclasses."?

fabiocruz (changes requested):
Thank you @cristianacarpinteiro!

I have one concern only, related to keeping `method_name` and `gromacs_flags` as part of the arguments to the `run()` method of the `GROMACS` simulator.

I think it would be much cleaner if this method received a `commands` argument holding either:
- a list with the commands to be executed;
- the name of a json file with a list of commands to be executed.

Such list (in either Python code or the json file) could be something similar to what you envisage, e.g.:

```python
commands = [
    {"command_name": "first_command", "flags": {...}},
    {"command_name": "second_command", "flags": {...}},
]
```

If you want to run one command only, then the list should contain one element only. Does this make sense?

inductiva/templates/protein_solvation/gromacs/commands.json
fabiocruz, line 1: I think the key holding the name of the command should be "command_name" and not "method_name". Otherwise, this would be a list of methods, right?

IvanPombo:
> Thank you @cristianacarpinteiro!
> 
> I have one concern only, related to keeping `method_name` and `gromacs_flags` as part of the arguments to the `run()` method of the `GROMACS` simulator.
> 
> I think it would be much cleaner if this method received a `commands` argument holding either:
> 
> * a list with the commands to be executed;
> * the name of a json file with a list of commands to be executed.
> 
> Such list (in either Python code or the json file) could be something similar to what you envisage, e.g.:
> 
> ```python
> commands = [
>     {"command_name": "first_command", "flags": {...}},
>     {"command_name": "second_command", "flags": {...}},
> ]
> ```
> 
> If you want to run one command only, then the list should contain one element only. Does this make sense?

Nicely thought @fabiocruz! Maybe the python code fits best, and instead of loading the json file in the backend we would load the json file in the scenario.

inductiva/templates/protein_solvation/gromacs/commands.json
cristianacarpinteiro: Yes, I just kept the method_name because it is like that everywhere! But I can change it at least here.

inductiva/templates/protein_solvation/gromacs/commands.json
```diff
@@ -0,0 +1,9 @@
+[{"method_name": "pdb2gmx", "f": "protein.pdb", "o": "protein.gro", "water": "tip3p", "user_input": "6"},
```
hpenedones, line 1: I am guessing that @sarmento would suggest that here we keep things even more "barebones" without using any of our abstractions for commands, and just list the raw command lines, such that it is a trivial "copy-paste" from GROMACS tutorials, etc and there is nothing Inductiva-specific here. 

Note: this might seem to go against the direction we were following to support security validation of the commands (against code injection for example), but maybe we can still achieve that to some extent while living the json file very clean and simple?

cristianacarpinteiro: I agree that this goes a bit against what has been already implemented by David to ensure that there isn't code injection.

cristianacarpinteiro: But I agree that that solution is more user friendly.

IvanPombo: I think here the focus is the scenarios and thus users won't need to copy-paste anything. For developing new scenarios, this does not seem like a large overhead to implement and the .json file seems clean enough to understand. 

In my view, this implementation even simplifies further the executers implementation and can be made more generic for other simulators. Hence, the deployment of new simulators will be super easy!

The "barebone" implementation would require a new refactor on the backend to accomodate this and the security validation, which seems that would be going a few steps backward. Notice that, knowledgable users could change the commands.json file if they want and inject code on the executers, unless we save it on the backend. The latter will be an extra thing to add on the backend for each new scenario.

Moreover, this allows for a sequence of commands to be run on a low-level simulation (fixing also the concept of Pipe we were discussing on the sync) and the security of those commands is still validated by us before running them.

cristianacarpinteiro: I agree with @IvanPombo. My suggestion is to implement Fabio's changes, as this solution already solves the issue of executing the list of commands and unblocks me. Afterwards, we can discuss a more "barebones" solution, which would require bigger modifications to the backend.

cristianacarpinteiro:
> Thank you @cristianacarpinteiro!
> 
> I have one concern only, related to keeping `method_name` and `gromacs_flags` as part of the arguments to the `run()` method of the `GROMACS` simulator.
> 
> I think it would be much cleaner if this method received a `commands` argument holding either:
> 
> * a list with the commands to be executed;
> * the name of a json file with a list of commands to be executed.
> 
> Such list (in either Python code or the json file) could be something similar to what you envisage, e.g.:
> 
> ```python
> commands = [
>     {"command_name": "first_command", "flags": {...}},
>     {"command_name": "second_command", "flags": {...}},
> ]
> ```
> 
> If you want to run one command only, then the list should contain one element only. Does this make sense?

Yes, this would simplify the implementation. But then, we would be changing the low-level call? Now, for calling a gromacs method, you do `gromacs.run(method_name="pdb2gmx", o="protein.pdb")`. With this it would be ´gromacs.run(commands= {"command_name": "pdb2gmx", "flags": {"o":"protein.pdb"}})´, which isn't as clean, in my opinion.   However, I think that the user will always have to run more than 1 gromacs command to perform a simulation. So, it is okay for me.

fabiocruz:
> Yes, this would simplify the implementation. But then, we would be changing the low-level call? Now, for calling a gromacs method, you do `gromacs.run(method_name="pdb2gmx", o="protein.pdb")`. With this it would be ´gromacs.run(commands= {"command_name": "pdb2gmx", "flags": {"o":"protein.pdb"}})´, which isn't as clean, in my opinion. However, I think that the user will always have to run more than 1 gromacs command to perform a simulation. So, it is okay for me.

Yep, I think it should be fine to update the low-level call. In the future we can modify this to accept different "prescriptions" of `commands` e.g.  a list of commands in full string format (e.g. `pdb2gmx -o protein.pdb` and things of the sort).

fabiocruz (approved):
LGTM!

PR #305, opened 2023-06-26
Water box scenario

cristianacarpinteiro:
In this pull request, I am introducing a new Molecular Dynamics scenario that involves simulating a simple box filled with water. The purpose of this scenario is to facilitate the construction of the dataset used in this [article](https://arxiv.org/abs/2112.03383). Unlike the protein solvation scenario, I needed to make changes to the `commands.json` file to define the` box_size`. As a result, there are differences in the simulate method, while the remaining code remains largely the same.
The simulation parameters in the mdp files were altered to comply with the description of the simulation in the article: 
image


inductiva/molecules/scenarios/water_box/water_box.py
```diff
@@ -0,0 +1,145 @@
+"""Water box scenario."""
```
IvanPombo, line 1: I think here we should be more picky, since we have fluids scenarios as well.

Suggestion: "Water molecules box scenario".

IvanPombo: Maybe on the rest of the variables `water_box` this should also be added. Notice that, we have on the templates "templates/water_box/gromacs". If we have a CFD simulation with openfoam, for example, should it also go here, or this folder is just for MD?

inductiva/molecules/scenarios/water_box/water_box.py
```diff
@@ -0,0 +1,145 @@
...
+from inductiva.types import Path
+from inductiva.molecules.simulators import GROMACS
+from inductiva.simulation import Simulator
+from inductiva.utils.templates import (TEMPLATES_PATH,
+                                       batch_replace_params_in_template,
+                                       replace_params_in_template)
+from inductiva.scenarios import Scenario
+from inductiva.utils.files import remove_files_with_tag
+
+SCENARIO_TEMPLATE_DIR = os.path.join(TEMPLATES_PATH, "water_box")
```
IvanPombo, line 17: Consider the above change here.

inductiva/molecules/scenarios/water_box/water_box.py
```diff
@@ -0,0 +1,145 @@
...
+GROMACS_TEMPLATE_INPUT_DIR = "gromacs"
+
+
+class WaterBox(Scenario):
```
IvanPombo, line 21: Here.

inductiva/molecules/scenarios/water_box/water_box.py
```diff
@@ -0,0 +1,145 @@
...
+    def __init__(
+        self,
+        temperature: float = 300,
+        box_size: float = 3.0,
+    ):
+        """The scenario involves simulating a box filled with water molecules.
+        The simulation consists of three main steps: energy minimization, water 
+        molecule position decorrelation, and molecular dynamics simulation. 
+        Currently, this scenario exclusively supports the GROMACS simulator.
+        By default, for a given box size the water molecules are always
```
IvanPombo, line 33: Is there a way to change this or is it to complicated? In my view, this is the first source of general randomness.

IvanPombo: Ok, I see this right below. But maybe it should be phrased differently, to notify that, we establish the initial positions uniformly and thereafter apply a decorrelation step to introduce randomness.

inductiva/molecules/scenarios/water_box/water_box.py
```diff
@@ -0,0 +1,145 @@
...
+            https://manual.gromacs.org/current/user-guide/mdp-options.html.
+
+            nsteps_minim: Number of steps for energy minimization.
+        """
+        self.nsteps = int(
+            simulation_time * 1e6 / 2
+        )  # convert to fs and divide by the time step of the simulation (2 fs)
+        self.integrator = integrator
+        self.nsteps_minim = nsteps_minim
+        replace_params_in_template(
```
IvanPombo, line 80: This part should go in the gen_config, no?

inductiva/molecules/scenarios/water_box/water_box.py
```diff
@@ -0,0 +1,145 @@
...
+            self.template_dir, "commands.json.jinja",
+            {"box_size": self.box_size},
+            os.path.join(self.template_dir, "commands.json"))
+        commands = self.read_commands_from_file()
+        return super().simulate(simulator, output_dir, commands=commands)
+
+    def read_commands_from_file(self):
```
IvanPombo, line 87: Now that we see this will be used over most scenarios, we can set it in the scenario class, wdyt?

IvanPombo (approved):
To move forward, LGTM! But this scenario points a few things that we can discuss further:
1. It seems that the only thing that changes are the template files. In principle, for the `ProteinSolvation` we can also have a box, right? Thus, the above scenario can be made more general and then re-used for the childs by just changing the template files and commands.

2. While I see the goal of having this scenario to replicate the dataset of the paper, I still wonder: How can we insert a protein in the box and everything works nicely? Does it make sense? In this way, we could have both the scenarios in a single one. If this does not work, it is completely ok. But our (me included) next steps need to be documenting the principles behind each scenario :)

cristianacarpinteiro:
> To move forward, LGTM! But this scenario points a few things that we can discuss further:
> 
> 1. It seems that the only thing that changes are the template files. In principle, for the `ProteinSolvation` we can also have a box, right? Thus, the above scenario can be made more general and then re-used for the childs by just changing the template files and commands.
> 2. While I see the goal of having this scenario to replicate the dataset of the paper, I still wonder: How can we insert a protein in the box and everything works nicely? Does it make sense? In this way, we could have both the scenarios in a single one. If this does not work, it is completely ok. But our (me included) next steps need to be documenting the principles behind each scenario :)

I understand your points and I considered them when writing the code! However, the commands for this scenario are very different from the Protein Solvation scenario. For example, when you have a protein inside the box you need to insert ions to balance the charges of the system and here you don't need it. Also, the parameters files are different because I made them to fit the description of the dataset of the article. So, I think it is more readable to have it as a different scenario!

inductiva/molecules/scenarios/water_box/water_box.py
cristianacarpinteiro: good point, should we had another layer of folders? Like, fluids, MD? Maybe it would be more organised.

inductiva/molecules/scenarios/water_box/water_box.py
cristianacarpinteiro: so water_molecules_box?

inductiva/molecules/scenarios/water_box/water_box.py
cristianacarpinteiro: maybe I can make it more clear! What I meant to say is that for two runs for a box of the same size, the water molecules will be positioned in the same places everytime, this is the default of GROMACS. So to construct the dataset and have multiple experiments for the same box size starting with different positions I have to introduce some randomness in the initial positions.

inductiva/molecules/scenarios/water_box/water_box.py
cristianacarpinteiro: since the commands go directly in the super().simulate(commands=commands) I need to read it before the call. Because the gen_config is called inside super.simulate().

inductiva/molecules/scenarios/water_box/water_box.py
cristianacarpinteiro: the read_commands_from_file right? yes, I can move it :)

inductiva/molecules/scenarios/water_box/water_box.py
IvanPombo: water_md_box?

inductiva/molecules/scenarios/water_box/water_box.py
IvanPombo: Keep it simple, just change the name.

inductiva/molecules/scenarios/water_box/water_box.py
IvanPombo: Should the scenario be random, or should the user be able to pass a `seed` to randomize? By this I mean, that running two times the scenario I obtain different results. 

Here, it seems that this will be very focused on the dataset, instead of the general scenarios context. Wdyt, is it possible to have a seed for this?

inductiva/molecules/scenarios/water_box/water_box.py
IvanPombo: Seems good! Maybe later we can tackle this if it repeats in other scenarios.

inductiva/molecules/scenarios/water_box/water_box.py
cristianacarpinteiro: The trajectories from the same starting point will always end up being different, because the simulation is inherently stochastic, due to temperature effects. And I agree that is very focused on the dataset, but the objective was indeed that: I want to repeat the dataset from the article to then be able to create a blogpost and a notebook with it to demonstrate the potential of using the API to create datasets for ML! I don't think that this step of the decorrelation hinders the scenario, but maybe I can redo the description to not be so focused on that small step.

inductiva/molecules/scenarios/md_water_box/md_water_box.py
```diff
@@ -23,19 +23,14 @@ class MDWaterBox(Scenario):
     def __init__(
         self,
         temperature: float = 300,
-        box_size: float = 3.0,
+        box_size: float = 2.3,
```
IvanPombo, line 26: Why this change? :)

inductiva/molecules/scenarios/water_box/water_box.py
IvanPombo: I understand your point! Maybe I am confused about the nature of the randomness, since in CFD that does not exist.

IvanPombo (approved):

inductiva/molecules/scenarios/md_water_box/md_water_box.py
IvanPombo: Let's inform the user about that!

inductiva/scenarios.py
```diff
@@ -47,6 +49,13 @@ def get_config_filename(self, simulator: Simulator):
         """To be implemented in subclasses."""
         pass
 
+    def read_commands_from_file(self, template_dir):
```
fabiocruz, line 52: This should probably accept the full path to the commands file, and not a template dir and search for "commands.json". This can be tacked in a new PR.

inductiva/templates/md_water_box/gromacs/simulation.mdp.jinja
```diff
@@ -0,0 +1,29 @@
+title                   = MD with Nose-Hoover thermostat
+
+; Run parameters
+integrator              = md                  ; leap-frog integrator
```
fabiocruz, line 4: Should the integrator parameter be used here and/or in other template files?

fabiocruz (approved):
LGTM overall.

I agree with @IvanPombo in that we should think in the future if merging this scenario and `ProteinSolvation` is possible, and how to achieve it in a user-friendly manner.

inductiva/molecules/scenarios/md_water_box/md_water_box.py
cristianacarpinteiro: Updated documentation!

inductiva/scenarios.py
cristianacarpinteiro: I've already changed it to a path but I agree with doing the verification.

inductiva/molecules/scenarios/md_water_box/md_water_box.py
IvanPombo: Ok, what I meant was to maybe add a logging.info telling the user what happened, instead of him requiring to go see the docs about why the simulation failed.

inductiva/molecules/scenarios/md_water_box/md_water_box.py
```diff
@@ -0,0 +1,132 @@
...
+        params={
+            "integrator": self.integrator,
+            "nsteps": self.nsteps,
+            "ref_temp": self.temperature,
+            "nsteps_minim": self.nsteps_minim,
+        },
+        output_filename_paths=[
+            os.path.join(input_dir, "simulation.mdp"),
+            os.path.join(input_dir, "energy_minimization.mdp"),
+        ])
```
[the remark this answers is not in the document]
cristianacarpinteiro: I agree.

inductiva/molecules/scenarios/md_water_box/md_water_box.py
cristianacarpinteiro: Ah, I've corrected it :)

PR #312, opened 2023-06-28
Add launch/delete executers and create resource pool call

luispcunha:
This PR starts adding some functions to manage resources from the client side.
This adds the following functions:
```python
id = inductiva.admin.create_resource_pool()
inductiva.admin.launch_executer(
    name="new-vm",
    machine_type="e2-small",
    executer_type="swash",
    spot=True,
    resource_pool_id=id,
)
inductiva.admin.kill_executer(
    name="new-vm",
)
```

This is not yet implemented with user experience in mind, but we can use this to start thinking about how we are going to introduce the resource_pool_ids in calls to simulators/scenarios.

### DIFF
--- a/inductiva/admin/__init__.py
+++ b/inductiva/admin/__init__.py
@@ -1,2 +1,3 @@
 """Methods for interacting with the admin APIs."""
 from .tasks import get_tasks_info
+from .executers import create_resource_pool, launch_executer, kill_executer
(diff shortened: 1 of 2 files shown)


fabiocruz (approved):
LGTM. This is fine for now. I think the best user experience would be not to have to deal with this complexity of launching/killing resources, and the API should manage that for the user by default. But I understand that this may be useful now for experimentation and even in the future for advanced users to fine tune the resources they want to use.

PR #313, opened 2023-06-29
Added init files to molecules scenarios

david451lima:
There were some init files missing from the protein_solvation scenario - this short pr fixes that


cristianacarpinteiro (approved):
Thanks @david451lima it was my mistake because I only tested it locally!

PR #315, opened 2023-06-30
Add simulate async method to gromacs scenarios

cristianacarpinteiro:
In this PR, I implement the simulate_async in the GROMACS scenarios. 
For brevity I'm doing it this way, replicating the code on the simulate method. However, I think this can be improved and we could just add an argument called async to the simulate method and when it is True we call the run_async method.


david451lima (approved):
LGTM! Now it seems that the scenario is able to simulate asynchronously

PR #320, opened 2023-06-30
Add resource pool ID to simulate calls/Remove track logs functionality

luispcunha:
This PR adds support for passing resource pool IDs when calling simulations with the API.
As there was already a "track_logs" parameter common for several API calls which is no longer supported, I ended up replacing that with the new one in some cases, so this PR also removes the support for tracking logs in the client, which is something that is also no longer supported by the backend.

### DIFF
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -3,17 +3,14 @@
 The relevant function for usage outside of this file is invoke_api().
 Check the demos directory for examples on how it is used.
 """
-import asyncio
-import json
 import os
 import pathlib
 import signal
 import time
 from contextlib import contextmanager
 from typing import Any, Dict, Optional, Type
-from urllib.parse import urlparse
+from uuid import UUID
 
-import websockets
 from absl import logging
 
 import inductiva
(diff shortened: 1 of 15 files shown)


david451lima (approved):
LGTM

inductiva/api/methods.py
```diff
@@ -290,7 +287,7 @@ def invoke_api(method_name: str,
                type_annotations: Dict[Any, Type],
                output_dir: Optional[Path] = None,
                return_type: Type = pathlib.Path,
-               log_remote_execution: bool = False):
+               resource_pool_id: Optional[UUID] = None):
```
fabiocruz, line 290: Add this new argument to the docstring below.

inductiva/api/methods.py
```diff
@@ -371,8 +366,10 @@ def invoke_api(method_name: str,
     return unpack_output(result_list, output_dir, return_type)
 
 
-def invoke_async_api(method_name: str, params, type_annotations: Dict[Any,
-                                                                      Type]):
+def invoke_async_api(method_name: str,
+                     params,
+                     type_annotations: Dict[Any, Type],
+                     resource_pool_id: Optional[UUID] = None):
```
fabiocruz, line 372: Update args description in the docstring.

fabiocruz (approved):
LGTM.

PR #321, opened 2023-07-04
ProteinSolvation now accepts charged proteins

david451lima:
Previous ProteinSolvation scenario only took into account neutral proteins. When the protein is charged there is a necessity to ionize the solution that surounds the protein. This induces a couple of changes in the command pipeline. In this PR we implement a method that runs in the initialization of the scenario - runs gmx pdb2gmx in the cloud and analizes the topol.top file to get the charge information. Then the scenario runs either commands_neutral or commands_charged in the pipeline


inductiva/templates/protein_solvation/gromacs/commands_charged.json
fabiocruz, line 1: Most of these comments are the same as those in `commands_neutral.json`. Why not create a jinja template, `commands.json.jinja` with if-else logic to accommodate the differences between the neutral and charged cases?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -104,10 +106,39 @@ def simulate_async(
         )  # convert to fs and divide by the time step of the simulation (2 fs)
         self.integrator = integrator
         self.nsteps_minim = nsteps_minim
+
+        self.command_file = "commands_charged.json" if self.charged\
+                                else "commands_neutral.json"
```
fabiocruz, line 115: See my comment below on having a single file for this.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -104,10 +106,39 @@ def simulate_async(
...
+                      }],
+                      output_dir=output_dir)
+
+        topology_file = os.path.join(output_dir, "topol.top")
+        with open(topology_file, "r", encoding = "UFT-8") as file:
+            charge = np.sum([
+                float(line.split()[-1])
+                for line in file
+                if line.startswith("; residue")
+            ])
```
fabiocruz, line 141: This bit is cryptic. You need to add at least a few comments explaining what you are doing here.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -104,10 +106,39 @@ def simulate_async(
...
+        output_dir = os.path.dirname(self.protein_pdb)
+
+        simulator.run(output_dir,
+                      commands=[{
+                          "cmd":
+                              "gmx pdb2gmx -f \
+                         protein.pdb -o protein.gro \
+                         -water tip3p -ff amber99sb-ildn",
+                          "prompts": []
+                      }],
```
fabiocruz, line 132: Do you want to move this to a file, for consistency with the `simulate()` method?

fabiocruz (commented):
Thanks @david451lima, this looks like a nice improvement on the protein solvation scenario. I left a few comments to improve the overall readability and organization of the code/files, but everything should be simple and quick to implement.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -23,13 +24,14 @@ def __init__(
         self,
         protein_pdb: str,
         temperature: float = 300,
+        charged: bool = None
```
fabiocruz, line 27: I think some explanation of this parameter is needed in the docstring of the class `init` method.

david451lima:
> Thanks @david451lima, this looks like a nice improvement on the protein solvation scenario. I left a few comments to improve the overall readability and organization of the code/files, but everything should be simple and quick to implement.

Thanks. Just to specify that this is not an improvment but a bug fix. The scenario didn't work with charged proteins.

inductiva/templates/protein_solvation/gromacs/commands_charged.json
david451lima: You are right, that would be much cleaner; I will try to implement that

fabiocruz:
> > Thanks @david451lima, this looks like a nice improvement on the protein solvation scenario. I left a few comments to improve the overall readability and organization of the code/files, but everything should be simple and quick to implement.
> 
> Thanks. Just to specify that this is not an improvment but a bug fix. The scenario didn't work with charged proteins.

I read that as a new feature. We now support charged proteins 😄

david451lima:
Addresses yesterday's discussion. @IvanPombo @fabiocruz

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -72,13 +73,25 @@ def simulate(
...
+        #Compute charge if it is not provided
+        if self.charged is None:
+            self.charged = self.compute_charge()
+
+        #Edit commands.json according to the charge of the protein
+        commands_path = os.path.join(self.template_dir, "commands.json")
+        replace_params_in_template(self.template_dir, "commands.json.jinja", {
+            "warnings": "-maxwarn 1",
+            "solvents": "SOL"
+        }, commands_path)
```
fabiocruz, line 86: Why do you need to replace these parameters in a jinja template if they are constant?

inductiva/templates/protein_solvation/gromacs/commands_neutral.json
fabiocruz, line 1: This file can now be removed?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -86,28 +105,67 @@ def simulate_async(
...
             nsteps_minim: Number of steps for energy minimization.
         """
+
+        #Compute charge if it is not provided
+        if self.charged is None:
+            self.charged = self.compute_charge()
+
+        #Edit commands.json according to the charge of the protein
+        commands_path = os.path.join(self.template_dir, "commands.json")
+        replace_params_in_template(self.template_dir, "commands.json.jinja", {
```
IvanPombo, line 128: The replacement for this specific parameters should happen if it is charged, right? Otherwise, it is empty.

inductiva/templates/protein_solvation/gromacs/commands_charged.json
fabiocruz, line 1: This file can now be removed?

fabiocruz (commented):
Thanks @david451lima. I'm not sure I get the use of the jinja templating engine here, since your options seem to be constant... Did I miss something?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -86,28 +105,67 @@ def simulate_async(
...
+
+        commands = self.read_commands_from_file(commands_path)
         return super().simulate_async(simulator, commands=commands)
 
+    def compute_charge(self, simulator: Simulator = GROMACS()):
+        """Check if the protein is charged."""
+
+        logging.info("Computing the charge of the protein")
+
+        output_dir = os.path.dirname(self.protein_pdb)
```
IvanPombo, line 147: I still vouch to execute the command below only once, instead of repeating it in the `commands.json.jinja`...

Hence, here you could immediately write to your current folder, instead of creating a new one just for this purpose. Otherwise, you could use a temporary folder, that disappears after this execution.

IvanPombo (changes requested):
It looks almost good to go, I still have some questions about the way current structure of the simulate and respective commands.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -1,8 +1,12 @@
 """Protein solvation scenario."""
+
+from absl import logging
+
 from functools import singledispatchmethod
```
IvanPombo, line 5: To follow Google style, abs should come after shutil, since it is not a root package of Python.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -68,13 +74,24 @@ def simulate(
...
+        #Compute charge if it is not provided
+        if self.charged is None:
+            self.charged = self.compute_charge()
+
+        #Edit commands.json according to the charge of the protein
+        commands = self.read_commands_from_file(
+            os.path.join(self.template_dir, "commands.json"))
+
+        if self.charged:
+            commands[3]["cmd"] = commands[3]["cmd"] + " -maxwarn 1"
```
IvanPombo, line 87: I was thinking this would be done via jinja... to systematize with other approaches, wdyt @fabiocruz?

fabiocruz: Agreed. You can write if statements in jinja templates, see e.g. [this](https://jinja.palletsprojects.com/en/3.1.x/templates/#if).

david451lima: >
Thanks @david451lima. I'm not sure I get the use of the jinja templating engine here, since your options seem to be constant... Did I miss something?
>
This was done with jinja templating but I changed because @fabiocruz requested it. Did I misunderstand what you meant Fábio?

fabiocruz: I think there was a misunderstanding, yes. You were replacing variables in the jinja template by constant strings, i.e. strings that did not change with e.g. `self.charged`. This suggested that you didn't need a template after all, a simple json file with commands would suffice.

Now, you are fetching commands from a json file, and then modifying them by hand.

A better solution is to use a jinja template where you pass `self.charged` and use that in an if clause inside the template. Does it make sense?

IvanPombo (approved):
LGTM!

PR #325, opened 2023-07-07
Bug fix: resource_pool utilization in proteinSolvation now works

david451lima:
- Corrected an issue with the charge_computation command;
- If the user specified a resour_pool_id the code wouldn't work, since it wasn't passing the resource_pool_id to the super.simulate methods.

There are some things (change in requirements and __init__ file) that do not relate to this bug fix; it's something that will be used in the next PR (I splitted the changes to have a more manageable PR, but some things were ill divided)

### DIFF
--- a/inductiva/molecules/__init__.py
+++ b/inductiva/molecules/__init__.py
@@ -1,3 +1,3 @@
 #pylint: disable=missing-module-docstring
 from .simulators import GROMACS
-from .scenarios import protein_solvation
+from .scenarios import protein_solvation, _post_processing
--- a/inductiva/molecules/__init__.py
+++ b/inductiva/molecules/__init__.py
@@ -1,3 +1,3 @@
 #pylint: disable=missing-module-docstring
 from .simulators import GROMACS
-from .scenarios import protein_solvation#, _post_processing
+from .scenarios import protein_solvation  #, _post_processing
--- a/inductiva/molecules/__init__.py
+++ b/inductiva/molecules/__init__.py
@@ -1,3 +1,3 @@
 #pylint: disable=missing-module-docstring
 from .simulators import GROMACS
-from .scenarios import protein_solvation, _post_processing
+from .scenarios import protein_solvation#, _post_processing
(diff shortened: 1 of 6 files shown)


inductiva/molecules/__init__.py
```diff
@@ -1,3 +1,3 @@
 #pylint: disable=missing-module-docstring
 from .simulators import GROMACS
-from .scenarios import protein_solvation
+from .scenarios import protein_solvation, _post_processing
```
IvanPombo, line 3: I think this part is on a next PR?

inductiva/templates/protein_solvation/gromacs/charge_computation.json
```diff
@@ -1,5 +1,5 @@
 [{
     "cmd":
-        "gmx pdb2gmx -f protein.pdb -o protein.gro -water tip3p -ff amber99sb-ildn",
+        "gmx pdb2gmx -f 1AKI_clean.pdb -o protein.gro -water tip3p -ff amber99sb-ildn",
```
IvanPombo, line 3: This is extra in here, no?

inductiva/templates/protein_solvation/gromacs/charge_computation.json.jinja
```diff
@@ -0,0 +1,5 @@
+[{
```
IvanPombo, line 1: I think we should follow the approach below of uniformization when calling the protein, but this is ok for now!

requirements.txt
```diff
@@ -18,3 +18,5 @@ imageio-ffmpeg
 ipython
 websockets
 typing-extensions
+MDAnalysis
```
IvanPombo, line 21: This is of a next PR.

IvanPombo (approved):
LGTM! Notice that you have here updates for your next PR on viz.

inductiva/molecules/__init__.py
```diff
@@ -1,3 +1,3 @@
 #pylint: disable=missing-module-docstring
 from .simulators import GROMACS
-from .scenarios import protein_solvation
+from .scenarios import protein_solvation  #, _post_processing
```
fabiocruz, line 3: This comment was left in the code.

fabiocruz (commented):
Nice catch, @david451lima! Left one minor comment only.

PR #326, opened 2023-07-07
Added visualization class to protein scenarios

david451lima:
Added a new class that can be integrated with the ProteinSolvation scenario in its default state. Makes so that the user only needs to run the methods .render() and .visualize() to obtain the vizualization in a notebook

Idealized workflow:

```
vizualization = GROMACSSimulationOutput(task_id = task_id)
vizualization.render()
vizualization.visualize()
```

And then the trajectory output is displayed in the notebook.


fabiocruz:
Can you show examples of the outputs produced with the visualization utilities, please?

david451lima:
> 

I will show them in the API sync today

david451lima:
> Can you show examples of the outputs produced with the visualization utilities, please?


https://github.com/inductiva/inductiva/assets/70587819/9799d622-062b-4430-b1d3-241d3d9bb8ff

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,100 @@
...
+from inductiva.tasks import fetch_task_output, get_task_info
+
+import json
+
+SCENARIO_TEMPLATE_DIR = os.path.join(TEMPLATES_PATH, "protein_visualization")
+GROMACS_TEMPLATE_INPUT_DIR = "gromacs"
+
+
+class GROMACSSimulationOutput:
+    """Post process Gromacs simulation outputs."""
```
fabiocruz, line 25: Capitalize "GROMACS".

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,100 @@
...
+        if not get_task_info(task_id)["status"]:
+            raise ValueError("Simulation not finished.")
+
+        self.sim_output_dir = sim_output_path
+
+        if sim_output_path is None:
+            self.sim_output_dir = os.path.join(os.getcwd(),
+                                               f"inductiva_output/{task_id}")
+
+        fetch_task_output(task_id, output_dir=sim_output_path, return_type=None)
```
fabiocruz, line 41: I think we should avoid this pattern where a visualization class/function is also performing other tasks such as downloading task outputs. That should be done separately. All the visualization helper should do is take files stored locally and produce image/video(s). I'd only accept `sim_output_path` as an argument to `__init__`.

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,100 @@
...
+
+        logging.info("Rendering trajectory file .")
+        GROMACS().run(self.sim_output_dir,
+                      commands=commands,
+                      output_dir=self.sim_output_dir,
+                      resource_pool_id=resource_pool_id)
+
+        logging.info("Trajectory file rendered.")
+
+    def visualize(self, pdb_file=None):
```
fabiocruz, line 75: The `render` method in other scenario output classes is usually responsible for creating the visualization. I don't see why the pattern should be different in this case. The `render` method should prepare the render **and show it**. WDYT?

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,100 @@
...
+
+        trajectory = os.path.join(self.sim_output_dir, self.trajectory)
+        system = mda.Universe(pdb_file, trajectory)
+
+        view = nv.show_mdanalysis(system)
+        view.add_ball_and_stick(
+            "all")  # Render the molecules as ball-and-stick models
+        view.center()  # Center the view
+        view.parameters = {
+            "backgroundColor": "white"
```
fabiocruz, line 91: Make this an argument of `visualize` (or `render`).

inductiva/molecules/scenarios/_post_processing.py
fabiocruz, line 75: Is this supposed to be called after `render`?

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,100 @@
...
+                      resource_pool_id=resource_pool_id)
+
+        logging.info("Trajectory file rendered.")
+
+    def visualize(self, pdb_file=None):
+        """Visualize the simulation outputs in a notebook."""
+
+        if pdb_file is None:
+            for filename in os.listdir(self.sim_output_dir):
+                if filename.endswith(".pdb"):
```
fabiocruz, line 80: If this fails, `pdb_file` will be `None` in the rest of the function. An exception needs to be thrown somehow.

fabiocruz (changes requested):
Thank you @david451lima. I have a number of questions about new patterns used here. See below.

inductiva/molecules/scenarios/_post_processing.py
IvanPombo: However, this already poses a nice methodology to run things remotely. So in the future this might be required. But I agree with Fabio, that for now this should not occur, and this post-processing could happen on the scenario side.

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,100 @@
...
+        logging.info("Rendering trajectory file .")
+        GROMACS().run(self.sim_output_dir,
+                      commands=commands,
+                      output_dir=self.sim_output_dir,
+                      resource_pool_id=resource_pool_id)
+
+        logging.info("Trajectory file rendered.")
+
+    def visualize(self, pdb_file=None):
+        """Visualize the simulation outputs in a notebook."""
```
IvanPombo, line 76: What is the output view here? Could this be easily adapted to run locally on VSC?

IvanPombo (commented):
Overall, I agree with Fábio! For now we should maintain these classes with only the visualisation capabilities and the post-processing should be performed on the scenario side. I do not see any reason for this not to be done, since in essence it is just adding this command line there.

fabiocruz:
> Overall, I agree with Fábio! For now we should maintain these classes with only the visualisation capabilities and the post-processing should be performed on the scenario side. I do not see any reason for this not to be done, since in essence it is just adding this command line there.

@IvanPombo maybe I wasn't clear, but that was not my point. I'm ok with this post-processing class/function executing a task in our API. What I meant was that the simulation and post-processing were not decoupled enough. The post-processing shouldn't have to know about a task ID of the simulation. All it should need is a directory where data is stored.

IvanPombo:
In this sense, you would have to download before, or perform a post-processing with data inside the bucket. 

Maybe we should clarify what this types of classes are supposed to do then! Because they can contain further post-processing steps.

cristianacarpinteiro (changes requested):
My main comment is that adding the command to transform the output file can be moved to the scenario this would make this cleaner.

sarmento:
A couple of quick comments: 

1) would you mind adding a few visualization here so we can comment?
2) wrt to the visualization above, it would be great to understand what is the time frame. There should be an optional setting to have the time (in nano seconds?) being shown in the video.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -54,7 +55,8 @@ def simulate(
             resource_pool_id: Optional[UUID] = None,
             simulation_time: float = 10,  # ns
             integrator: Literal["md", "sd", "bd"] = "md",
-            nsteps_minim: int = 5000):
+            nsteps_minim: int = 5000,
+            visualized_section: str = "Protein-H"):
         """Simulate the solvation of a protein.
 
         Args:
```
cristianacarpinteiro, line 62: Add the new parameter to the documentation and mention the options possible for this argument.

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,66 @@
...
+
+        if pdb_file is None:
+            raise ValueError("No PDB file found in the output directory.")
+
+        protein_file = os.path.join(self.sim_output_dir, pdb_file)
+        trajectory = os.path.join(self.sim_output_dir, trajectory_name)
+        system = mda.Universe(pdb_file, trajectory)
+
+        view = nv.show_mdanalysis(system)
+        view.add_ball_and_stick(
```
cristianacarpinteiro, line 59: It could be a good idea to allow the user to have more visualizations than just ball and stick, like liccorice or other. But maybe we can address it in another PR :)

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,66 @@
...
+        """Visualize the simulation outputs in a notebook using NGLView.
+        
+        Args:
+            pdb_file: Path to the PDB file to be visualized.
+            trajectory_name: Name of the trajectory file to be visualized."""
+
+        if pdb_file is None:
+            for filename in os.listdir(self.sim_output_dir):
+                if filename.endswith(".pdb"):
+                    pdb_file = filename
```
cristianacarpinteiro, line 49: This will give the last file with the pdb extension. Is there just one file in the directory? What do we do when there is more than one?  

Maybe use glob here to avoid the for cycle: 
```
pdb_pattern = os.path.join(self.sim_output_dir, "*.pdb")
pdb_files = glob.glob(pdb_pattern)
```

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,66 @@
...
+        Given a simulation output directory that contains the standard files
+        generated by a GROMACS simulation run, this class provides methods to
+        visualize the simulation outputs in a notebook.
+        
+        Args:
+            sim_output_path: Path to the simulation output directory."""
+
+        self.sim_output_dir = sim_output_path
+
+    def render(self, pdb_file=None, trajectory_name: str = "trajectory.xtc"):
```
cristianacarpinteiro, line 39: I don't think it is a good idea to initialize the name of the trajectory file, since it may be error inducing. But maybe mention that we only support xtc types of files!

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -146,14 +151,17 @@ def simulate_async(
         self.integrator = integrator
         self.nsteps_minim = nsteps_minim
 
-        return super().simulate_async(simulator,
+        self.task_id = super().simulate_async(simulator,
```
cristianacarpinteiro, line 154: Why dont you return the task_id ?

cristianacarpinteiro: You can set the variable and still return the task_id.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -189,6 +197,14 @@ def compute_charge(self,
         is_charged = abs(charge) > 1e-6
         return is_charged
 
+    def download_outputs(self, output_dir: str = None, resource_pool_id: Optional[UUID] = None):
```
cristianacarpinteiro, line 200: This will only work when you do the simulate_async method so maybe say that in the docstring

cristianacarpinteiro (approved):
My main question is: should we assume that the simulationOutput will be only used by scenarios constructed by us? Or should we allow the user to  give a folder of his choice with its own structure? Also, don't forget to update the description of the PR when you finish development :)

cristianacarpinteiro:
Also, shouldn't we do the modifications you made to the scenario to the waterbox?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
cristianacarpinteiro: Maybe this can go to the base scenario, no ? WDYT?

inductiva/molecules/scenarios/_post_processing.py
david451lima: Wdym with "initialize the name of the trajectory file"?

inductiva/molecules/scenarios/_post_processing.py
david451lima: Thanks

inductiva/molecules/scenarios/_post_processing.py
fabiocruz: Why not do both within `render`? This is just a rendering method that performs a complex operation (involving a call to the API). But, in the end, it should actually render the output!

inductiva/molecules/scenarios/_post_processing.py
fabiocruz, line 39: Add type annotation `str` to `pdb_file`. Also, maybe be consistent in the argument naming. These are file names, right? Why not call them `pdb_file_name` and `trajectory_file_name`?

inductiva/molecules/scenarios/_post_processing.py
```diff
@@ -0,0 +1,66 @@
...
+            pdb_file: Path to the PDB file to be visualized.
+            trajectory_name: Name of the trajectory file to be visualized."""
+
+        if pdb_file is None:
+            for filename in os.listdir(self.sim_output_dir):
+                if filename.endswith(".pdb"):
+                    pdb_file = filename
+
+        if pdb_file is None:
+            raise ValueError("No PDB file found in the output directory.")
```
fabiocruz, line 52: I think this can be replaced by a different, simpler pattern that doesn't need looping over files:
```python
pdb_file_path = os.path.join(self.sim_output_dir, pdb_file_name)
if not os.path.exists(pdb_file_path):
    raise ...
```

fabiocruz: Also, why are you not checking if the trajectory file exists as well?...

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -54,7 +55,8 @@ def simulate(
             resource_pool_id: Optional[UUID] = None,
             simulation_time: float = 10,  # ns
             integrator: Literal["md", "sd", "bd"] = "md",
-            nsteps_minim: int = 5000):
+            nsteps_minim: int = 5000,
+            visualized_section: str = "Protein-H"):
```
fabiocruz, line 59: It's weird to be setting visualization parameters in the scenario methods. How about calling this `output_trajectory_section`? This way you are defining the section of the trajectory to be computed, and not exactly what should/will be visualized...

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -85,7 +87,8 @@ def simulate(
 
         replace_params_in_template(self.template_dir, "commands.json.jinja", {
             "pdb_file": self.protein_pdb,
-            "charged": self.charged
+            "charged": self.charged,
+            "selected_part": visualized_section
```
fabiocruz, line 91: Why are you giving these parameters/variables different names, if you can control both?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
fabiocruz, line 59: This new parameter should be documented.

fabiocruz (approved):
LGTM. Left a few comments to improve the code quality and readability, but it looks otherwise ok.

inductiva/templates/protein_solvation/gromacs/commands.json
cristianacarpinteiro, line 1: Don't forget to remove this file.

PR #335, opened 2023-07-13
Added Ambertools simulator to the API

david451lima:
Added AmberTools simulator (necessary for the ProteinLigand scenario) to the API.
Low-level calls to ambertools are now supported.


sarmento:
This is great.
Could you explain how this works by giving an example?

inductiva/molecules/simulators/ambertools.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pathlib
+from typing import Optional, List
+from uuid import UUID
+
+from inductiva import types
+from inductiva.simulation import Simulator
+
+
+class Ambertools(Simulator):
```
sarmento, line 11: I am still struggling with the fact that this is a set of tools (as the name explains) and we are still forcing this to fit the "Simulator" use case and then have "commands" being sent.

I suggest we discuss this with @hpenedones soon.

david451lima:
> This is great.
> Could you explain how this works by giving an example?

Sure!
AmberTools provides a suite of commands, passed through the CLI in the standard fashion (e.g. "parmchk2 -i ligand.mol2 -f mol2 -o ligand.frcmod") that allow the ligand part of the protein + ligand model to be pre-processed in order to be simulated by GROMACS.

The structure of the executer is completely equivalent to a standard executer.

sarmento:
> > This is great.
> > Could you explain how this works by giving an example?
> 
> Sure! AmberTools provides a suite of commands, passed through the CLI in the standard fashion (e.g. "parmchk2 -i ligand.mol2 -f mol2 -o ligand.frcmod") that allow the ligand part of the protein + ligand model to be pre-processed in order to be simulated by GROMACS.
> 
> The structure of the executer is completely equivalent to a standard executer.

I see: but why does the class extend "Simulator". It seems awkward. Maybe this is just an effect related with what we understand by simulator has been changing but we still haven't updated the nomenclature? This is what I think needs to be discussed with @hpenedones

inductiva/molecules/simulators/ambertools.py
```diff
@@ -0,0 +1,52 @@
...
+    """Class to invoke any Ambertools command on the API."""
+
+    @property
+    def api_method_name(self) -> str:
```
cristianacarpinteiro, line 15: Is this implemented on the backend? Did you test it already?

cristianacarpinteiro (approved):

hpenedones:
> I see: but why does the class extend "Simulator". It seems awkward. Maybe this is just an effect related with what we understand by simulator has been changing but we still haven't updated the nomenclature? This is what I think needs to be discussed with @hpenedones

Yes, let's think deeper about this and discuss it. At the moment I don't have a "solution", however I can list some guiding principles:

1) Simplicity: we have been keeping the number of abstraction layers / concepts to a minimum, and I think this as served well so far. However, like Luís said, we don't need to go to the extreme of trying to fit everything under the same concept (e.g "Simulator"), if it's not a natural fit. Similarly, recently we avoided incorporating "design optimization" aspects into a simulation "Scenario" concept, as it would not feel natural.

2) Naming vs Interfaces: if we have two concepts (e.g "Simulator" and "Tool") that have exactly the same interface (e.g. a method called "run" that accepts exactly the same kind of parameters), this would be and indication that the implementation of the two can be unified, but we just need a more generic name (e.g. "Processor" or "Runner" or whatever). On the other hand, if the two concepts differ in the natural interface they should expose, (e.g. they should have methods with different names, with different arguments, or returns values) then we should not try to merge them, and we can create two different classes.

3) Usability and convenience: from what I said in 2) one could go to the extreme of defining a super abstract, all-encompassing concept called "ComputationalProcess", or "TuringMachine" that can do anything inside it, as long as it receives some bits and outputs some bits. However, we are in the business of developing a useful and convenient API, where the user can talk about things at a high level of abstraction that "fell natural" instead of everything being super low-level. (Think: you can always implement anything in byte code, or ASSEMBLY, but it is much more convenient to write code in Python)

Therefore I suggest to @david451lima and @cristianacarpinteiro and others: 

- What do this kind of auxiliary tools do (pre-processing and post-processing)?
- What is the "natural" interface to call them, does it overlap totally with the interface for simulators or just partially? Is there anything that differentiates them (e.g. one command versus many? being able to call Linux shells commands as well?)?
- How about runtime assumptions? Are these faster than running simulations?
- Other aspects?

david451lima:
@sarmento @hpenedones
I just checked and Ambertools can be used as a simulator as well (albeit a "worse" one compared to Amber), so, regardless of the fact that we will use this primarily for its pre-processing capabilities in our scenarios, I think that we can leave it as a Simulator for now.
WDYT?

fabiocruz:
I agree with most of what @hpenedones wrote above.

Just wanted to leave another option here to gather your thoughts: if this is a set of tools commonly used with GROMACS, we could also provide them within our GROMACS executer (I am assuming the list of tools available in each executer will be well documented somewhere in the future).

cristianacarpinteiro:
I like @fabiocruz suggestion. This would simplify scenarios and avoid having more than 1 executer being called for the binding scenario, for example. And in the future, when we want to run a scenario end-to-end running on amber we can implement the executer.

PR #338, opened 2023-07-13
Fix Gromacs visualization and remove files commited by mistake 

cristianacarpinteiro:
As the name of the PR suggests I am removing the commited files by mistake and fixing the reading of the pdb file in the render method of the visualization.


david451lima (approved):
LGTM

PR #340, opened 2023-07-14
Add Commands to WindTunnel

IvanPombo:
In this PR, I adapt the WindTunnel to fit with the commands workflow of the scenarios. 

To keep running things in parallel, I made a commitment here to add a command for `mpirun`, which is not appropriate but does not blocking me for moving forward with dataset creation, demos and more. Next week, I will discuss more clearly this and implement the appropriate approach in the backend.

Left to do is to remove the `windtunnel_simulation.py` from the backend, but this will not interfere anymore with the execution.


inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -69,8 +73,7 @@ def __init__(self,
             self.domain = domain
 
     def simulate(self,
-                 simulator: Simulator = OpenFOAM(
-                     "windtunnel.openfoam.run_simulation"),
+                 simulator: Simulator = OpenFOAM(),
```
begaiym-k, line 76: Was this change addressed in a backend, will this work?

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -141,15 +145,32 @@ def simulate_async(self,
...
         return task_id
 
+    def get_commands(self):
+        """Returns the commands for the simulation.
+        """
+        templates_path = os.path.join(SCENARIO_TEMPLATE_DIR,
+                                      OPENFOAM_TEMPLATE_INPUT_DIR)
+        commands_file_path = os.path.join(templates_path, "commands.json")
+
+        replace_params_in_template(templates_path, "commands.json.jinja",
```
begaiym-k, line 166: `"commands.json.jinja"` is a global variable.

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -141,15 +145,32 @@ def simulate_async(self,
...
         )
 
         return task_id
 
+    def get_commands(self):
+        """Returns the commands for the simulation.
+        """
+        templates_path = os.path.join(SCENARIO_TEMPLATE_DIR,
+                                      OPENFOAM_TEMPLATE_INPUT_DIR)
+        commands_file_path = os.path.join(templates_path, "commands.json")
```
begaiym-k, line 164: Should this be a global variable as well? Or just remove jinja extension from this `COMMANDS_TEMPLATE_NAME` env variable since I assume they should be similar.

inductiva/fluids/simulators/openfoam.py
```diff
@@ -44,10 +44,9 @@ def run(
     def run_async(
         self,
         input_dir: types.Path,
-        method_name: str,
+        commands: List[dict],
```
begaiym-k, line 47: Since this is going to be accessible from a low level Openfoam calls, is there any security check done to the commands?

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -15,11 +15,15 @@
...
 from inductiva.utils.templates import (TEMPLATES_PATH,
-                                       batch_replace_params_in_template)
+                                       batch_replace_params_in_template,
+                                       replace_params_in_template)
 from inductiva.utils.files import remove_files_with_tag
+from inductiva.fluids.scenarios.wind_tunnel.post_processing import WindTunnelSimulationOutput
 
 SCENARIO_TEMPLATE_DIR = os.path.join(TEMPLATES_PATH, "wind_tunnel")
 OPENFOAM_TEMPLATE_INPUT_DIR = "openfoam"
+FILES_SUBDIR = "files"
```
begaiym-k, line 25: For curiosity purpose only, what are the files under `files` dir?

inductiva/templates/wind_tunnel/openfoam/commands.json.jinja
begaiym-k, line 1: Since this can be easily modified and sent to the backend, which causes security concerns. Maybe we can have a copy of this file in the backend and every time we have a new call we can compare the body of this file received from the client side to the body on the backend side. Wdyt? (it doesn't have to be done now)

begaiym-k (approved):
Thanks @IvanPombo! I just left a few comments and questions, but mainly I have the security concerns, but I think that is something we will be working on over time, so I am approving it now to not block you.

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
IvanPombo: This works, on the backend just needs to be removed to reduce entropy.

inductiva/fluids/simulators/openfoam.py
IvanPombo: Not yet, that will be done on the backend, as for the other simulators, this is still a security concern.

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
IvanPombo: Are the actual template files to configure the simulation. This folder structure simplifies the work flow with the commands files. In this way, they are not passed to the input_dir.

inductiva/templates/wind_tunnel/openfoam/commands.json.jinja
IvanPombo: I agree that this needs to be taken care of. I do not know if that is the right solution or not, since for the low-level calls this also needs to be address. This is a piece of technical debt that I am comfortable to pass for the next days.

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
fabiocruz, line 166: Use global variables defined above?

inductiva/templates/wind_tunnel/openfoam/commands.json.jinja
fabiocruz, line 1: Cool temporary solution! We need to understand how to deal with this in the backend later.

PR #359, opened 2023-07-18
Protein Solvation now works with a single task

david451lima:
The Portien Solvation scenario now only needs to create one task - the command pipeline now is in a configuration that is able to withstand both cases (neutral and charged proteins) without requiring pre-processing computations


sarmento:
I am really happy on how this is becoming simpler and simpler.
Thank you for this!

PR #363, opened 2023-07-19
Updated demo for gromacs low level call

david451lima:
Updated demo for gromacs low level call with the utilization of gmx pdb2gmx in a given .pdb file


demos/py_scripts/gromacs.py
```diff
@@ -38,20 +24,23 @@
 
 
 def main(_):
-    """Run a Gromacs energy minimization using user-provided input files."""
+    """Run a Gromacs command."""
+
+    inductiva.api_key = "6e2a3936beba573da6c4f81b23b2650395c6911633397ee9dec8307b51c5f451"
```
begaiym-k, line 29: remove the key!

demos/py_scripts/gromacs.py
```diff
@@ -8,28 +8,14 @@
...
-                    required=True)
-
-flags.DEFINE_string("sim_config_filename",
-                    None,
-                    "Name of the input file.",
-                    required=True)
-
-flags.DEFINE_string("protein_filename",
+flags.DEFINE_string("input_dir",
                     None,
```
begaiym-k, line 12: Can you also add the directories of the correct files from the `config_files` please, because when I am testing I am not sure which files I am supposed to use.

david451lima:
@begaiym-k can I close this PR?

demos/py_scripts/gromacs.py
hpenedones: and it's not enough to remove it, because it stays here on github forever, you need to invalidate this one, and regenerate a new one for you.

begaiym-k (approved):
Thanks for updating the demo script @david451lima! LGTM!

PR #366, opened 2023-07-19
Add visualization to MDWaterBox scenario

cristianacarpinteiro:
In this PR I've created the MDWaterBoxSimulationOutput class that creates a visualization for the water box simulation. 
Example code: 
```
from inductiva.molecules.scenarios import MDWaterBox
output = MDWaterBox().simulate(output_dir="water_small", simulation_time=0.1)
view = output.render()
view
```
I was able to solve the problem with the infinite boundaries: 
Before: 

https://github.com/inductiva/inductiva/assets/114397668/323c13cc-ab00-4d1b-a911-f91b20da4b47

After:

https://github.com/inductiva/inductiva/assets/114397668/7568b73b-d134-4635-9580-605f73a46083


cristianacarpinteiro:
@fabiocruz @IvanPombo can we change the name of the visualization classes from MDWaterBoxSimulationOutput to MDWaterBoxOutput and WindTunnelSimulationOutput to WindTunnelOutput, so on... It's too big and I think it hinders readability.

david451lima (approved):
Great Work Cristiana!
LGTM

fabiocruz:
> @fabiocruz @IvanPombo can we change the name of the visualization classes from MDWaterBoxSimulationOutput to MDWaterBoxOutput and WindTunnelSimulationOutput to WindTunnelOutput, so on...? It's too big and I think it hinders readability.

Sounds great to me 👍

inductiva/molecules/scenarios/md_water_box/post_processing.py
```diff
@@ -0,0 +1,39 @@
+"Postprocessing steps for the MDWaterBox scenario."
+import os
+import MDAnalysis as mda
+import nglview as nv
+from pathlib import Path
+
+
+class MDWaterBoxSimulationOutput:
```
IvanPombo, line 8: Yes, you can decrease the name. I even want to discuss further if this is the right usage inside the scenario, but we can all talk tomorrow about it.

inductiva/templates/md_water_box/gromacs/commands.json
fabiocruz, line 1: This shouldn't be committed, right?

fabiocruz (approved):
LGTM. I think you accidentally committed the generated commands.json file, though.

inductiva/templates/md_water_box/gromacs/commands.json
```diff
@@ -0,0 +1,9 @@
+[
```
IvanPombo, line 1: I think this file is not supposed to be loaded here.

inductiva/templates/md_water_box/gromacs/commands.json
cristianacarpinteiro: yes, we need to fix this! It keeps happening in every PR

inductiva/molecules/scenarios/md_water_box/post_processing.py
```diff
@@ -0,0 +1,39 @@
...
+        universe = mda.Universe(topology, trajectory, all_coordinates=True)
+        atoms = universe.atoms
+        transformation = mda.transformations.unwrap(atoms)
+        universe.trajectory.add_transformations(transformation)
+
+        view = nv.show_mdanalysis(universe)
+        view.add_ball_and_stick("all")
+        view.center()
+        view.parameters = {
+            "backgroundColor": "white"
```
IvanPombo, line 37: This should be configurable.

inductiva/molecules/scenarios/md_water_box/post_processing.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+import MDAnalysis as mda
+import nglview as nv
+from pathlib import Path
+
+
+class MDWaterBoxSimulationOutput:
+    """Post process the simulation output of a MDWaterBox scenario."""
+
+    def __init__(self, sim_output_path: Path = None):
```
IvanPombo, line 11: Is there a way to run them in VSC? Can we way to save the vizs?

IvanPombo (approved):
LGTM! Left only minor comments to provide more configuration capabilities.

inductiva/templates/md_water_box/gromacs/commands.json.jinja
```diff
@@ -1,9 +1,9 @@
 [
     {"cmd": "gmx solvate -cs tip4p -box {{ box_size }} -o conf.gro -p topol.top", "prompts": []},
     {"cmd": "gmx grompp -f energy_minimization.mdp -o min.tpr -pp min.top -po min.mdp -c conf.gro -p topol.top", "prompts": []},
-    {"cmd": "gmx mdrun -s min.tpr -o min.trr -x min.xtc -c min.gro -e min.edr -g min.log", "prompts": []},
+    {"cmd": "gmx mdrun -s min.tpr -o min.trr -c min.gro -e min.edr -g min.log", "prompts": []},
```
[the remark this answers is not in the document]
cristianacarpinteiro: the -x min.xtc is the flag for creating a trajectory file and it is not needed for this step, we just want the last one, the eql.xtc.

inductiva/molecules/scenarios/md_water_box/post_processing.py
cristianacarpinteiro: the background color? I don't think it is that of an important parameter. But we can pass a dict with the view parameters maybe, that would give more flexibility to the user.

inductiva/templates/md_water_box/gromacs/commands.json.jinja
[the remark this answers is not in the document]
IvanPombo: The last one, which is the compressed?

inductiva/molecules/scenarios/md_water_box/post_processing.py
cristianacarpinteiro: I am working on the  saving of the visualizations in another PR. The lib we are using NGLview works only on notebooks because it creates an interactive ipywidget. I am working on saving the viz using NGLview but it is creating some problems. If I can't do it with this I will do it with VMD.

inductiva/molecules/scenarios/md_water_box/post_processing.py
IvanPombo: It might be for design presentation reasons. 

What do you think are the parameters in the view_parameters users should have available?

inductiva/templates/md_water_box/gromacs/commands.json.jinja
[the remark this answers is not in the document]
IvanPombo: I know I am being picky here, but what thus the compress mean here? Is it a selection of "frames" of the full trajectory? Is the frame rate already adequate for most vizs?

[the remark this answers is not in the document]
cristianacarpinteiro: GROMACS generates the precision trajectory .trr file which contains the forces, the velocities and the coordinates of every atom. The compressed .xtc file contains only the coordinates taken every x steps. This is why the trr file has GB of size and the compressed file has only some MB. In this case I defined it as being every 50 steps. Which given the timestep of 0.2 fentoseconds is every 10 fs. The framerate is defined elsewhere, when you render the video.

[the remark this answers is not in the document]
IvanPombo: Wonderful thank you for the explanation! With it, I notice that the 'output_time_step' is not important for the rendering, since the simulations cannot be visualized in real time :D
For processing the video the frame rate can also be one parameter for the user.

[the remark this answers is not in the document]
cristianacarpinteiro: That's a good idea :) I will make that configurable in the jinja file.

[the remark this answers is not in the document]
IvanPombo: But it needs to be in the commands?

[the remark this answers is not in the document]
cristianacarpinteiro: Not in the commands, it is defined in .mdp files - the parameters files!

Issue #362, opened 2023-07-19
Merge `simulate` and `simulate_async`?


IvanPombo:
Following the issues #355 and #356, we could easily merge the `simulate` and `simulate_async` methods into one by using the task object output. In general, both methods are relatively the same, with the exception that we are "logging" and holding in the sync. 

The main reason we had not done it before, was due to a difference in inputs (namely, we did not have `output_dir` for the async) and a difference in outputs (`output_dir` and `task_id`).  With a task object, we could immediately have everything concerning a task, including an attribute for the output path. When doing `task.download()` it would be immediate to that output. Furthermore, adding the task info 

The only problem is that we would discard obtaining the visuals object at the end of `simulate` method. 

Wdyt?

fabiocruz:
I think the plan is for `simulate` to return an output object and `simulate_async` to return a task object, so I don't see how these can be merged (based on the argument that inputs and outputs should be the same).

Returning a task object in `simulate` would require importing the output class in each scenario, e.g.
```python
from inductiva... import DamBreak
from inductiva... import DamBreakOutput

scenario = DamBreak(...)
task = scenario.simulate(...)
task.downlaod_output()
output = DamBreakOutput("output_dir")
```

IvanPombo:
> I think the plan is for `simulate` to return an output object and `simulate_async` to return a task object, so I don't see how these can be merged (based on the argument that inputs and outputs should be the same).
> 
> Returning a task object in `simulate` would require importing the output class in each scenario, e.g.
> 
> ```python
> from inductiva... import DamBreak
> from inductiva... import DamBreakOutput
> 
> scenario = DamBreak(...)
> task = scenario.simulate(...)
> task.downlaod_output()
> output = DamBreakOutput("output_dir")
> ```

You would not need to download the output, the simulate would do it. However, I agree with the import of the output class. 

Notice that most meaningful simulations will need the async method anyway, so the usual workflow will be to always import the output anyway.

fabiocruz:
> You would not need to download the output, the simulate would do it.

You're right!

> Notice that most meaningful simulations will need the async method anyway, so the usual workflow will be to always import the output anyway.

Good point. I'm ok with importing the output.

I am also ok with merging the methods, as long as it is still explicit that we are running sync/async simulations. Were you thinking of adding a parameter `async=True|False` to `simulate`?

cristianacarpinteiro:
I like the idea parameter async=True|False! That's what I thought too. 

[...]

PR #367, opened 2023-07-19
Add task class for async simulations

luispcunha:
This PR adds a task class to wrap methods that can be performed on a simulation that was started asynchronously.

### DIFF
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -281,7 +281,7 @@ def submit_task(api_instance, task_request, params, type_annotations):
     )
 
     task_id = task["id"]
-    logging.info("Simulation ID: %s", task_id)
+    logging.info("Task ID: %s", task_id)
 
     with blocking_task_context(api_instance, task_id):
         if task["status"] == "pending-input":
@@ -293,7 +293,7 @@ def submit_task(api_instance, task_request, params, type_annotations):
                 type_annotations=type_annotations,
             )
 
-            logging.info("Simulation successfully submitted.")
+            logging.info("Task request submitted.")
 
     return task_id
 
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -19,7 +19,7 @@ from inductiva.client.apis.tags.tasks_api import TasksApi
 from inductiva.client.models import (BodyUploadTaskInput, TaskRequest,
                                      TaskStatus)
 from inductiva.exceptions import RemoteExecutionError
-from inductiva.tasks.task import Task
+from inductiva.tasks import Task
 from inductiva.types import Path
 from inductiva.utils.data import (extract_output, get_validate_request_params,
                                   pack_input, unpack_output)
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -8,7 +8,7 @@ import pathlib
 import signal
 import time
 from contextlib import contextmanager
-from typing import Any, Dict, Optional, Type
+from typing import Any, Dict, List, Optional, Tuple, Type
 from uuid import UUID
 
 from absl import logging
@@ -19,10 +19,12 @@ from inductiva.client.apis.tags.tasks_api import TasksApi
 from inductiva.client.models import (BodyUploadTaskInput, TaskRequest,
                                      TaskStatus)
 from inductiva.exceptions import RemoteExecutionError
+from inductiva.tasks.task import Task
 from inductiva.types import Path
 from inductiva.utils.data import (extract_output, get_validate_request_params,
                                   pack_input, unpack_output)
 from inductiva.utils.meta import get_method_name, get_type_annotations
+from inductiva import utils
 
 
 def validate_api_key(api_key: Optional[str]) -> Configuration:
@@ -37,6 +39,13 @@ def validate_api_key(api_key: Optional[str]) -> Configuration:
     api_config.api_key["APIKeyHeader"] = api_key
 
     return api_config
+
+
+def get_client() -> ApiClient:
+    """Returns an ApiClient instance."""
+    api_config = validate_api_key(inductiva.api_key)
+
+    return ApiClient(api_config)
 
 
 def submit_request(api_instance: TasksApi, request: TaskRequest) -> TaskStatus:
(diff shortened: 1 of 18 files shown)


david451lima (approved):
LGTM

inductiva/api/methods.py
```diff
@@ -265,7 +281,7 @@ def submit_task(api_instance, task_request, params, type_annotations):
     )
 
     task_id = task["id"]
-    logging.info("This simulation has the task_id: %s", task_id)
+    logging.info("Simulation ID: %s", task_id)
```
IvanPombo, line 284: Shall we call it Task ID, to uniformize things in the logs?

inductiva/api/methods.py
```diff
@@ -70,13 +79,17 @@ def upload_input(api_instance: TasksApi, task_id, original_params,
...
     """
+    logging.info("Creating input archive...")
     input_zip_path = pack_input(
         params=original_params,
         type_annotations=type_annotations,
         zip_name=task_id,
     )
+    file_size = os.path.getsize(input_zip_path)
+    logging.info("Input archive created.")
+    logging.info("Input archive size: %s", utils.format_bytes(file_size))
```
IvanPombo, line 90: Do you want to add this to the downloads or you will do it in another way?

inductiva/api/methods.py
```diff
@@ -277,7 +293,7 @@ def submit_task(api_instance, task_request, params, type_annotations):
                 type_annotations=type_annotations,
             )
 
-            logging.info("Request submitted.")
+            logging.info("Simulation successfully submitted.")
```
IvanPombo, line 296: Task submitted? I would not say successfully here.

inductiva/api/methods.py
```diff
@@ -428,36 +444,20 @@ def run_simulation(
...
-        output_dir=output_dir,
-        return_type=pathlib.Path,
-        resource_pool_id=resource_pool_id,
-    )
-
-    if not isinstance(result, pathlib.Path):
-        raise RuntimeError(f"Expected result to be a Path, got {type(result)}")
-
-    return result
+    return task.download_output(output_dir)
```
IvanPombo, line 452: This goes in hand with a discussion we are having on #362. But we can move forward and when we merge we update this.

inductiva/api/methods.py
```diff
@@ -428,36 +444,20 @@ def run_simulation(
     resource_pool_id: Optional[UUID] = None,
 ) -> pathlib.Path:
     """Run a simulation synchronously via Inductiva Web API."""
+    task = run_async_simulation(api_method_name, input_dir, params,
```
IvanPombo, line 447: This method then should be called differently, if it can be used inside the run_simulation.... However, note my comment below and we will adapt it on the next iteration.

inductiva/tasks/task.py
```diff
@@ -0,0 +1,147 @@
+"""Interact with a running/completed task on the Inductiva API."""
```
IvanPombo, line 1: This represents the task, failed or sucessfully, so it is more than just interaction.

inductiva/tasks/task.py
```diff
@@ -0,0 +1,147 @@
...
+from typing_extensions import TypedDict
+
+from inductiva.client.models import TaskStatusCode
+from inductiva import api
+from inductiva.client.apis.tags.tasks_api import TasksApi
+from absl import logging
+
+
+class Task:
+    """Represents a running/completed task on the Inductiva API.
```
IvanPombo, line 14: Same.

inductiva/tasks/task.py
```diff
@@ -0,0 +1,147 @@
...
+
+        self.kill()
+        return False
+
+    def get_status(self) -> TaskStatusCode:
+        """Get status of the task.
+
+        This method issues a request to the API.
+        """
+        resp = self._api.get_task_status(self._get_path_params())
```
IvanPombo, line 63: This is querying the redis stream right? Should it query the database? What is the correct option?

inductiva/tasks/task.py
```diff
@@ -0,0 +1,147 @@
...
+        return TaskStatusCode(resp.body["status"])
+
+    def get_info(self) -> Dict[str, Any]:
+        """Get a dictionary with information about the task.
+
+        This method issues a request to the API.
+        """
+        params = self._get_path_params()
+        resp = self._api.get_task(params, skip_deserialization=True).response
```
IvanPombo, line 72: Same.

cristianacarpinteiro (approved):
Looks good! Can you provide a code snippet with the usage of the task class and some outputs of the get_status, get_info and wait?

inductiva/tasks/task.py
```diff
@@ -0,0 +1,147 @@
...
+        return resp.json()
+
+    def wait(self, polling_period: int = 5) -> TaskStatusCode:
```
IvanPombo, line 75: This is great and seems to provide both functionality to manage the tasks in the backend and frontend.

IvanPombo (approved):
This looks great @luispcunha! I think there are a couple of things that could be adapted moving on, but for now LGTM!

inductiva/api/methods.py
luispcunha: It will have to be done differently, so I'll do it later in a different PR.

inductiva/api/methods.py
luispcunha: I don't think there's any issue with the name. It's natural that "run_simulation" involves running the simulation asynchronously and then blocking waiting for it to end. This way there's only one path irregardless if the user calls a sync simulation or calls it asynchronously and then waits, the logs and behavior will be the same.

inductiva/tasks/task.py
luispcunha: This is sending a request to the API, without any awareness of what the API uses underneath to serve the request.

inductiva/tasks/task.py
luispcunha: What do you suggest? This is just the one-liner, and I think reflects well what the class does.

luispcunha:
> Looks good! Can you provide a code snippet with the usage of the task class and some outputs of the get_status, get_info and wait?

Updated the docstring of the Task class.

inductiva/tasks/task.py
```diff
@@ -0,0 +1,153 @@
...
+    def get_status(self) -> TaskStatusCode:
+        """Get status of the task.
+
+        This method issues a request to the API.
+        """
+        resp = self._api.get_task_status(self._get_path_params())
+        return TaskStatusCode(resp.body["status"])
+
+    def get_info(self) -> Dict[str, Any]:
+        """Get a dictionary with information about the task.
```
fabiocruz, line 73: Which type of information? Detail this in the docstring.

fabiocruz (approved):
Thanks @luispcunha, good job! LGTM.

inductiva/tasks/task.py
luispcunha: Its still not "set in stone". An upcoming iteration could replace this with a dataclass

Issue #347, opened 2023-07-14
Add parameter to function that replaces parameters in template files to remove template file


fabiocruz:
In some cases, it is useful to delete the template file after it is filled with parameters. This happens for instance when we copy a directory with files describing a scenario, some of them template files.

Removing the template file after it is used to generate a final config file is an action being repeated in multiple scenarios that could be avoided with a parameter in this function that removes the template file at the end of the parameter replacement.

PR #370, opened 2023-07-19
Function that replaces parameters now receives template path, can delete template

### LINKED ISSUE #344: Simplify methods in scenario classes to generate auxiliary files
We currently have 2-3 methods to generate auxiliary files. Consolidate them into one method. Particular scenario classes can decompose auxiliary file generation into multiple functions, but it shouldn't be a requirement from the base class that multiple functions exist for the same purpose.

### LINKED ISSUE #346: Update function that replaces parameters in template files to receive template file path
It currently receives the parent directory of the template file and its filename separately. This introduces unnecessary complexity in scenario classes.

fabiocruz:
Fixes #346, #347.

In an upcoming PR that fixes #344, I plan to further simplify template generation in scenario classes.

### DIFF
--- a/inductiva/utils/templates.py
+++ b/inductiva/utils/templates.py
@@ -15,6 +15,7 @@ def replace_params_in_template(
     template_path: str,
     params: Dict,
     output_file_path: str,
+    remove_template: bool = False,
 ) -> None:
     """Replaces parameters in a template file."""
 
@@ -27,6 +28,9 @@ def replace_params_in_template(
     template = environment.get_template(template_filename)
     stream = template.stream(**params)
     stream.dump(output_file_path)
+
+    if remove_template:
+        os.remove(template_path)
 
 
 def batch_replace_params_in_template(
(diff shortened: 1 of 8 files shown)


inductiva/utils/templates.py
```diff
@@ -10,24 +12,33 @@
...
+
+    template_dir = template_path.parent
+    template_filename = template_path.name
+
+    environment = Environment(loader=FileSystemLoader(template_dir))
     template = environment.get_template(template_filename)
     stream = template.stream(**params)
     stream.dump(output_file_path)
 
+    if remove_template:
```
IvanPombo, line 32: Can you clarify if this removes the template files from the new output_path? As it is, it just seems to remove from the local folder where we are pointing for the templates.

IvanPombo (approved):
Minor comment. Lgtm.

inductiva/utils/templates.py
fabiocruz: This function takes as input a template file and produces an output file. This if condition determines whether to remove the template file after creating the output file.

This is particularly useful when we copy a directory with one or more template files to another location that serves as the input directory for a simulation, and then replace parameters in those template files. It can be used to immediately remove the template files after their conversion is completed.

cristianacarpinteiro (approved):
Looks much cleaner @fabiocruz !

Issue #354, opened 2023-07-17
[Launch machines] Refactor method and create a Machines object


begaiym-k:
- Simplify the existing script, ex: inductiva.resources.launch_machine_group().
- Create a class object that will have terminate(), start(), estimate_cost() methods.
- Pass that object to every simulate/run methods available.

david451lima:
Example code that we want the user to write:

```
machine_group = inductiva.resources.create_machine_group(
    machine_type="c2-standard-4",
    spot=True,
    num_machines=1,
)

machine_group.estimate_cost()
machine_group.start()
machine_group.terminate()
```

IvanPombo:
I had this idea to integrate a name in the machine group that associates it with a project (ours or the user project). In this way, the user can query the cost and resources consumed for that specific project (time and cost).

PR #371, opened 2023-07-20
Add resources sub directory and machines class

begaiym-k:
This PR adds a Machines class to the resources directory. This closes issue #354.

### DIFF
--- a/inductiva/resources/__init__.py
+++ b/inductiva/resources/__init__.py
@@ -1,0 +1,2 @@
+from machines import Machines
+from resources import *
--- a/inductiva/resources/machines.py
+++ b/inductiva/resources/machines.py
@@ -1,0 +1,107 @@
+"""Containes Machines class to manage resources."""
+
+from absl import logging
+from typing import Optional
+from uuid import UUID
+import uuid
+import inductiva
+from inductiva import api
+
+from inductiva.client import ApiClient, ApiException
+from inductiva.client.apis.tags.instance_api import InstanceApi
+from inductiva.client.model.instance import Instance
+from inductiva.client.model.instance_group import InstanceGroup
+
+
+class Machines():
+    """Class to Google Cloud manage resources."""
+
+    def __init__(
+        self,
+        machine_type: str,
+        executer: str,
+        num_machines: int = 1,
+        spot: bool = False,
+        disk_size_gb: int = 30,
+        label: str = None,
+        resource_pool_id: Optional[UUID] = None,
+    ) -> None:
+        """Initialize the Machines class.
+
+        Args:
+            machine_type: The type of GC machine to launch. Ex: "e2-medium".
+            executer: The type of executer to launch. Ex: "gromacs", "openfoam".
+            num_machines: The number of machines to launch.
+            spot: Whether to use spot instances.
+            disk_size_gb: The size of the disk in GB. (min. 30 GB)
+            label: The label to assign to the machine group.
+            resource_pool_id: The resource pool ID to use. A new resource pool
+              id can be created using the
+              inductiva.resources.create_resource_pool() method.
+        """
+        self.machine_type = machine_type
+        self.num_machines = num_machines
+        self.spot = spot
+        self.disk_size_gb = disk_size_gb
+        self.label = label
+        self.executer = executer
+        self.resource_pool_id = resource_pool_id
+        self.name = self._generate_instance_name()
+
+        self.api_config = api.validate_api_key(inductiva.api_key)
+
+    def start(self):
+        with ApiClient(self.api_config) as client:
+            api_instance = InstanceApi(client)
+
+            body = InstanceGroup(
+                name=self.name,
+                machine_type=self.machine_type,
+                num_instances=self.num_machines,
+                executer=self.executer,
+                spot=self.spot,
+                resource_pool_id=self.resource_pool_id,
+                disk_size_gb=self.disk_size_gb,
+            )
+            try:
+                logging.info("Creating a machine group. \
+                             This may take a few minutes.")
+                api_instance.create_instance_group(body=body)
+                logging.info("Machine group is successfully created.")
+            except ApiException as e:
+                raise e
+
+    def terminate(self):
+
+        with ApiClient(self.api_config) as client:
+            api_instance = InstanceApi(client)
+
+            try:
+                logging.info("Terminating machine group. \
+                             This may take a few minutes.")
+                api_instance.delete_instance_group(body=Instance(
+                    name=self.name))
+                logging.info("Machine group is successfully terminated.")
+
+            except ApiException as e:
+                raise e
+
+    def estimate_cost(self):
+
+        with ApiClient(self.api_config) as client:
+            api_instance = InstanceApi(client)
+
+            try:
+                instance_price = api_instance.get_instance_price(body=Instance(
+                    name=self.machine_type))
+                logging.info("Estimated on-demand cost: %f",
+                             instance_price.body["on_demand"])
+                logging.info("Estimated spot cost: %f",
+                             instance_price.body["preemptible"])
+            except ApiException as e:
+                raise e
+
+    def _generate_instance_name(self):
+        unique_id = uuid.uuid4().hex[:8]
+        instance_name = f"api-{unique_id}"
+        return instance_name
--- a/inductiva/resources/resources.py
+++ b/inductiva/resources/resources.py
@@ -1,0 +1,43 @@
+"""Methods related to managing executers and resource pools."""
+from typing import Optional
+from uuid import UUID
+
+import inductiva
+from inductiva import api
+from inductiva.resources import Machines
+from inductiva.client import ApiClient, ApiException
+from inductiva.client.apis.tags.executers_api import ExecutersApi
+
+
+def create_resource_pool() -> UUID:
+    api_config = api.validate_api_key(inductiva.api_key)
+
+    with ApiClient(api_config) as client:
+        api_instance = ExecutersApi(client)
+
+        try:
+            api_response = api_instance.create_resource_pool()
+        except ApiException as e:
+            raise e
+
+    return UUID(api_response.body["id"])
+
+
+def launch_machine_group(
+    machine_type: str,
+    executer: str,
+    num_machines: int = 1,
+    spot: bool = True,
+    disk_size_gb: int = 30,
+    label: Optional[str] = None,
+) -> None:
+
+    resource_pool_id = create_resource_pool()
+
+    return Machines(machine_type=machine_type,
+                    num_machines=num_machines,
+                    spot=spot,
+                    executer=executer,
+                    disk_size_gb=disk_size_gb,
+                    label=label,
+                    resource_pool_id=resource_pool_id)


inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+        self,
+        machine_type: str,
+        executer: str,
+        num_machines: int = 1,
+        spot: bool = False,
+        disk_size_gb: int = 30,
+        label: str = None,
+        resource_pool_id: Optional[UUID] = None,
+    ) -> None:
+        """Initialize the Machines class.
```
hpenedones, line 29: A constructor method (i.e. __init__ in Python) does not "initialize a class", it just creates / constructes and object that is an instance of this class. 

Say instead: "Creates a MachineGroup instance".

inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+from inductiva import api
+
+from inductiva.client import ApiClient, ApiException
+from inductiva.client.apis.tags.instance_api import InstanceApi
+from inductiva.client.model.instance import Instance
+from inductiva.client.model.instance_group import InstanceGroup
+
+
+class Machines():
+    """Class to Google Cloud manage resources."""
```
hpenedones, line 17: typo? word order... ?
Maybe you meant "Class to manage Google Cloud resources."?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+import inductiva
+from inductiva import api
+
+from inductiva.client import ApiClient, ApiException
+from inductiva.client.apis.tags.instance_api import InstanceApi
+from inductiva.client.model.instance import Instance
+from inductiva.client.model.instance_group import InstanceGroup
+
+
+class Machines():
```
hpenedones, line 16: Maybe call it "MachineGroup"?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+    """Class to Google Cloud manage resources."""
+
+    def __init__(
+        self,
+        machine_type: str,
+        executer: str,
```
hpenedones, line 22: The user shouldn't need to provide an executer...

inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+        num_machines: int = 1,
+        spot: bool = False,
+        disk_size_gb: int = 30,
+        label: str = None,
+        resource_pool_id: Optional[UUID] = None,
```
hpenedones, line 27: If this will be the id of this set of MachineGroup, then let's call this, machine_group_id, otherwise we are using two different names (machine_group and resource_pool, for the same thing).

inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+    ) -> None:
+        """Initialize the Machines class.
+
+        Args:
+            machine_type: The type of GC machine to launch. Ex: "e2-medium".
+            executer: The type of executer to launch. Ex: "gromacs", "openfoam".
```
hpenedones, line 33: again, executer should no longer be needed...

inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+            num_machines: The number of machines to launch.
+            spot: Whether to use spot instances.
+            disk_size_gb: The size of the disk in GB. (min. 30 GB)
+            label: The label to assign to the machine group.
+            resource_pool_id: The resource pool ID to use. A new resource pool
+              id can be created using the
```
hpenedones, line 39: It's unclear why we tell the user that a resource pool id can be created by invoking that method (does that mean that the user should worry about doing it?)... Probably what you meant was to specify the default behaviour if None is passed, right? That's a good idea. Rephrase to say "If None is provided, a unique id will be automatically generated.... I don't think we need to say how... but if so, say something like "... by invoking inductiva.resources.new_machine_group_id() " ....

inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+        self.spot = spot
+        self.disk_size_gb = disk_size_gb
+        self.label = label
+        self.executer = executer
+        self.resource_pool_id = resource_pool_id
+        self.name = self._generate_instance_name()
+
+        self.api_config = api.validate_api_key(inductiva.api_key)
+
+    def start(self):
```
hpenedones, line 53: should this method return something?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+            )
+            try:
+                logging.info("Creating a machine group. \
+                             This may take a few minutes.")
+                api_instance.create_instance_group(body=body)
+                logging.info("Machine group is successfully created.")
+            except ApiException as e:
+                raise e
+
+    def terminate(self):
```
hpenedones, line 74: should this method return something?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,107 @@
...
+                logging.info("Terminating machine group. \
+                             This may take a few minutes.")
+                api_instance.delete_instance_group(body=Instance(
+                    name=self.name))
+                logging.info("Machine group is successfully terminated.")
+
+            except ApiException as e:
+                raise e
+
+    def estimate_cost(self):
```
hpenedones, line 89: should this method return something?

inductiva/resources/resources.py
```diff
@@ -0,0 +1,43 @@
...
+from uuid import UUID
+
+import inductiva
+from inductiva import api
+from inductiva.resources import Machines
+from inductiva.client import ApiClient, ApiException
+from inductiva.client.apis.tags.executers_api import ExecutersApi
+
+
+def create_resource_pool() -> UUID:
```
hpenedones, line 12: if this creates an ID change the name of the method to make that clear, otherwise it seems it creates a resource pool (also, see from my other comments, that maybe we should no longer use the term "resource pool" if we are starting to call it "machine group"

hpenedones (changes requested):

inductiva/resources/machines.py
begaiym-k: in principle no, the endpoint returns the instance name , but since the users don't define the name of the instances, I didn't think it was necessary to return the name.

inductiva/resources/machines.py
begaiym-k: Yes, this should return the dict with the prices. I will add it.

inductiva/resources/resources.py
begaiym-k: Yes, I agree, it should have a more explicit name. I will rename it.

inductiva/resources/machines.py
begaiym-k: Since I we decided that users will create the machines by calling `inductiva.resources.launch_machine_group` (which is in the other script), that's where this id is automatically created and is passed here. So users will have to create a new id, only if they wanted to create a machine group directly from this object. Which raises a question, do we even need `inductiva.resources.launch_machine_group` method or should we just let them use this object to create resources?

hpenedones: I think we settled on not having that method. And instead use the machine_group object with methods start() and terminate()...

inductiva/resources/machines.py
```diff
@@ -0,0 +1,112 @@
...
+        label: str = None,
+        machine_group_id: Optional[uuid.UUID] = None,
+    ) -> None:
+        """Create a MachineGroup object.
+
+        Args:
+            machine_type: The type of GC machine to launch. Ex: "e2-medium".
+            num_machines: The number of machines to launch.
+            spot: Whether to use spot instances.
+            disk_size_gb: The size of the disk in GB. (min. 30 GB)
```
fabiocruz, line 33: I don't think this minimum is enforced, so maybe it should say something like "recommended minimum of 30 GB"?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,112 @@
...
+class MachineGroup():
+    """Class to manage Google Cloud resources."""
+
+    def __init__(
+        self,
+        machine_type: str,
+        num_machines: int = 1,
+        spot: bool = False,
+        disk_size_gb: int = 30,
+        label: str = None,
```
fabiocruz, line 24: I think you mean `label: Optional[str] = None` here.

inductiva/resources/machines.py
```diff
@@ -0,0 +1,112 @@
...
+
+        Args:
+            machine_type: The type of GC machine to launch. Ex: "e2-medium".
+            num_machines: The number of machines to launch.
+            spot: Whether to use spot instances.
+            disk_size_gb: The size of the disk in GB. (min. 30 GB)
+            label: The label to assign to the machine group.
+            machine_group_id: If None the Machine Group ID will be generated
+              automatically. A new ID can be created by invoking the
+              inductiva.resources.create_resource_pool() method.
```
fabiocruz, line 37: What is the goal of being able to set the machine group ID when creating a new machine group? This kind of ID is usually automatically generated (to guarantee uniqueness, etc.).

inductiva/resources/machines.py
```diff
@@ -0,0 +1,112 @@
...
+                num_instances=self.num_machines,
+                spot=self.spot,
+                resource_pool_id=self.machine_group_id,
+                disk_size_gb=self.disk_size_gb,
+            )
+            try:
+                logging.info("Creating a machine group. \
+                             This may take a few minutes.")
+                api_instance.create_instance_group(body=body)
+                logging.info("Machine group is successfully created.")
```
fabiocruz, line 68: Remove "is".

inductiva/resources/machines.py
```diff
@@ -0,0 +1,112 @@
...
+                logging.info("Estimated on-demand cost: %f/hour, resulting in \
+                             %f/hour for all machines.",
+                             instance_price.body["on_demand"],
+                             instance_price.body["on_demand"] * \
+                                self.num_machines)
+                logging.info("Estimated spot cost: %f/hour, resulting in \
+                             %f/hour for all machines.",
+                             instance_price.body["preemptible"],
+                             instance_price.body["preemptible"] * \
+                                self.num_machines)
```
fabiocruz, line 104: Why are we logging both of these, since the user has already chosen whether to use spot instances (at initialization)?

inductiva/resources/machines.py
begaiym-k: Yes!

inductiva/resources/machines.py
fabiocruz: I see. I'm just saying the docstring shouldn't say "min. 30 GB" if that minimum isn't enforced somewhere else in the code. Say instead that you recommend a minimum of 30 GB.

inductiva/resources/machines.py
begaiym-k: The goal was to create and ID for the machines and send all the tasks to that specific machine group. Ex: when a user wants to create two consistent machine groups and send the tasks to a specific group, they can use this ID explicitly.

inductiva/resources/machines.py
begaiym-k: I guess I intended to output both prices to give a view at all options. But I think I can only output one.

inductiva/resources/__init__.py
```diff
@@ -0,0 +1,3 @@
+#pylint: disable=missing-module-docstring
+from .machines import MachineGroup
+from .resources import *
```
luispcunha, line 3: we shouldn't import things like this. By doing this you'll also be exposing things that you use in resources.py (e.g. absl.logging and things like that)

inductiva/resources/machines.py
```diff
@@ -0,0 +1,112 @@
...
+            label: The label to assign to the machine group.
+            machine_group_id: If None the Machine Group ID will be generated
+              automatically. A new ID can be created by invoking the
+              inductiva.resources.create_resource_pool() method.
+        """
+        self.machine_type = machine_type
+        self.num_machines = num_machines
+        self.spot = spot
+        self.disk_size_gb = disk_size_gb
+        self.label = label
```
luispcunha, line 43: why label if there's already a name? The user can just simply optionally pass the name or not, do we need the label?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,112 @@
...
+                logging.info("Estimated spot cost: %f/hour, resulting in \
+                             %f/hour for all machines.",
+                             instance_price.body["preemptible"],
+                             instance_price.body["preemptible"] * \
+                                self.num_machines)
+            except ApiException as e:
+                raise e
+        return instance_price.body
+
+    def _generate_instance_name(self):
```
luispcunha, line 109: I don't think this should be here in the client. For now, it's OK if it unblocks and allows you to merge this, but this logic of generating unique IDs and stuff like that should be in the backend.

inductiva/resources/machines.py
begaiym-k: Yes, I would argue for the same workflow, but this is how we started it initially. I will leave it as it is for now and will bring it up to discussion in the meeting.

inductiva/resources/machines.py
```diff
@@ -0,0 +1,112 @@
...
+            )
+            try:
+                logging.info("Creating a machine group. \
+                             This may take a few minutes.")
+                api_instance.create_instance_group(body=body)
+                logging.info("Machine group is successfully created.")
+            except ApiException as e:
+                raise e
+
+    def terminate(self):
```
luispcunha, line 72: How will the user be able to call this method in a different python session?

inductiva/resources/machines.py
luispcunha: I think it's better if you simply use it internally and don't expose it in the MachineGroup API. Later if we implement it differently there will be no changes to user's interface

inductiva/resources/machines.py
begaiym-k: Label is just for the user her/himself to make a note of the machine group, it is not a name. Passing the name of the machine group is not a good idea because that's what is used to create resources and we need to ensure that each resource has a unique name.

inductiva/resources/machines.py
begaiym-k: yes, we can move it to backend later.

inductiva/resources/__init__.py
begaiym-k: Thanks for pointing that out, I will fix it now.

inductiva/resources/machines.py
```diff
@@ -0,0 +1,110 @@
...
+
+            try:
+                logging.info("Terminating machine group. \
+                             This may take a few minutes.")
+                api_instance.delete_instance_group(body=Instance(
+                    name=self.name))
+                logging.info("Machine group successfully terminated.")
+
+            except ApiException as e:
+                raise e
```
sarmento, line 81: Shouldn't you also log here that something didn't go well? What extra info can you give to the user?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def estimate_cost(self):
+        """Returns an estimated cost per hour of a single machine."""
```
hpenedones, line 84: Why does it say it's the price of a single machine but then the method returns the price * num_machines? I think you can update this comment to say it's the total price of the whole machine group per hour, right?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,110 @@
...
+
+        with ApiClient(self.api_config) as client:
+            api_instance = InstanceApi(client)
+
+            try:
+                logging.info("Terminating machine group. \
+                             This may take a few minutes.")
+                api_instance.delete_instance_group(body=Instance(
+                    name=self.name))
+                logging.info("Machine group successfully terminated.")
```
sarmento, line 78: What extra info can you give to the user? How many machines were terminated? How long did it take to terminate? Do you want to say something about the estimated cost so far?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,110 @@
...
+        """Returns an estimated cost per hour of a single machine."""
+        with ApiClient(self.api_config) as client:
+            api_instance = InstanceApi(client)
+
+            try:
+                instance_price = api_instance.get_instance_price(body=Instance(
+                    name=self.machine_type))
+                if self.spot:
+                    logging.info(
+                        "Estimated spot cost: %s/hour per machine, resulting \
```
hpenedones, line 93: In what currency are the prices listed?

inductiva/resources/machines.py
begaiym-k: If there is an internal error in gcloud, it will be printing in the exception. Otherwise the only thing that can go wrong is if the name is wrong and the machine group doesn't exist, but since we are handling the naming that shouldn't be an issue. So there is no need to handle any errors for now.

begaiym-k:
> Hey Begaiym! Thanks for this!
> I have a few comments regarding the info that you are logging back to the user.
> Also: are we storing these events in the database? Are these things being registered to ensure we can have stats later.

Yes, everything is begin logged in the database in the background, since we didn't make nay changes there and just calling the same methods with different interface. 
I just added some extra logs [here](https://github.com/inductiva/inductiva/blob/9b59fd4f1aafe086a1758096ca58e311aface544/inductiva/resources/machines.py#L64), check them out.

cristianacarpinteiro (approved):

begaiym-k:
Since we need to move forward with this feature but I cannot merge it without @hpenedones's approval and @hpenedones is out, I will close this PR and open a new one.

PR #373, opened 2023-07-20
Simplify functions that generate scenario input files

### LINKED ISSUE #307: Refactor gen_aux_files and gen_config in scenarios
If i understood this bit correcly:
- `gen_config` gets the jinja files from the templates directory and creates the final (.mdp) files to be consumed by GROMACS
- `gen_aux_files` copies all files from the templates directory and deletes the jinja files.

This suggests that the workflow (and location of template files) should probably be different. We should not have to copy files, then delete them... Let's open an issue to discuss this further.

_Originally posted by @fabiocruz in https://github.com/inductiva/inductiva/pull/305#discussion_r1243796596_

### LINKED ISSUE #318: Water box scenario generates commands.json file in templates directory
When running `MDWaterBox.simulate()`, a `commands.json` file is incorrectly created under `inductiva/templates/md_water_box/gromacs`.

fabiocruz:
Fixes #307.

This PR refactors the functions that generate scenario input files. There is now one function that does this instead of two, and much of the template code previously required is removed.

I have also taken this opportunity to do a very minor refactoring on scenarios for consistency. This should fix #318 too.


inductiva/fluids/scenarios/fluid_block/fluid_block.py
```diff
@@ -116,22 +115,12 @@ def simulate(
         return SPHSimulationOutput(output_path)
 
     @singledispatchmethod
-    @classmethod
-    def get_config_filename(cls, simulator: Simulator):  # pylint: disable=unused-argument
-        raise ValueError(
-            f"Simulator not supported for `{cls.__name__}` scenario.")
+    def get_config_filename(self, simulator: Simulator):
```
IvanPombo, line 118: This is only required when we have multiple simulators right? We could have a dict for this when we define the valid_simulators.

inductiva/fluids/scenarios/fluid_block/fluid_block.py
```diff
@@ -186,14 +173,9 @@ def _(cls, simulator: DualSPHysics):  # pylint: disable=unused-argument
...
 def _(self, simulator: DualSPHysics, input_dir):  # pylint: disable=unused-argument
-    pass
-
-
-@FluidBlock.gen_config.register
-def _(self, simulator: DualSPHysics, input_dir: str):  # pylint: disable=unused-argument
-    """Generates the configuration file for DualSPHysics."""
+    """Creates DualSPHysics simulation input files."""
 
     replace_params_in_template(
```
IvanPombo, line 180: Are you going to leave the update on the replace_params for a next PR?

I think you could do it here, while you are at it!

IvanPombo (approved):
Wonderfull clean-up @fabiocruz! I just want to add, while you are at it, don't you to update the replace_template from your previous PR?

fabiocruz:
@david451lima @IvanPombo @cristianacarpinteiro, will you help me making sure that this does not introduce breaking changes in the scenarios you own, please?

inductiva/fluids/scenarios/fluid_block/fluid_block.py
fabiocruz: Nop, this is required when there is a main config file for the simulation, which is the case for SPlisHSPlasH, DualSPHysics, or SWASH. This is used to pass a parameter to the backend that specifies which file in the input dir is this config file.

inductiva/fluids/scenarios/fluid_block/fluid_block.py
IvanPombo: Indeed it was, I viewed it wrong due to a line break.

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -18,14 +19,13 @@
...
 from inductiva.utils import files
-from inductiva.utils.files import remove_files_with_tag
 from inductiva.fluids.scenarios.wind_tunnel.post_processing import WindTunnelSimulationOutput
 from inductiva.tasks import Task
 
 SCENARIO_TEMPLATE_DIR = os.path.join(TEMPLATES_PATH, "wind_tunnel")
 OPENFOAM_TEMPLATE_INPUT_DIR = "openfoam"
 FILES_SUBDIR = "files"
-COMMANDS_TEMPLATE_NAME = "commands.json.jinja"
+COMMANDS_TEMPLATE_FILE_NAME = "commands.json.jinja"
```
cristianacarpinteiro, line 28: Do you think it is necessary to define it as a global variable? It will be the same for all the simulators, right?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -145,46 +138,43 @@ def simulate_async(
...
+        commands_template_path = os.path.join(SCENARIO_TEMPLATE_DIR,
+                                              GROMACS_TEMPLATE_INPUT_DIR,
+                                              COMMANDS_TEMPLATE_FILE_NAME)
 
-    @singledispatchmethod
-    def get_config_filename(self, simulator: Simulator):  # pylint: disable=unused-argument
-        raise ValueError(
-            f"Simulator not supported for `{self.__class__.__name__}` scenario."
-        )
+        with tempfile.NamedTemporaryFile() as commands_file:
```
cristianacarpinteiro, line 148: This is awesome and will solve the problem of having always the commands.json commited by mistake. Thanks Fábio :)

cristianacarpinteiro (approved):
Looks much much cleaner now :) Thanks Fábio for all this effort!

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
fabiocruz: Some simulators don't require commands, so it can't be general for all simulators. However, I think this and other functionalities (like the function to replace parameters in a template commands file) should be somewhere where they could be shared. I am still thinking where that could go, so I think I'll leave it for another PR!

IvanPombo:
For the windtunnel is giving the following error:
Screenshot 2023-07-20 at 17 39 01

It concerns the opening of a temporary file when calling the function 

`commands = self.read_commands_from_file(commands_file)`

fabiocruz:
Thanks @IvanPombo for pointing out the error, it should be fixed now.

When trying to run the wind tunnel scenario with default parameters (and no object) I get an error, though:
```shell
AttributeError: 'WindTunnel' object has no attribute 'object_path'
```

Are you sure that the scenario can be run without specifying an object path?

IvanPombo:
It requires an object.

fabiocruz:
> It requires an object.

Then the scenario class should raise an error at initialization when an object is not given.

IvanPombo:
> > It requires an object.
> 
> Then the scenario class should raise an error at initialization when an object is not given.

Actually the object is just passed on the simulate. But I will remove the None...

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -166,13 +166,14 @@ def get_commands(self):
...
         with tempfile.NamedTemporaryFile() as commands_file:
+            logging.info(commands_file.name)
             replace_params_in_template(
                 template_path=commands_template_path,
                 params={"n_cores": self.n_cores},
                 output_file_path=commands_file.name,
             )
 
-            commands = self.read_commands_from_file(commands_file)
+            commands = self.read_commands_from_file(commands_file.name)
```
IvanPombo, line 175: This does not seem to be working. 

`TypeError: expected str, bytes or os.PathLike object, not _TemporaryFileWrapper`

fabiocruz: I confirmed it works on my side. This is likely because your local version isn't updated.

PR #374, opened 2023-07-20
Get force coefficients from WindTunnel Sim

IvanPombo:
Add a post-processing function to fetch the data of the force coefficients in a specific time step.

I would like to note that to compute the true force coefficients we need to a-priori know the frontal surface area. This is a step that I will add afterwards to an object class, instead of directly here.

### DIFF
--- a/inductiva/fluids/scenarios/wind_tunnel/post_processing.py
+++ b/inductiva/fluids/scenarios/wind_tunnel/post_processing.py
@@ -12,10 +12,11 @@ import os
 from dataclasses import dataclass
 from enum import Enum
 from typing import Literal
+import csv
+
 from base64 import b64encode
 from IPython.display import HTML
 
-import pandas as pd
 import pyvista as pv
 
 from inductiva.types import Path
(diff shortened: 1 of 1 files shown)


inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -112,6 +113,27 @@ def get_flow_plane(self,
...
+                                               "postProcessing", "forceCoeffs1",
+                                               "0", "forceCoeffs.dat")
+        force_coefficients_df = pd.read_csv(force_coefficients_path,
+                                            header=None)
+
+        # Skip the first lines of the header to read the force coefficients
+        # on the respective time_step.
+        force_coefficients.loc[0] = force_coefficients_df.loc[self.time_step +
+                                                              num_header_lines,
+                                                              0].split()
```
fabiocruz, line 133: It seems to be that what you're trying to achieve here can be done with a better manipulation of `read_csv`.

- Why are you defining headers by hand? Are they written in the file? They can most likely be read from there.
- The last line is cryptic. What are you doing there?

cristianacarpinteiro: I agree with Fábio. Maybe share an example of the dataframe to see if there is a better way of reading this!

IvanPombo: I added a clarification on the docstring and process the file differently. 

Basically, I am reading a .dat file with the force coefficients over various iterations. On the file header there are several lines that are useless, so I remove them. The goal is to provide the users with the labels and the force coefficients at a respective time step.

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
```diff
@@ -112,6 +114,43 @@ def get_flow_plane(self,
...
+        with open(force_coefficients_path, "r",
+                  encoding="utf-8") as forces_file:
+            for index, line in enumerate(forces_file.readlines()):
+                # Pick the line 8 of the file:
+                # [#, Time, Cm, Cd, Cl, Cl(f), Cl(r)]
+                if index == num_header_lines:
+                    force_coefficients.append(line.split()[1:])
+                # Add the force coefficients for the time_step chosen
+                elif index == num_header_lines + self.time_step + 1:
+                    force_coefficients.append(line.split())
```
fabiocruz, line 145: This is still a bit cryptic. Why are you indexing with `[1:]` the header line and not the others?

fabiocruz (commented):
This is much better @IvanPombo, but I still have a question. Good job on dropping the dependency of pandas, too, btw!

inductiva/fluids/scenarios/wind_tunnel/post_processing.py
IvanPombo: I explain on top that is meant to drop the "#" entry on that list.

fabiocruz (approved):

PR #376, opened 2023-07-20
Render NGLview videos

cristianacarpinteiro:
This PR addresses issue #353 and implements the ability to render a video from a molecular dynamics simulation using NGLview. However, please note that NGLview is an ipywidget specifically designed for Jupyter notebooks, which restricts its functionality to that environment.

Due to a limitation in NGLview, I couldn't directly incorporate this feature into the MDWaterBox.render() method. As per [NGLview's FAQ](https://github.com/nglviewer/nglview/blob/master/docs/FAQ.md), the script must run in a separate thread and wait for the image data using time.sleep(). This threading approach is necessary to prevent blocking the notebook code and maintain smooth notebook execution. 

```
from inductiva.molecules.scenarios.md_water_box.post_processing import MDWaterBoxOutput
from inductiva.molecules.scenarios.utils import MovieMaker

out = MDWaterBoxOutput("output_path")
view = out.render()
view 

# in another cell 
movie = MovieMaker(view, fps=100, stop=100)
movie.make()
```

I've explored another solution to render the video using the [python interface for VMD](https://vmd.robinbetz.com/), however, the documentation is poor and the library was being really tricky to install. Also it's not on pypi which is a blocker for us, at this time.


IvanPombo (commented):
There are two design aspects that need discussion:
- The code proposed isn't simple and requires a particular work flow of using two cells to do the video;
- We need to consider that some videos won't be done in a Jupyter, so moving forward we need to incorporate the ability to visualize in other editors. 

Further, to better understand this workflow I suggest commenting further down the code. 

To my main question: to create images frames do we need to be in a separate cell?

cristianacarpinteiro:
> There are two design aspects that need discussion:
> 
> * The code proposed isn't simple and requires a particular work flow of using two cells to do the video;
> * We need to consider that some videos won't be done in a Jupyter, so moving forward we need to incorporate the ability to visualize in other editors. E.g. https://github.com/sciapp/mogli. I know there are not many options, but it can be the case that having rendering on the backend simplifies the workflow.
> 
> Further, to better understand this workflow I suggest commenting further down the code.
> 
> To my main question: to create images frames do we need to be in a separate cell?

I agree with you Ivan, those are limitations of the current implementation. And I agree with you that we need an option for video rendering outside notebooks. The library you suggested  doesn't deal with molecular dynamics trajectories.

inductiva/molecules/scenarios/md_water_box/post_processing.py
```diff
@@ -1,6 +1,7 @@
 "Postprocessing steps for the MDWaterBox scenario."
 import os
 import MDAnalysis as mda
+import MDAnalysis.transformations
```
fabiocruz, line 4: Use either `from MDAnalysis import transformations` or `import MDAnalysis` and `MDAnalysis.transformations....` below.

inductiva/molecules/scenarios/md_water_box/post_processing.py
```diff
@@ -21,13 +22,14 @@ def __init__(self, sim_output_path: Path = None):
         self.sim_output_dir = sim_output_path
 
     def render(self):
-        """Visualize the simulation outputs in a notebook using NGLView."""
+        """Visualize the simulation outputs in a notebook widget 
+        using NGLView."""
```
fabiocruz, line 26: One-liner. The info about this being generated with NGLView isn't relevant to the user, so it can be moved to a second sentence/paragraph.

requirements.txt
```diff
@@ -20,3 +20,4 @@ websockets
 typing-extensions
 MDAnalysis
 nglview
+moviepy
```
fabiocruz, line 23: Is this really needed? You can do the same with Pillow or imageio, which we have as a dependency already...

inductiva/molecules/scenarios/utils.py
```diff
@@ -0,0 +1,109 @@
...
+import PIL
+from moviepy.editor import ImageSequenceClip
+from ipywidgets import Output, IntProgress
+
+
+class MovieMaker:
+    """ 
+    Class to create a movie from a trajectory rendered on 
+    NGLview.
+    """
```
fabiocruz, line 18: I'm not sure why we need this class at all. All it is doing is looping over frames, saving each individual frame, and then combining the frames into a movie.

This could easily be achieved by having the loop over frames in the output class, and use the function `create_movie_from_frames` in `inductiva/utils/visualization.py` to merge the frames, right?

Also, just checked that `NGLViewer` has a method `download_image` that renders the view and exports it to a file, so all the encoding could be spared?

fabiocruz (commented):
Thanks, @cristianacarpinteiro! I think this could be significantly simplified, but maybe I'm not seeing the whole picture...

inductiva/molecules/scenarios/utils.py
```diff
@@ -0,0 +1,109 @@
...
+                if not event.is_set():
+                    self.view.frame = i
+                    self.sleep()
+                    iw = self.view.render_image()
+                    self.sleep()
+                    im_bytes = base64.b64decode(self.view._image_data)
+                    im_bytes = io.BytesIO(im_bytes)
+                    try:
+                        image = PIL.Image.open(im_bytes)
+                        filename = os.path.join(self.frames_dir,
```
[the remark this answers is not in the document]
cristianacarpinteiro: Thank you, I didn't know this function I will try to use it to remove unnecessary code :)

inductiva/molecules/scenarios/utils.py
cristianacarpinteiro: yes, nglview has that method, however, It was causing problems because it was downloading to the downloads folder independently of the path I put, that's why I did it like this! I will try to use the create_movies_from_frames :)

inductiva/utils/visualization.py
```diff
@@ -30,8 +31,9 @@
     "figure.dpi": 100,
 }
 
+
 def create_movie_from_view(view,
-                           output_path='movie.mp4',
+                           output_path="movie.mp4",
```
IvanPombo, line 36: I think this needs to be more explicit, since here we are not aware what a view is here.

IvanPombo (approved):
I believe this won't be the right solution to move forward, since it is not super clean and not adaptable to under editors.
However, I think it provides the visuals now for the scenario, but I would keep this an high priority to change this in the near future.

inductiva/utils/visualization.py
```diff
@@ -26,6 +32,65 @@
 }
 
 
+def create_movie_from_widget(view,
```
fabiocruz, line 35: Should this function be called from `render_interactive` or from somewhere else?

fabiocruz (approved):
LGTM!

inductiva/utils/visualization.py
cristianacarpinteiro: I don't want to put it in the render_interactive because it messes up the visualization, because whenever you do render_image it updates it. So I would leave it like this and if the user wants to save the video he calls this directly and when we have a better solution in the backend we can have the functionality implemented in the class.

fabiocruz: Sounds good, thanks for explaining the rationale!

PR #380, opened 2023-07-24
Add visualization to heat sink scenario

### LINKED ISSUE #342: [Heat sink scenario] Add ability to visualize simulation results
This should be a first minimalist visualization that gives an impression of the overall air dynamics and temperature around the heat sink. The sink and heater objects should be rendered.

fabiocruz:
Fixes #342.

This PR introduces a `HeatSinkOutput` class with a `render` method. This method produces a video with the temporal evolution of temperature in a simulation of the `HeatSink` scenario.

The temperature is shown on the surface of the heat sink itself, and on (a slice cutting through the center of the heat sink of) the surrounding air.

Here is an example movie obtained with the new class and method.

https://github.com/inductiva/inductiva/assets/11545632/6fa20561-a54e-4e7c-92d4-d7698f25c6c1

### DIFF
--- a/inductiva/fluids/scenarios/heat_sink/__init__.py
+++ b/inductiva/fluids/scenarios/heat_sink/__init__.py
@@ -1,2 +1,3 @@
 #pylint: disable=missing-module-docstring
 from .heat_sink import HeatSink
+from .output import HeatSinkOutput
--- a/inductiva/fluids/scenarios/heat_sink/output.py
+++ b/inductiva/fluids/scenarios/heat_sink/output.py
@@ -1,7 +1,7 @@
-"""TODO
-"""
+"""Post-processing and visualization utilities of the heat sink scenario."""
+
 import os
-from IPython.display import HTML
+import pathlib
 
 import pyvista as pv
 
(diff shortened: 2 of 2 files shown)


IvanPombo (approved):
Well done @fabiocruz! This looks great! To be honest I had troubles making it work with the .vtk data, but with this simplicity I will try again!

Let's start thinking what data could we give the user. Moreover, what other views are important? Front view? Horizontal cut?

fabiocruz:
> Well done @fabiocruz! This looks great! To be honest I had troubles making it work with the .vtk data, but with this simplicity I will try again!

Thanks @IvanPombo!
 
> Let's start thinking what data could we give the user. Moreover, what other views are important? Front view? Horizontal cut?

I think this `render` method could have a `view` parameter that could be either "front", "rear", "top",  "side".

New methods could be added to plot other stuff. I've been thinking that something like the average/max temperature in the fins as a function of time could be useful, for instance.

sarmento:
I think this looks great. The most important thing for me is to ensure the user can choose a viewport. For example, ideally, we don't need to see so much of the space that is located in the side the flow comes from. And maybe we also don't need to have so much area on top of the sink.

I would say that is the most important aspect. 

Also, it would be great to see from above.

fabiocruz:
> I think this looks great. The most important thing for me is to ensure the user can choose a viewport.

Thanks. Added an option to control the view. Here are the videos of the "top" and "side" views.


https://github.com/inductiva/inductiva/assets/11545632/ac0dada5-60a5-403e-80d1-e3e687a31ad1


https://github.com/inductiva/inductiva/assets/11545632/829356c1-9050-4820-97db-427e7731639b

sarmento:
These are much more informative.
Here are a few more suggestions to help improve readibility:

1) Maybe the color of the "background/plane" should be more neutral (black? white?). We don't really care about the temperature of the plane. 

2) if we could zoom in to just see the area where there is actually heat exchange, that would be great. As it is, we actually loose importante detail: what is happening in between the blades? Could we just zoom in to just capture the area where interesting things are happening?

PR #390, opened 2023-07-25
Add coastal area simulation scenario

fabiocruz:
This short PR adds the first version of the coastal area simulation scenario.

For now, it uses a fixed bathymetry for the simulation (Praia do Carneiro, Porto) and injects waves toward the shore line. The user can set the wave amplitude and period, as well as the simulation time, time step and output time step.

I plan to improve this incrementally soon to allow the user to:
1) visualize the simulation results;
2) load an arbitrary bathymetry file.


inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -0,0 +1,131 @@
+"""Coastal area scenario."""
+
+from functools import singledispatchmethod
+import math
+import os
+from typing import Optional
+import shutil
+from uuid import UUID
+
+from inductiva.types import Path
```
sarmento, line 10: As you know, I am not a big fun of importing classes, since it reduces readability down in the code.
wdyt @hpenedones 

I would prefer a simple:

```
import inductiva
```

or maybe have one special type of alias that we use consistently as:
```
import from inductiva.utils.templates as T
```
this is similar to what is done in Pytorch and TF.

sarmento (approved):
LGTM

inductiva/fluids/scenarios/coastal_area/coastal_area.py
fabiocruz: This is done across the client package and demo/example scripts, so I think this deserves an housekeeping issue to be tackled throughout.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -0,0 +1,131 @@
...
+            "simulation_time_hms": _convert_time_to_hmsms(self.simulation_time),
+            "time_step": self.time_step,
+            "output_time_step": self.output_time_step,
+        },
+        output_file_path=config_file_path,
+        remove_template=True,
+    )
+
+
+def _convert_time_to_hmsms(time: float) -> str:
```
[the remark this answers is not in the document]
fabiocruz: I tried using an instance of the built-in `datetime.timedelta` and formatting it as needed, but `timedelta` representation as string doesn't allow for fine control.

This is not thought for the execution time. This is for the simulation time, that SWASH requires to be formated as HHMMSS.sss in input files.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __init__(self, wave_amplitude: float = 2, wave_period: float = 10):
+        """Initializes a `CoastalArea` object.
+
+        Args:
+            wave_amplitude: The amplitude of the wave, in meters.
+            wave_period: The period of the wave, in seconds.
+        """
+        self.wave_amplitude = wave_amplitude
+        self.wave_period = wave_period
```
IvanPombo, line 36: Could we add water_level or orientation of the waves, for example?

inductiva/templates/coastal_area/swash/input.sws.jinja
```diff
@@ -0,0 +1,29 @@
+$$$ HEADING
+PROJECT 'inductiva' '00'
+$$$ PHYSICAL CONSTANTS
+SET temp = 14.0
+SET salinity = 0.1
+SET dynvis = 0.001
+MODE DYNAMIC TWODIMENSIONAL
+CGRID 0. 0. 0. 1200.000000 400.000000 299 99
+INPGRID BOTTOM REGULAR 0.0 0.0 0. 299 99 4.000000 4.000000
+READINP BOTTOM 1. 'bathymetry.bot' 5 0 FREE
```
IvanPombo, line 10: Presumably, to add the bathymetry is just to change here.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def simulate(
+        self,
+        simulator: Simulator = SWASH(),
```
IvanPombo, line 40: We can simply pass the bathymetry here as a .bot file. In this way, it will be simple for anyone to use. 

This follows even if we add a bathymetry class to manage them, in the end, what we would pass is .bot file.

IvanPombo (changes requested):
Great work @fabiocruz! It seems super fast to add a new scenario at this point. 

Do you think we can have a few more parameters available here? Moreover, I think we can allow users to pass their own bathymetries here with a .bot file. What we do later on to generate them will be split from this. In this way, our users could start with this scenario immediately.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -0,0 +1,131 @@
...
+
+        output_path = super().simulate(
+            simulator,
+            sim_config_filename=SWASH_CONFIG_FILENAME,
+            output_dir=output_dir,
+            resource_pool_id=resource_pool_id,
+        )
+
+        return output_path
+
```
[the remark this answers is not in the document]
sarmento: I am not sure we want to have a simulate_async method. We may want to have only a simulate method that has a parameter async=True to behave asynchronously. This way we would *necessarily* have a consistent API in terms of outputs by design. Please chat with @hpenedones about this.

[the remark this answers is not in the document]
fabiocruz: You're right, thanks @IvanPombo. I forgot the async method. We may want to merge these into one soon, but this scenario should have it now anyway.

inductiva/templates/coastal_area/swash/input.sws.jinja
fabiocruz: Not exactly. The region defined by the grid in the bathymetry file is not necessarily that to be simulated, nor is its resolution. It is more involved than that, but we have ideas that we worked on in a previous project.

IvanPombo: I would say that the grid and the resolution can also be responsibility of the user. But I will leave to you, how you prefer doing it.

PR #392, opened 2023-07-25
Coils class (init)

joaobernardo23:


inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -0,0 +1,46 @@
...
+
+class Coil:
+    """Represents one coil of a stellarator.
+
+    The curve provided must be a list of lists, [sin_x, cos_x, 
+    sin_y, cos_y, sin_z, cos_z] where each list has the Fourier 
+    coeffiecients wanted.
+     
+    Attributes: 
+        curve (list): List of the Fourier coefficients defining the coil.
```
fabiocruz, line 38: Should this be called `curve_coefficients` then?

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def simulate(self):
+        pass
+
+class Coil:
+    """Represents one coil of a stellarator.
+
+    The curve provided must be a list of lists, [sin_x, cos_x, 
+    sin_y, cos_y, sin_z, cos_z] where each list has the Fourier 
+    coeffiecients wanted.
```
fabiocruz, line 35: Since this is a list of lists, all of them with the same number of elements, why not receive these in a numpy array of dimensionality `order` x 6?

Also, this requires further explanation, e.g. something along the lines:

"The curve is represented in 3D cartesian coordinates (x, y, z) as a combination of Fourier series Fx * Fy * Fz, each of which is an expansion over trigonometric functions, Fi = sum_j (cj * cos(j * theta) + sj * sin(j * theta)). The coefficients in each row of the array are the cj and sj coefficients of each series Fi."

fabiocruz (changes requested):
Nice first step @joaobernardo23. Left a couple of requests for improvements, but this can quickly be merged so we can incorporate the classmethods of `StellaratorCoils`.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -0,0 +1,46 @@
+"""Class for the coils creation."""
+
+class StellaratorCoils:
```
sarmento, line 3: If this is supposed to be a "scenario", I am not 100% sure that StelleratorCoils is a good name for the class. This is not important for now, but I think it would be beneficial to have a name that invokes the idea of a process, i.e of something that is going to be simulated.

joaobernardo23: Ok, I will take note of this for now. I will try to come up with a better name for the class.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class Coil:
+    """Represents one coil of a stellarator.
+
+    The curve is represented in 3D cartesian coordinates (x, y, z) as a 
+    combination of Fourier series [Fx, Fy, Fz], each of which is an expansion 
+    over trigonometric functions: 
+        Fi = sum_j (sj * sin(j * theta) + cj * cos(j * theta)).
+    The curve provided must then me a numpy array with dimensions 6*order,
```
fabiocruz, line 40: Typo, "be".

fabiocruz: Also, "dimensions 6 * order" should read "shape \`(6, order)\`"

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -0,0 +1,55 @@
...
+
+class Coil:
+    """Represents one coil of a stellarator.
+
+    The curve is represented in 3D cartesian coordinates (x, y, z) as a 
+    combination of Fourier series [Fx, Fy, Fz], each of which is an expansion 
+    over trigonometric functions: 
+        Fi = sum_j (sj * sin(j * theta) + cj * cos(j * theta)).
+    The curve provided must then me a numpy array with dimensions 6*order, 
+    where 'order' (the number of columns) is the maximum order of the series
```
fabiocruz, line 41: Wrap "order" with `, not '.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -0,0 +1,55 @@
...
+    over trigonometric functions: 
+        Fi = sum_j (sj * sin(j * theta) + cj * cos(j * theta)).
+    The curve provided must then me a numpy array with dimensions 6*order, 
+    where 'order' (the number of columns) is the maximum order of the series 
+    (maximum value of j) and the number of rows is 6, one for each of the 
+    sj and cj coefficients of each series Fi.
+     
+    Attributes: 
+        curve_coefficients (list): List of the Fourier coefficients defining
+        the coil.
```
fabiocruz, line 47: The type here should be np.ndarray now, and the description should start with "Array with Fourier coefficients..."

fabiocruz (approved):
Thank you @joaobernardo23! Got minor comments only about wording this time.

PR #396, opened 2023-07-25
[Coastal area scenario] Add output class with basic render capabilities

fabiocruz:
This PR adds a class to represent outputs of the coastal area simulation scenario.

It can be used as other output classes to create renders, e.g.
```python

scenario = CoastalArea()
output = scenario.simulate()
output.render()
```

In this case, the `render` method creates movies of the temporal evolution of quantities computed by SWASH.

Here is the default output:

https://github.com/inductiva/inductiva/assets/11545632/3b1ac579-04b4-491c-acb9-09ba15eb2795

The rendered quantity can be configured by setting e.g. `quantity="velocity_magnitude"` in the call to `render`. The output in this case is the following:

https://github.com/inductiva/inductiva/assets/11545632/7c96caff-5ffd-4346-923a-734b114a30f2

### DIFF
--- a/inductiva/fluids/scenarios/coastal_area/output.py
+++ b/inductiva/fluids/scenarios/coastal_area/output.py
@@ -72,28 +72,28 @@ def read_swash_grid_positions_file(file_path: str) -> Tuple[np.ndarray]:
     return x_array, y_array
 
 
-def read_swash_quantity_file(file_path: str, quantity: str,
-                             grid_positions_file_path: str):
-    """Reads a SWASH output file containing a quantity."""
+def read_swash_quantity_file(file_path: str, quantity: str):
+    """Reads a SWASH output file containing a quantity.
 
-    # Read grid positions
-    x_array, y_array = read_swash_grid_positions_file(grid_positions_file_path)
-
-    # Read data file
-    data_dict = read_swash_output_file(file_path)
-
-    # Convert data dictionary to lists of times and data arrays
+    SWASH outputs are stored in .mat files. Here, we first read the contents
+    of one of these files as a dictionary, and then convert the dictionary
+    to a list of times and a list of data arrays.
 
     # The list of times is built from the keys of the dictionary:
     # the keys end with a number, in format 123456_789, identifying the time
     # instant 12 hours, 34 minutes, 56 seconds and 789 milliseconds; we
     # strip the times from the keys and convert them to seconds.
 
-    # The list of data arrays is build from the values of the dictionary:
+    # The list of data arrays is built from the values of the dictionary:
     # the values correspond to the data values at each of the time instants
     # represented by the corresponding keys; we read each value as a 2D
     # tensor with indices (x, y).
+    """
 
+    # Read data file
+    data_dict = read_swash_output_file(file_path)
+
+    # Convert data dictionary to lists of times and data arrays
     time_list = []
     data_list = []
 
@@ -117,7 +117,7 @@ def read_swash_quantity_file(file_path: str, quantity: str,
             time_list.append(time_s)
             data_list.append(transposed_data_values)
 
-    return (x_array, y_array), time_list, data_list
+    return time_list, data_list
 
 
 def _render_quantity_data(x_array: np.ndarray,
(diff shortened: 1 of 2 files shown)


IvanPombo (approved):
In general LGTM, I think adding a bit more description to the args would be perfect. 

It seems though all of the functions before the class can be used for all SWASH simulations, so I am not sure if we should keep them here or create a swash_viz_utils.py somewhere to be useful for low-level users as well.

inductiva/fluids/scenarios/coastal_area/output.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # The list of times is built from the keys of the dictionary:
+    # the keys end with a number, in format 123456_789, identifying the time
+    # instant 12 hours, 34 minutes, 56 seconds and 789 milliseconds; we
+    # strip the times from the keys and convert them to seconds.
+
+    # The list of data arrays is build from the values of the dictionary:
+    # the values correspond to the data values at each of the time instants
+    # represented by the corresponding keys; we read each value as a 2D
+    # tensor with indices (x, y).
```
cristianacarpinteiro, line 95: why not put this in the description of the function ?

inductiva/fluids/scenarios/coastal_area/output.py
```diff
@@ -0,0 +1,242 @@
...
+            # again for consistency
+            transposed_data_values = data_values.transpose()
+
+            time_list.append(time_s)
+            data_list.append(transposed_data_values)
+
+    return (x_array, y_array), time_list, data_list
+
+
+def _render_quantity_data(x_array: np.ndarray,
```
cristianacarpinteiro, line 123: why not include these functions as methods of the class?

inductiva/fluids/scenarios/coastal_area/output.py
```diff
@@ -0,0 +1,242 @@
...
+            "velocity_x",
+            "velocity_y",
+            "velocity_magnitude",
+        ] = "water_level",
+        movie_path: Path = "movie.mp4",
+        fps: int = 10,
+        cmap: str = "viridis",
+        clim: Optional[Sequence[float]] = None,
+    ):
+        """Renders temporal evolution of a quantity.
```
cristianacarpinteiro, line 208: maybe say what type of quantity, like a physical quantity such as velocity. Quantity seems too general!

cristianacarpinteiro (approved):
LGTM! Just left some comments to understand better your implementation!

inductiva/fluids/scenarios/coastal_area/output.py
fabiocruz: As @IvanPombo said, part of these functions can be used to plot results of other SWASH simulations, so I didn't want to tie them to the class introduced here.

cristianacarpinteiro: Ah makes sense, then shouldn't it be elsewhere?

fabiocruz:
> It seems though all of the functions before the class can be used for all SWASH simulations, so I am not sure if we should keep them here or create a swash_viz_utils.py somewhere to be useful for low-level users as well.

Sort of. These can be used to read and plot outputs of SWASH **defined over a grid**. SWASH exports these and other types of diagnostics, so it can't be used to plot arbitrary output files from SWASH. I added this specification to the names of the methods and corresponding docstrings.

This being said, I agree with you that these functions could be store somewhere where they could be used by other scenarios simulated with SWASH, or even by users using the low-level call. I don't have a good suggestion of where to put these now, though. Let's discuss this when we organize our visualization routines (soon, hopefully!), WDYT?

inductiva/fluids/scenarios/coastal_area/output.py
fabiocruz: I think so, but I don't think this is the moment to make a decision about that. See [my comment](https://github.com/inductiva/inductiva/pull/396#issuecomment-1651421002) below.

IvanPombo:
> > It seems though all of the functions before the class can be used for all SWASH simulations, so I am not sure if we should keep them here or create a swash_viz_utils.py somewhere to be useful for low-level users as well.
> 
> Sort of. These can be used to read and plot outputs of SWASH **defined over a grid**. SWASH exports these and other types of diagnostics, so it can't be used to plot arbitrary output files from SWASH. I added this specification to the names of the methods and corresponding docstrings.
> 
> This being said, I agree with you that these functions could be store somewhere where they could be used by other scenarios simulated with SWASH, or even by users using the low-level call. I don't have a good suggestion of where to put these now, though. Let's discuss this when we organize our visualization routines (soon, hopefully!), WDYT?

Copy!
Yes, it seems appropriate! We can move it to the right place at that point :)

PR #397, opened 2023-07-25
Method to create simple circular coils

joaobernardo23:


inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -26,6 +28,48 @@ def __init__(self, coils, num_field_periods):
...
+            (the central axis) to the outer edge of the plasma region.
+            minor_radius (float): Radius of the simple circular curves.
+
+        Returns:
+            StellaratorCoils: The created StellaratorCoils instance.
+        """
+
+        stellarator_symmetry = True
+        angle_factor = (1+int(stellarator_symmetry)) * \
+            num_field_periods*num_coils
```
fabiocruz, line 50: Since we will always want to have `stellarator_symmetry = True`, why don't we simplify this and have `angle_factor = 2 * num_field_periods * num_coils`?

fabiocruz: Also explain (in a comment) that `angle_factor` is there to spread the coils equally over an angle that is a fraction of 2 $\pi$.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -26,6 +28,48 @@ def __init__(self, coils, num_field_periods):
...
+
+        coils = []
+        for i in range(num_coils):
+            angle = (i + 0.5) * (2 * np.pi) / (angle_factor)
```
fabiocruz, line 54: Explain why 0.5? And maybe rename this to `toroidal_angle`?

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -26,6 +28,48 @@ def __init__(self, coils, num_field_periods):
         self.num_coils = len(coils)
         self.num_field_periods = num_field_periods
 
+    @classmethod
+    def from_circular_curves(cls, num_field_periods, num_coils, coil_currents,
+                             major_radius, minor_radius):
+        """Create simple circular and equally spaced curves.
```
fabiocruz, line 34: You could explain here briefly which coefficients of the Fourier series are non-zero.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -26,6 +28,48 @@ def __init__(self, coils, num_field_periods):
...
+        stellarator_symmetry = True
+        angle_factor = (1+int(stellarator_symmetry)) * \
+            num_field_periods*num_coils
+
+        coils = []
+        for i in range(num_coils):
+            angle = (i + 0.5) * (2 * np.pi) / (angle_factor)
+            curve_coefficients = np.zeros((6, 2))
+
+            # Set the coefficients
```
fabiocruz, line 57: "Set the **non-zero** coefficients."

inductiva/stellarators/scenarios/coils/coils.py
joaobernardo23: I am not sure why the 0.5, but I believe it is necessary so that the first angle is not zero. It may be because of the way simsopt handles the symmetries afterwards.

PR #406, opened 2023-07-26
[Fluid tank scenario] Add output class

fabiocruz:
This PR adds a class to represent the output of a fluid tank simulation. The output produced by the `render` method looks as follows:

https://github.com/inductiva/inductiva/assets/11545632/e2c4852c-386f-427b-a4de-e370728d7283

This PR also adds the method `simulate_async` to the scenario class, which was previously missing.

### DIFF
--- a/inductiva/fluids/scenarios/fluid_tank/fluid_tank.py
+++ b/inductiva/fluids/scenarios/fluid_tank/fluid_tank.py
@@ -18,7 +18,6 @@ from inductiva.fluids.shapes import Cylinder
 from inductiva.fluids.fluid_types import FluidType
 from inductiva.fluids.fluid_types import WATER
 from inductiva.fluids.simulators import SPlisHSPlasH
-from inductiva.fluids.post_processing.splishsplash import convert_vtk_data_dir_to_netcdf
 from inductiva.utils.templates import replace_params_in_template
 from inductiva.fluids.scenarios.fluid_tank.output import FluidTankOutput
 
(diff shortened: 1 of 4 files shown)


inductiva/fluids/scenarios/fluid_tank/fluid_tank.py
```diff
@@ -191,7 +192,42 @@ def simulate(
...
+                radius and time step.
+            particle_sorting: Whether to use particle sorting.
+        """
+
+        self.simulation_time = simulation_time
+        self.particle_radius = ParticleRadius[resolution.upper()].value
+        self.time_step = TimeStep[resolution.upper()].value
+        self.output_time_step = output_time_step
+        self.particle_sorting = particle_sorting
+
```
IvanPombo, line 224: I think here is missing the conversion to netcdf data. Otherwise, running async want allows to obtain any vizs after downloading the outputs.

inductiva/fluids/scenarios/fluid_tank/output.py
```diff
@@ -0,0 +1,61 @@
...
+import os
+from typing import Optional
+
+import xarray as xr
+
+from inductiva.types import Path
+from inductiva.utils import files, visualization
+
+
+class FluidTankOutput:
```
IvanPombo, line 12: Later on we need to standardize the methods of other SPH scenarios with this one. Although this one contains some external dependencies that I am not aware if they will get misconfigured.

IvanPombo (approved):

inductiva/fluids/scenarios/fluid_tank/output.py
fabiocruz: Agreed.

PR #407, opened 2023-07-26
[Fluid tank scenario] Add documentation

fabiocruz:
This PR adds documentation to the fluid tank scenario.


README.md
```diff
@@ -101,6 +101,47 @@ output.render()
...
+scenario = fluids.scenarios.FluidTank(fluid=fluids.WATER, fluid_level=0.5)
+```
+
+The user can specify the fluid (e.g. water, honey or oil), the fluid level (in
+meters), as well as the tank shape and its inlet and outlet properties.
+
+Run the simulation:
+
+```python
+output = scenario.simulate(simulation_time=5,
```
IvanPombo, line 128: I think we should specify that this is only available for SPlisHSPlasH at the moment, since other SPH scenarios contain both.

README.md
```diff
@@ -101,6 +101,47 @@ output.render()
...
+the [Smoothed Particle Hydrodynamics](https://en.wikipedia.org/wiki/Smoothed-particle_hydrodynamics)
+method.
+
+#### Example
+
+Initialize the scenario:
+
+```python
+from inductiva import fluids
+scenario = fluids.scenarios.FluidTank(fluid=fluids.WATER, fluid_level=0.5)
```
IvanPombo, line 119: Maybe, we should clarify how to represent the the shapes.

IvanPombo (approved):
LGTM, with only minor comments. I will adequately change them and merge.

PR #411, opened 2023-07-27
[Protein Solvation] Extract conformational changes metrics

cristianacarpinteiro:
In this PR, I implement the calculation of the root mean square fluctuation over a trajectory. The intent is to show how the ProteinSolvation Scenario can be used to study conformational changes.

This metric quantifies how much a structure diverges from a reference over time, the RSMF can reveal which areas of the system are the most mobile. An area of the structure with high RMSF values frequently diverges from the average, indicating high mobility. When RMSF analysis is carried out on proteins, it is typically restricted to backbone or alpha-carbon atoms; these are more characteristic of conformational changes than the more flexible side-chains. 
Here is an example output of the RMSF per residue: 
image
The user can optionally render this information in a visualisation of the protein structure backbone. Where the darker areas show the most mobile residues. 
image

### DIFF
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -90,8 +90,7 @@ class ProteinSolvationOutput:
         })
 
         # Plot the data
-        plt.plot(df_rmsf_per_atom["residue_number"],
-                     df_rmsf_per_atom["rmsf"])
+        plt.plot(df_rmsf_per_atom["residue_number"], df_rmsf_per_atom["rmsf"])
         plt.xlabel("Residue Number")
         plt.ylabel("RMSF")
         plt.title("RMSF per residue")
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -66,7 +66,12 @@ class ProteinSolvationOutput:
 
     def calculate_rmsf_trajectory(self, nglview_visualization=True):
         """Calculate the root mean square fluctuation (RMSF) of the protein 
-        residues over the trajectory.
+        residues over the trajectory. It is typically calculated for the C&alpha 
+        atom of each residue. It is the square root of the variance of 
+        the fluctuation around the average position:
+        &rhoi = √⟨(xi - ⟨xi⟩)²⟩
+        It quantifies how much a structure diverges from a reference over time, 
+        the RSMF can reveal which areas of the system are the most mobile. 
         Args:
             nglview_visualization: Whether to return visualization of the 
             RMSF using nglview or not."""
@@ -77,27 +82,19 @@ class ProteinSolvationOutput:
         align_universe = mda.Universe(self.topology, aligned_trajectory_path)
 
         # Calculate RMSF per atom
-        atoms = align_universe.select_atoms("protein")
-        rmsf = rms.RMSF(atoms).run()
+        c_alphas = align_universe.select_atoms("protein and name CA")
+        rmsf = rms.RMSF(c_alphas).run()
         df_rmsf_per_atom = pd.DataFrame({
             "rmsf": rmsf.rmsf,
-            "residue_number": atoms.resids
+            "residue_number": c_alphas.resids
         })
-        df_rmsf_per_residue = df_rmsf_per_atom.groupby("residue_number").agg(
-            mean_rmsf=("rmsf", "mean"), std_rmsf=("rmsf", "std"))
-        residue_number = df_rmsf_per_residue.index
-        mean_rmsf = df_rmsf_per_residue["mean_rmsf"]
-        std_rmsf = df_rmsf_per_residue["std_rmsf"]
 
         # Plot the data
-        plt.errorbar(residue_number,
-                     mean_rmsf,
-                     yerr=std_rmsf,
-                     fmt="o",
-                     capsize=5)
+        plt.plot(df_rmsf_per_atom["residue_number"],
+                     df_rmsf_per_atom["rmsf"])
         plt.xlabel("Residue Number")
         plt.ylabel("RMSF")
-        plt.title("RMSF with Standard Deviation per residue")
+        plt.title("RMSF per residue")
         plt.grid(True)
         plt.show()
 
@@ -106,7 +103,7 @@ class ProteinSolvationOutput:
                 "tempfactors")  # add empty attribute for all atoms
             protein = self.universe.select_atoms(
                 "protein")  # select protein atoms
-            for residue, r_value in zip(protein.residues, rmsf.results.rmsd):
+            for residue, r_value in zip(protein.residues, rmsf.rmsf):
                 residue.atoms.tempfactors = r_value
 
             view = nv.show_mdanalysis(self.universe)
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -1,21 +1,18 @@
 "Postprocessing steps for the MDWaterBox scenario."
 import os
 import nglview as nv
+import matplotlib.pyplot as plt
 from pathlib import Path
 from typing import Literal
-from ..utils import unwrap_trajectory, align_trajectory_to_average
 from MDAnalysis.analysis import rms
 import MDAnalysis as mda
-import pandas as pd
-from matplotlib import pyplot as plt
+from inductiva.molecules.scenarios.utils import unwrap_trajectory, align_trajectory_to_average
 
 
 class ProteinSolvationOutput:
     """Post process the simulation output of a ProteinSolvation scenario."""
 
-    def __init__(self,
-                 sim_output_path: Path = None,
-                 use_compressed_trajectory: bool = False):
+    def __init__(self, sim_output_path: Path = None):
         """Initializes a `ProteinSolvationOutput` object.
 
         Given a simulation output directory that contains the standard files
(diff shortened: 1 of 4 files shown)


inductiva/molecules/scenarios/utils.py
```diff
@@ -14,3 +15,21 @@ def unwrap_trajectory(topology, trajectory):
...
+                                     universe,
+                                     select="protein and name CA",
+                                     ref_frame=0).run()
+    average_trajectory = average.results.universe
+
+    align.AlignTraj(universe,
+                    average_trajectory,
+                    select="protein and name CA",
+                    filename=trajectory_output_path,
+                    in_memory=False).run()
```
david451lima, line 35: Can you elaborate in what does this do?

david451lima (approved):
LGTM

inductiva/molecules/scenarios/utils.py
cristianacarpinteiro: First we compute an average structure with the average coordinates of each atom. Then we align the trajectory to that  reference. We can look into more details [here](https://userguide.mdanalysis.org/stable/examples/analysis/alignment_and_rms/rmsf.html).

cristianacarpinteiro:
This Kind of metrics could be cool to study something like this: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4572606/. To note that to observe the conformational change between open and close states there was the need for a temperature change. Something that can easily be implemented.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -3,56 +3,110 @@
 import nglview as nv
 from pathlib import Path
 from typing import Literal
-from ..utils import unwrap_trajectory
+from ..utils import unwrap_trajectory, align_trajectory_to_average
```
IvanPombo, line 6: won't this be part of the inductiva package?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -3,56 +3,110 @@
...
+from MDAnalysis.analysis import rms
+import MDAnalysis as mda
+import pandas as pd
+from matplotlib import pyplot as plt
```
IvanPombo, line 10: Cleaner: import matplotlib.pyplot as plt

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -3,56 +3,110 @@
...
 
 
 class ProteinSolvationOutput:
     """Post process the simulation output of a ProteinSolvation scenario."""
 
-    def __init__(self, sim_output_path: Path = None):
+    def __init__(self,
+                 sim_output_path: Path = None,
+                 use_compressed_trajectory: bool = False):
```
IvanPombo, line 18: Remind that ProteinSolvationOutput will be used in the scenario.simulate(...) output. Hence, we would need to pass the argument `use_compressed_trajectory` there, which means it would be defined in the simulate.

Hence, I suggested this argument is served as an argument in every function it is used below, instead of being an attribute.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -3,56 +3,110 @@
...
         print(f"Number of solvent molecules:"
-              f"{universe.select_atoms('not protein').n_residues}")
-        print(f"Number of trajectory frames: {len(universe.trajectory)}")
+              f"{self.universe.select_atoms('not protein').n_residues}")
+        print(f"Number of trajectory frames: {len(self.universe.trajectory)}")
         return view
+
+    def calculate_rmsf_trajectory(self, nglview_visualization=True):
+        """Calculate the root mean square fluctuation (RMSF) of the protein 
+        residues over the trajectory. It is typically calculated for the C&alpha
```
IvanPombo, line 69: What is this C&alpha?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -3,56 +3,110 @@
...
+        print(f"Number of trajectory frames: {len(self.universe.trajectory)}")
         return view
+
+    def calculate_rmsf_trajectory(self, nglview_visualization=True):
+        """Calculate the root mean square fluctuation (RMSF) of the protein 
+        residues over the trajectory. It is typically calculated for the C&alpha 
+        atom of each residue. It is the square root of the variance of 
+        the fluctuation around the average position:
+        &rhoi = √⟨(xi - ⟨xi⟩)²⟩
+        It quantifies how much a structure diverges from a reference over time,
```
IvanPombo, line 73: What is the reference? Is the initial protein state?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -3,56 +3,110 @@
...
+
+        # Calculate RMSF per atom
+        c_alphas = align_universe.select_atoms("protein and name CA")
+        rmsf = rms.RMSF(c_alphas).run()
+        df_rmsf_per_atom = pd.DataFrame({
+            "rmsf": rmsf.rmsf,
+            "residue_number": c_alphas.resids
+        })
+
+        # Plot the data
```
IvanPombo, line 92: I think this graph should be in another function and called inside here.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -3,56 +3,110 @@
...
+
+        aligned_trajectory_path = os.path.join(self.sim_output_dir,
+                                               "aligned_traj.dcd")
+        align_trajectory_to_average(self.universe, aligned_trajectory_path)
+        align_universe = mda.Universe(self.topology, aligned_trajectory_path)
+
+        # Calculate RMSF per atom
+        c_alphas = align_universe.select_atoms("protein and name CA")
+        rmsf = rms.RMSF(c_alphas).run()
+        df_rmsf_per_atom = pd.DataFrame({
```
IvanPombo, line 87: Do you actually need pandas? It is a slightly heavy package to add to inductiva. Could you do it in another way?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -3,56 +3,110 @@
...
-              f"{universe.select_atoms('protein').n_residues}")
+              f"{self.universe.select_atoms('protein').n_residues}")
         print(f"Number of solvent molecules:"
-              f"{universe.select_atoms('not protein').n_residues}")
-        print(f"Number of trajectory frames: {len(universe.trajectory)}")
+              f"{self.universe.select_atoms('not protein').n_residues}")
+        print(f"Number of trajectory frames: {len(self.universe.trajectory)}")
         return view
+
+    def calculate_rmsf_trajectory(self, nglview_visualization=True):
```
IvanPombo, line 67: I feel that the calculate and visualize should be separate. WDYT?

inductiva/molecules/scenarios/utils.py
IvanPombo: Add a comment with that link here.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -3,56 +3,110 @@
...
+        # Plot the data
+        plt.plot(df_rmsf_per_atom["residue_number"], df_rmsf_per_atom["rmsf"])
+        plt.xlabel("Residue Number")
+        plt.ylabel("RMSF")
+        plt.title("RMSF per residue")
+        plt.grid(True)
+        plt.show()
+
+        if nglview_visualization:
+            self.universe.add_TopologyAttr(
```
IvanPombo, line 101: Why are you adding an empty attribute?

IvanPombo (changes requested):
I have left several questions regarding the implementation. Overall, it seems good but I would like more clarification.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: No, in my implementation is the average structure calculated over the whole simulation.

cristianacarpinteiro: No, in my implementation I calculated the average structure in the whole simulation.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: it's a alpha carbon, which makes the protein backbone.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: I dont need but I thought it looked cleaner to return it to the user. But I can have a dictionary.  what do you think ?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
IvanPombo: Can you add that to the doc. Now that I see it, I would first have a single line docstring and then document the rest in the lines below.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
IvanPombo: I would add that when you say a reference.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
IvanPombo: I agree, but I think users can do that by themselves for now.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: Yes, I agree. I will do it.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: Yes, it is. Should it not be in the utils?

IvanPombo: It can be, but in this way you are doing a local import, instead of a package one.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: it is just to initialise  I set the value by residue atoms after.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: I think it is simple enough to keep it here, now that I've cleaned up a bit more.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -25,34 +28,108 @@ def render_interactive(self,
...
+        plt.ylabel("RMSF")
+        plt.title("RMSF per residue")
+        plt.grid(True)
+        plt.show()
+        return rmsf_values
+
+    def render_attribute_per_residue(self,
+                                     residue_attributes,
+                                     use_compressed_trajectory: bool = False):
+        """Render a specific protein attribute in an interactive visualization.
```
IvanPombo, line 120: This might be specific of proteins, but what here is a residue? I think the name is a bit to long, but I am fine with it.

IvanPombo (approved):
LgTm!

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: a residue is an aminoacid, which is the building unit of a protein. You can have this attributes per atom, for example, that's why I wanted to be specific!

hpenedones:
Cool stuff @cristianacarpinteiro, I am curious to see how this metric of "rigidity" per amino-acid in the chain, correlates to the uncertainty estimates from AlphaFold 2. :) It'a good use case of this scenario. If we have the time at some point (not right now) we might be able to make a small analysis / blog post. :)

PR #413, opened 2023-07-27
Update  GROMACS section README.md

cristianacarpinteiro:
Added some information to the GROMACS scenarios and visualization

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,59 +146,75 @@ coastal area.
 
 ### MDWaterBox
 
-This scenario simulates a system that consists of a cubic box of water molecules, evolving according to the rules of Molecular Dynamics - the position of the water molecules is updated using Newton's equation in discrete time steps. The force that acts upon the particles is computed using standard molecular force fields. This approach is inspired by [this article](https://arxiv.org/abs/2112.03383).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -146,59 +146,75 @@ coastal area.
...
-            nsteps_minim = 5000)
+```python
+output = scenario.simulate(simulation_time = 10, nsteps_minim = 5000)
 ```
 
-The simulate method initializes a simulation in the cloud. In this call, we set the parameters:
+The simulate method initializes a simulation using cloud resources. In this call, we set the following parameters:
  - simulation_time = 10: sets the trajectories time to span 10 **nanosecons**;
- -nsteps_minin = 5000: sets the number of minimization steps in the energy minimization step. 
+ - nsteps_minin = 5000: sets the number of minimization steps in the energy minimization step.
```
IvanPombo, line 170: I feel these defaults are weird here, but I will check over all.

README.md
```diff
@@ -146,59 +146,75 @@ coastal area.
...
+
+The simulation output files can be found in inductiva_output/task_id folder.
```
IvanPombo, line 172: What does this path actually mean? If we are sync, then task_id is irrelevant.

IvanPombo (approved):
LgTm

README.md
cristianacarpinteiro: I will remove this part.

README.md
cristianacarpinteiro: I will take this part out too, in the next PR.

PR #420, opened 2023-07-31
Refactor README

IvanPombo:
In this PR, I refactor the README based on the discussion we had last friday.

The main key points are:
- the removal of the FluidBlock and WaterBox
- setting first the ProteinSolvation and WindTunnel
- Adding a section for machines
- Adding the async method
- Adding a FAQ (in construction)

These are basically the key points of the issue #418.

PR #421, opened 2023-07-31
[Coastal area scenario] Add support for arbitrary bathymetric profiles

### LINKED ISSUE #419: [CoastalArea] Add the capability to change the bathymetry
The scenario will only be usable when users can select their own bathymetries.

fabiocruz:
Fixes #419.

This PR adds support for arbitrary bathymetric profiles in the coastal area scenario. Bathymetries can now be generated arbitrarily, or read from text files:

```python
from inductiva import fluids
bathymetry = fluids.scenarios.Bathymetry(text_file_path="bathymetry.txt", x_range=(0, 1200), y_range=(0, 400))
scenario = fluids.scenarios.CoastalArea(bathymetry=bathymetry, wave_source_location="W")
scenario.simulate()
```

I tried this with the bathymetry from Praia do Carneiro, in Porto (Portugal):
a) using the right bathymetry orientation, and injecting waves from West (left)
https://github.com/inductiva/inductiva/assets/11545632/f1266c94-a723-46c8-b9ce-7eba12abd795

b) using a bathymetry flipped in the $x$ direction, and injecting waves from East (right)
https://github.com/inductiva/inductiva/assets/11545632/d6379df0-56c3-4de5-af7b-1e66ef300aa9


inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -19,22 +21,103 @@
 SWASH_TEMPLATE_SUBDIR = "swash"
 SWASH_CONFIG_TEMPLATE_FILENAME = "input.sws.jinja"
 SWASH_CONFIG_FILENAME = "input.sws"
+SWASH_BATHYMETRY_FILENAME = "bathymetry.bot"
+
+
+class Bathymetry:
```
sarmento, line 27: I love this class. 
A few of comments:

1)  it's not clear if the need for the x_range and y_range is just because we are using txt files. Is there another format where the range info is provided so that we don't need to provide it in the constructor?

2) I would add a clip() method, as well as potentially other utility methods. This could be done later.

3) I would consider adding a method (either in this class or as an util method) to create random bathymetries. I think we need to really use different bathymetries to test this class.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -19,22 +21,103 @@
...
+    """Represents a bathymetric profile.
+    
+    A bathymetric profile defines the depth of the sea bottom as a function of
+    space, here described in Cartesian coordinates (x, y). Here, a bathymetry is
+    represented as a 2D array with the depths, in meters, at each point of a
+    regular grid. The grid is defined by the range of x and y values. Positive
+    depths are below the water level.
+    """
+
+    def __init__(
```
sarmento, line 37: maybe the constructor should be even simpler, and then you can either load or generate a random bathymetry.

sarmento (commented):
I love this.
But I feel that we are still stuck to Praia do Carneiro data.
I would feel much more confident in this Bathymetry class if we could either also generate random bathymetries (we already have a baseline method), or if we could load a different real bathymetry and test the whole thing with new data.

For example, can we load any of these:

https://www.opendem.info/download_bathymetry.html


https://download.gebco.net/

I just downloaded data near Vigo 
[GEBCO_01_Aug_2023_d5d7d17ddfeb.zip](https://github.com/inductiva/inductiva/files/12225398/GEBCO_01_Aug_2023_d5d7d17ddfeb.zip)

I think it is really important to move away from the Praia do Carneiro and see what happens when we try to load / generate other bathymetries.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -54,18 +139,27 @@ class CoastalArea(Scenario):
 
     def __init__(
         self,
+        bathymetry: Bathymetry,
         water_level: float = 0,
+        wave_source_location: Literal["N", "S", "E", "W"] = "W",
```
IvanPombo, line 144: This could be a combination, like "NE", or is it to complicated?

inductiva/fluids/scenarios/coastal_area/coastal_area.py
fabiocruz: > 1. it's not clear if the need for the x_range and y_range is just because we are using txt files. Is there another format where the range info is provided so that we don't need to provide it in the constructor?

Yes, there are other formats where the range info is provided. We can add support for those in the future. For now the goal was to read the bathymetry from a text file (the ones SWASH users load bathymetries from, in any case).

> 2. I would add a clip() method, as well as potentially other utility methods. This could be done later.

Agreed, to be done later.

> 3. I would consider adding a method (either in this class or as an util method) to create random bathymetries. I think we need to really use different bathymetries to test this class.

That's in our (short-term) plans too.

IvanPombo (approved):
LGTM @fabiocruz, wonderful work that closes the new version! 

I agree with the @sarmento point to test with other bathymetries, and would open an issue with the random bathymetries since it will be a wonderful feature for dataset generation we will certainly use!

fabiocruz:
I agree we need to be able to load bathymetries from other locations and sources.

But please be mindful that we can't expect just about any source to work. The one you suggest, GEBCO, has a resolution of 15 arc seconds = 450 meters. AFAIK this is too large for shallow water models. We need resolutions of ~ meters.

I think we did collect a few viable sources of bathymetries in the past. I'll look them up.

fabiocruz:
@IvanPombo I'm merging this now. Can you please update the README file on #420?

sarmento:
OK! I am ok with finding other sources.  But would you please consider that your next step on this thread?
We need to be able to show this scenario using bathymetry data from a known source, otherwise demonstrability is limited. What source can we use? Let's start with just one, but let's make sure we can use more than just the Praia do Carneiro.

IvanPombo:
> OK! I am ok with finding other sources. But would you please consider that your next step on this thread? We need to be able to show this scenario using bathymetry data from a known source, otherwise demonstrability is limited. What source can we use? Let's start with just one, but let's make sure we can use more than just the Praia do Carneiro.

I will link @fabiocruz to a single different bathymetry he can test.

PR #422, opened 2023-08-02
[Coastal area scenario] Add method to generate random bathymetries

### LINKED ISSUE #91: [Coastal area scenario] Add random bathymetry generation
As we develop our API, we want users to also benefit from a number of convenience functions that will help them achieve their job. 

One of the aspect we have been talking about is providing "generative" methods to the API that allow creating artifacts that can be input to simulations. We already have one of such methods: we can generate random bathymetric maps using a fractal methods. 

We should a bathymetry generation method to the API and use it in our notebooks to illustrate one of the aspects of our API. This will also avoid the need to have a (potentially proprietary) real bathymetry file as part of our example notebooks. 

The most important issues to discuss is possibly: should that generative method run locally (client side) or remotely (server side), or should we make it potentially runnable on both sides?

In the long run, we may have pretty complex generative methods (e.g. diffusion models) which will probably require a complex net of dependencies and significant computing power. In that sense, it seems reasonable to have the generative method running remotely, since it seems to be the most scalable pattern.

Then there is an question about how will these generative methods be run. Will they be workers like anyother or should we just have simpler function calls? Should we stick to "one size fits all" model or can there be other ways of execting jobs on the server? This requires discussion with @hpenedones .

Then, there are aspect related with placement in our API, i.e. the class structure.
Should we have something like (with many possible variations):
```
bathymetry = inductiva.fluids.utils.generate_bathymetry(params)
```

or should there be a "generative" section on the API?

fabiocruz:
Fixes #91.

This PR adds a classmethod to the `Bathymetry` class that generate instances of bathymetries with random depths.

The depths are randomly generated using the [Diamond-Square algorithm](https://en.wikipedia.org/wiki/Diamond-square_algorithm). The implementation of the algorithm is heavily based on code from a previous internal project. I did add the ability to control the initial roughness of the algorithm, which was previously set to unity. This caused bathymetries with maximum depth $\gg$ 1 (say $>$ 20) to be very smooth.

Here are a couple of bathymetries generated with the code snippet

```python
bathymetry = Bathymetry.from_random_depths(
    n=9,
    max_depth=10,
    initial_roughness=1,
    roughness_factor=0.5,
    percentile_above_water=20,
    x_range=(0, 1000),
    y_range=(0, 800),
)
```

bathymetry_random_0

bathymetry_random_1


inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -17,6 +17,8 @@
                                        replace_params_in_template)
 from inductiva.fluids.scenarios.coastal_area.output import CoastalAreaOutput
 
+from inductiva.fluids.scenarios.coastal_area import random_bathymetry_utils
```
sarmento, line 20: check Hugo notes on importing functions / classes

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -75,6 +77,58 @@ def from_text_file(
         depths = np.loadtxt(text_file_path)
         return cls(depths, x_range, y_range)
 
+    @classmethod
+    def from_random_depths(
+        cls,
```
sarmento, line 82: what is cls? is this self?

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -75,6 +77,58 @@ def from_text_file(
...
+            initial_roughness: Initial roughness value, in meters. Controls the
+              initial range of randomness of the Diamond-Square algorithm.
+            roughness_factor: Roughness factor. Must be between 0 and 1.
+              Controls the rate at which the range of randomness of the
+              Diamond-Square algorithm decreases over iterations.
+            percentile_above_water: Percentile of the depths that must be above
+              water. Must be between 0 and 100.
+        """
+
+        depths = random_bathymetry_utils.create_random_bathymetry(
```
sarmento, line 123: This is ok for now, but maybe we could generalize a bit the code in random_bathymetry_utils and have it in inductiva/generative because we can use it on a number of other situations (e.g. @IvanPombo may want to use this too).

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -75,6 +77,58 @@ def from_text_file(
         depths = np.loadtxt(text_file_path)
         return cls(depths, x_range, y_range)
 
+    @classmethod
+    def from_random_depths(
+        cls,
+        n: int,
```
sarmento, line 83: maybe you want to explicitly say this is a power of 2.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -75,6 +77,58 @@ def from_text_file(
...
+        x_range,
+        y_range,
+        max_depth: float = 10,
+        initial_roughness: float = 1,
+        roughness_factor: float = 0.5,
+        percentile_above_water: float = 20,
+    ):
+        """Creates a `Bathymetry` object with random depths.
+
+        A grid of random depths of size (2^n + 1) x (2^n + 1) is created, where
```
sarmento, line 93: Maybe we want to make this easier for the user.
I think we should allow the user to directly specify a grid size (e.g. 800 x 600) and then we internally select the right power of 2 and in the end crop the result and ensure the percentile_above_water.

I think this creates a more natural interface.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
fabiocruz: Nop, it's the class. This is a [classmethod](https://docs.python.org/3/library/functions.html?highlight=classmethod#classmethod). The syntax follows commons Python code (inc. the Google style guide, see [this](https://google.github.io/styleguide/pyguide.html#3191-general-rules)).

inductiva/fluids/scenarios/coastal_area/coastal_area.py
fabiocruz: That's a quite intricate choice, relating the percentile above water and the grid size. I think in general it is much more natural that the user "gets what they ask for". 

WDYT of the following:
1. We allow users to specify a grid size (e.g. 800 x 600)
2. We take the maximum size (e.g. 800) and figure out the corresponding n (e.g. 10)
3. We use that n to produce a random bathymetry
4. We interpolate the random grid to a grid with the size selected by the user (e.g. 800 x 600)

inductiva/fluids/scenarios/coastal_area/coastal_area.py
fabiocruz: This follows those notes. Other imports in this file don't, but I'd rather do a general cleanup later.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
sarmento: I think interpolation is fair and is a good solution, because this is a random bathymetry anyway so the interpolation will not "distort" anything. If we see some weird effect, then we can think about correcting it, but for now, I think this is a good solution. Please proceed with the interpolation.

sarmento:
This is looking good. Would you be able to also add a simulation using a quite large bathymetry (e.g. 1000 x 2000), just to test the whole procedure end to end?

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -75,6 +77,58 @@ def from_text_file(
...
+        n is an integer.
+
+        The depths are randomly generated as follows:
+        1. A maximum depth value `max_depth` is used to specify the range of
```
IvanPombo, line 97: With the description below this seems to be `max_amplitude` or something, since depth seems more related to below sea level. Wouldn't it make sense to have two different parameters for both, so that we can have plan beaches with deep waters for example.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
IvanPombo: I was actually thinking of using an open-source library, but given this is already here, I will adapt to my case and then we think about the generalization :)

inductiva/fluids/scenarios/coastal_area/coastal_area.py
IvanPombo: Just questioning:
- Shall we overcomplicate the code at the moment? 
- Could users just say "high", "medium", "low" resolution and we define an appropriate `n`based on that?

inductiva/fluids/scenarios/coastal_area/random_bathymetry_utils.py
```diff
@@ -0,0 +1,251 @@
...
+    for i in range(kernel_scale, map_length, inter_center_distance):
+        for j in range(kernel_scale, map_length, inter_center_distance):
+            bathymetry_array[i,
+                             j] = kernel_avg(diamond_offsets, bathymetry_array,
+                                             i, j, kernel_scale)
+            bathymetry_array[i, j] += random.uniform(-noise_range, noise_range)
+
+
+def square_step(bathymetry_array, inter_center_distance, noise_range):
+    """The square step performed at a given inter center distance on a map.
```
IvanPombo, line 60: I'm confused :D By the algorithm description you mention on Wiki and your drawings this seems to be the diamond and the one above the square...

inductiva/fluids/scenarios/coastal_area/random_bathymetry_utils.py
```diff
@@ -0,0 +1,251 @@
...
+    # Square Step, rows
+    for i in range(kernel_scale, map_length, inter_center_distance):
+        for j in range(0, map_length, inter_center_distance):
+            bathymetry_array[i,
+                             j] = kernel_avg(square_offsets, bathymetry_array,
+                                             i, j, kernel_scale)
+            bathymetry_array[i, j] += random.uniform(-noise_range, noise_range)
+
+    # Square Step, cols
+    for i in range(0, map_length, inter_center_distance):
```
IvanPombo, line 84: All of this double four loops could be transformed into a function, where the main change is in the for loop range.

IvanPombo: This would also point to the main differences between steps of the algorithm.

inductiva/fluids/scenarios/coastal_area/random_bathymetry_utils.py
```diff
@@ -0,0 +1,251 @@
...
+          bathymetries.
+        percentile_above_water: Percentage of the spatial domain, in range
+          (0, 100), that must be above the sea level (<0).
+        scale: Scale of the random variables used to initialize the bathymetry.
+          For a scale of 10, the initial condition will have random values in
+          the range (0, 10) on the left side and (-10,0) on the right side of
+          the bathymetry.
+    Returns:
+        Array with the random bathymetry.
+    """
```
IvanPombo, line 237: Wonderfull description @fabiocruz! Thank you so much for this!

IvanPombo (approved):
Great work @fabiocruz! I left minor comments, lgtm!

Could we add a method to visualize the bathymetry in 2d/3d? You are providing thus with images on the PR, but I do not see any function for me (the user) to visualize them by myself.

inductiva/fluids/scenarios/coastal_area/random_bathymetry_utils.py
```diff
@@ -0,0 +1,251 @@
...
+    inclined_bed[0, size - 1] = random.uniform(0, max_depth)
+
+    # define right corners above water
+    inclined_bed[size - 1, 0] = random.uniform(-max_depth, 0)
+    inclined_bed[size - 1, size - 1] = random.uniform(-max_depth, 0)
+
+    return inclined_bed
+
+
+def create_random_bathymetry(size: int,
```
IvanPombo, line 144: I was trying to run the code on my side and it seems that I can pass all sizes possible. Shouldn't the sizes be fixed for 2^k+1?

The bathymetry is not good for `sizes !=2^k+1`.

fabiocruz: This is a function in a utilities file, that (in principle) shouldn't be used directly by users, so I wouldn't bother much with making sure that this size is some 2^k + 1. I can always add a check, though. WDYT?

inductiva/fluids/scenarios/coastal_area/coastal_area.py
fabiocruz: I think it's ok to keep numbers instead of a choice between "high", "medium", "low" to control the size of the grid.

inductiva/fluids/scenarios/coastal_area/random_bathymetry_utils.py
IvanPombo: A check is enough!

inductiva/fluids/scenarios/coastal_area/random_bathymetry_utils.py
fabiocruz: The drawing is fine, but the description is wrong. Thanks for spotting this.

fabiocruz:
> This is looking good. Would you be able to also add a simulation using a quite large bathymetry (e.g. 1000 x 2000), just to test the whole procedure end to end?

While I'm still working to address the comments in this PR, I ran a few examples. Here is one with a bathymetry of 1025 x 1025:

Bathymetry:

bathymetry_random_10

Simulation output:

https://github.com/inductiva/inductiva/assets/11545632/6d15d623-5e34-481a-8454-eccb505a4988

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -17,6 +19,8 @@
                                        replace_params_in_template)
 from inductiva.fluids.scenarios.coastal_area.output import CoastalAreaOutput
 
+from inductiva.generative import diamond_square
```
sarmento, line 22: Maybe you want to take the chance to correct the import style as mentioned by Hugo. We should avoid directly importing function/classes.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -75,6 +79,96 @@ def from_text_file(
...
+        y_square = np.linspace(*y_range, size_square)
+
+        x_square, y_square = np.meshgrid(x_square, y_square, indexing="ij")
+
+        x = np.linspace(*x_range, x_num)
+        y = np.linspace(*y_range, y_num)
+
+        x, y = np.meshgrid(x, y, indexing="ij")
+
+        depths = scipy.interpolate.griddata(
```
sarmento, line 163: A side effect of this is that it smooths out the bathymetry. It's not a problem, but to get more interesting simulation cases, we probably need to set higher roughness factor (see my comment on the video).

sarmento (approved):
LGTM! This is very cool.
If you can, maybe fixe some of the imports according to Hugo's request.

sarmento:
> > This is looking good. Would you be able to also add a simulation using a quite large bathymetry (e.g. 1000 x 2000), just to test the whole procedure end to end?
> 
> While I'm still working to address the comments in this PR, I ran a few examples. Here is one with a bathymetry of 1025 x 1025:
> 
> Bathymetry:
> 
> bathymetry_random_10
> 
> Simulation output:
> 
>  water_level_random_10.mp4

Ah! this bathymetry seems too smooth :)
Can we try a more radical bathymetry? Also, can we try a different shape (to really see the effect of interpolation), such as fro example 1000 x 400?

sarmento:
I think this is looking very good.
My wish list for this now would be:

1) create a base class Bathymetry, which is what you need to pass to the simulator
2) create a RandomBathymetry class, which extends Bathymetry. You defined all the needed parameters via constructor, and all other methods come from Bathymetry
3) add a new  visualization method to the Bathymetry (base) class, including the contour lines.

Thanks!

fabiocruz:
> Ah! this bathymetry seems too smooth :) Can we try a more radical bathymetry?

Yep. Here is one with resolution 400 x 400.

bathymetry_random_400x400

And here is the corresponding simulation output.

https://github.com/inductiva/inductiva/assets/11545632/ec7523b5-54b9-42d6-a66d-dbc00e9b0a34

fabiocruz:
> Also, can we try a different shape (to really see the effect of interpolation), such as fro example 1000 x 400?

Sure, here is a bathymetry with resolution 1000 x 400:

bathymetry_random_1000x400

And here is the corresponding simulation output.

https://github.com/inductiva/inductiva/assets/11545632/4c4a633e-4a74-47d4-bd12-430e85a5f9ca

PR #423, opened 2023-08-02
Added task.get_time() method

david451lima:
Added task.get_time(), that returns the time the task took to run if it is compelted.

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -4,6 +4,7 @@ import time
 from absl import logging
 from typing import Dict, Any
 from typing_extensions import TypedDict
+from datetime import datetime
 
 from inductiva.client.models import TaskStatusCode
 from inductiva import api
@@ -155,3 +156,22 @@ class Task:
 
     def _get_path_params(self) -> _PathParams:
         return {"task_id": self.id}
+
+    def get_time(self) -> float:
+        """Get the time the task took to complete.
+
+        Returns:
+            The time in seconds.
+        """
+
+        if self.get_status() != TaskStatusCode.SUCCESS:
+            raise RuntimeError("Task is not completed.")
+
+        else:
+            info = dict(self._api.get_task(self._get_path_params()).body)
+            end_time = datetime.strptime(str(info["end_time"]),
+                                         "%Y-%m-%dT%H:%M:%S.%f+00:00")
+            start_time = datetime.strptime(str(info["start_time"]),
+                                           "%Y-%m-%dT%H:%M:%S.%f+00:00")
+
+            return (end_time - start_time).total_seconds()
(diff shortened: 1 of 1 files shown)


inductiva/tasks/task.py
```diff
@@ -155,3 +156,22 @@ class _PathParams(TypedDict):
 
     def _get_path_params(self) -> _PathParams:
         return {"task_id": self.id}
+
+    def get_time(self) -> float:
```
sarmento, line 160: maybe you want to actually name this get_execution_time?
Btw: should you also have other methods related to time? e.g. get_X_time where X can be (@luispcunha?):

* start
* end
* dispatch
*

inductiva/tasks/task.py
```diff
@@ -155,3 +156,22 @@ class _PathParams(TypedDict):
...
+        """Get the time the task took to complete.
+
+        Returns:
+            The time in seconds.
+        """
+
+        if self.get_status() != TaskStatusCode.SUCCESS:
```
IvanPombo, line 167: Shall this be just for successful tasks? I think this may also be useful for other events.

inductiva/tasks/task.py
IvanPombo: I see some potential for this, but there is no foreseeable use for them now.. It will just basically be replicating what David did below.

What we could obtain would be machine information that is also saved inside the task :)

david451lima: Those times are already available to the user, the only difference this method brings is that it computes the difference between two of them (end and start).

IvanPombo (approved):
LGTM!

inductiva/tasks/task.py
```diff
@@ -155,3 +156,22 @@ class _PathParams(TypedDict):
...
+            raise RuntimeError("Task is not completed.")
+
+        else:
+            info = dict(self._api.get_task(self._get_path_params()).body)
+            end_time = datetime.strptime(str(info["end_time"]),
+                                         "%Y-%m-%dT%H:%M:%S.%f+00:00")
+            start_time = datetime.strptime(str(info["start_time"]),
+                                           "%Y-%m-%dT%H:%M:%S.%f+00:00")
```
fabiocruz, line 175: What is the type of `info["end_time"]`? Can you avoid the conversion to string?

fabiocruz (approved):
LGTM. Left a couple of comments that should improve readability.

inductiva/tasks/task.py
fabiocruz: Ok, thank you for clarifying @david451lima. I was wondering if we could avoid the `str` casting, and instead instantiate the datetimes in some other way.

inductiva/tasks/task.py
```diff
@@ -155,3 +156,27 @@ class _PathParams(TypedDict):
 
     def _get_path_params(self) -> _PathParams:
```
sarmento, line 157: Can you comment on this method?
I know you didn't write it, but since it shows up here, I wonder why this is called _get_path_params and then we return a self.id?

PR #424, opened 2023-08-02
Added task.get_machine() method

david451lima:
Added task.get_machine(), that returns the string of the machine in which the task ran.

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -178,3 +178,21 @@ class Task:
                                                 "%Y-%m-%dT%H:%M:%S.%f+00:00")
 
         return (end_time - start_time).total_seconds()
+
+    def get_machine(self) -> str:
+        """Get the machine type used in the task.
+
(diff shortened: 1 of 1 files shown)


inductiva/tasks/task.py
```diff
@@ -178,3 +178,21 @@ def get_execution_time(self) -> float:
                                                 "%Y-%m-%dT%H:%M:%S.%f+00:00")
 
         return (end_time - start_time).total_seconds()
+
+    def get_machine(self) -> str:
```
IvanPombo, line 182: Just to add a bit more: Shall we also provide the ram and number of cores? This is not essential I think for the our current purpose, but it might be also useful for the users.

IvanPombo (approved):
LGTM!

inductiva/tasks/task.py
fabiocruz, line 182: Since you're returning the machine type, should this be named `get_machine_type`?

PR #428, opened 2023-08-03
Allow downloading output files individually

luispcunha:
This PR improves downloading/viewing info about the output files of a simulation/task.
A snippet on how this looks:
```python
task = scenario.simulate(...)

outputs_info = scenario.get_output_files_info()

# Access info from outputs_info properties
full_download_size = outputs_info.size
files = outputs_info.contents

# Print formatted info
print(outputs_info)
# Archive size: 32.96 MiB
# Contents:
#  Size         Compressed   Name        
#  191 B        96 B         stderr.txt  
#  138.24 KiB   16.73 KiB    stdout.txt  
# ...
#  3.85 MiB     3.66 MiB     50/phi      
#  3.80 MiB     3.61 MiB     50/U        
#  1.27 MiB     1.21 MiB     50/p 

path = task.download_outputs(
  filenames=["50/U", "50/p"],
)
# 100%|██████████| 4.82M/4.82M [00:00<00:00, 273MiB/s]
````

### CODE REVIEW

File: inductiva/tasks/task.py [hunk 1]
```diff
@@ -152,25 +157,91 @@ def get_output(self, output_dir=None):
...
 
         if self._output_class:
             return self._output_class(output_dir)
 
         return output_dir
 
-    def download_output(self, output_dir=None) -> pathlib.Path:
-        """Download the output of the task.
+    def get_outputs_info(self):
+        """Get information of the output files of the task.
```
IvanPombo, line 169: I think we need to think about this name `get_outputs_info` since with this change the `get_output` may just fetch some files of the output for each scenario.

IvanPombo (approved):
LGTM!

inductiva/tasks/task.py
luispcunha: `get_output_files_info` ?

inductiva/tasks/task.py
```diff
@@ -152,25 +157,91 @@ def get_output(self, output_dir=None):
...
 
         Returns:
-            The path to the downloaded output directory.
+            An instance of the OutputContents class, which can be used to
+            access info about the output files, such as the size of archive,
```
cristianacarpinteiro, line 173: size of the archive or archive's size?

inductiva/tasks/task.py
```diff
@@ -152,25 +157,91 @@ def get_output(self, output_dir=None):
...
+            number of files, and information about each file (name, size,
+            compressed size). It can also be used to print that information
+            in a formatted way.
```
cristianacarpinteiro, line 176: What do you mean by " It can also be used to print that information in a formatted way."? Isn't it always printed in a formatted way?

cristianacarpinteiro (approved):
This looks very nice and it will be very helpful! Thank you @luispcunha :)

PR #429, opened 2023-08-03
Simsopt Simulator class

joaobernardo23:


inductiva/stellarators/simulators/simsopt.py
```diff
@@ -0,0 +1,49 @@
+"""Simsopt module of the API."""
+from typing import Optional
+from uuid import UUID
+
+from inductiva import types, tasks
+from inductiva.simulation import Simulator
+
+
+class Simsopt(Simulator):
+    """Class to invoke a generic Simsopt simulation on the API."""
```
fabiocruz, line 10: Simplify, "Invokes a simsopt simulation on the API."

inductiva/stellarators/simulators/simsopt.py
```diff
@@ -0,0 +1,49 @@
+"""Simsopt module of the API."""
+from typing import Optional
+from uuid import UUID
+
+from inductiva import types, tasks
+from inductiva.simulation import Simulator
+
+
+class Simsopt(Simulator):
```
fabiocruz, line 9: I think we should adapt this name in the future to reflect the specific use-case we're using simsopt for but for now this should be fine.

fabiocruz (approved):
LGTM!

inductiva/stellarators/simulators/simsopt.py
joaobernardo23: Noted!

PR #430, opened 2023-08-03
Add FEniCSx simulator

Joao97ribeiro:


inductiva/structures/simulators/fenicsx.py
```diff
@@ -0,0 +1,40 @@
...
+        self.api_method = api_method
+
+    @property
+    def api_method_name(self) -> str:
+        return self.api_method
+
+    def run(
+        self,
+        input_dir: types.Path,
+        commands: List[dict],
```
fabiocruz, line 24: Your simulator doesn't need a list of commands. It runs based on a script.

inductiva/structures/simulators/fenicsx.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(self, api_method: str = "fem.fenicsx.run_simulation"):
+        super().__init__()
+        self.api_method = api_method
+
+    @property
+    def api_method_name(self) -> str:
+        return self.api_method
+
+    def run(
```
fabiocruz, line 21: This is missing a few parameters to identify the files containing the inputs necessary to run the simulator on the API, e.g. `mesh_filename`, `bc_filename`, `material_filename`.

inductiva/structures/simulators/fenicsx.py
```diff
@@ -0,0 +1,40 @@
...
+        self,
+        input_dir: types.Path,
+        commands: List[dict],
+        resource_pool_id: Optional[UUID] = None,
+        n_cores: int = 1,
```
fabiocruz, line 26: Do you know how to set the script you wrote on the backend to run in parallel, with a configurable number of cores?

fabiocruz (changes requested):
Thanks @Joao97ribeiro. This needs some (easy) work! Comments below.

inductiva/structures/simulators/fenicsx.py
Joao97ribeiro: "I've never done something similar before. Do you have any examples I can follow?

fabiocruz: The way this is done varies with the simulator, I'd have to figure out how to do this in FEniCSx. Let's remove this parameter for now, and deal with this in the future, if/when needed.

inductiva/structures/simulators/fenicsx.py
```diff
@@ -20,21 +20,25 @@ def api_method_name(self) -> str:
 
     def run(
         self,
-        input_dir: types.Path,
```
fabiocruz: `input_dir` is a mandatory parameter in this method, so don't remove it please! Simply add the new parameters `mesh_filename`, ...

inductiva/structures/simulators/fenicsx.py
```diff
@@ -20,21 +20,25 @@ def api_method_name(self) -> str:
...
-        commands: List[dict],
+        mesh_path: types.Path,
+        bc_path: types.Path,
+        material_path: types.Path,
```
fabiocruz, line 25: Make these refer to the filenames and not paths. The idea is that these files will live inside the same directory, `input_dir`.

Unless you think it's useful to load them from different directories? Is that the workflow you expect?

fabiocruz (changes requested):

inductiva/structures/simulators/fenicsx.py
Joao97ribeiro: make sense, I will remove this parameter! thanks

inductiva/structures/simulators/fenicsx.py
Joao97ribeiro: The challenge lies in situations where you already possess the meshes dataset, bc dataset, and material dataset, but you want to avoid duplicating each file in multiple directories. Otherwise, the same data would be copied across different directories, especially if the same mesh is being used. What do you believe would be the most effective approach to address this issue? One possible solution could be to selectively copy only the required files from the mesh, bc, and material datasets for each simulation, which might indeed be the best way to handle this situation.

inductiva/structures/simulators/fenicsx.py
Joao97ribeiro: I was considering starting with the material, mesh, and bc datasets and then running the simulations. However, if it makes things easier, we can opt to copy the required files instead

Joao97ribeiro: what do you think?

inductiva/structures/simulators/fenicsx.py
```diff
@@ -0,0 +1,44 @@
...
+
+from typing import Optional, List
+from uuid import UUID
+
+from inductiva import types, tasks
+from inductiva.simulation import Simulator
+
+
+class FEniCSx(Simulator):
+    """Class to invoke a generic FEniCSx simulation on the API."""
```
fabiocruz, line 11: In the future, we should name this differently to reflect the fact that the method we run on the API side is not for a generic simulation. This is ok for now, though, as we need to give this naming some more thought.

inductiva/structures/simulators/fenicsx.py
```diff
@@ -0,0 +1,44 @@
...
+    def api_method_name(self) -> str:
+        return self.api_method
+
+    def run(
+        self,
+        mesh_path: types.Path,
+        bc_path: types.Path,
+        material_path: types.Path,
+        resource_pool_id: Optional[UUID] = None,
+        n_cores: int = 1,
```
fabiocruz, line 27: The `n_cores` parameter is still here, remember to remove it.

fabiocruz (changes requested):

inductiva/structures/simulators/fenicsx.py
```diff
@@ -0,0 +1,44 @@
+"""FEniCSx module of the API for finite element analysiss."""
+
+from typing import Optional, List
+from uuid import UUID
+
+from inductiva import types, tasks
```
sarmento, line 6: Try to follow Inductiva's guideline on imports.
Check Hugo's email.
It goes along these lines:

https://google.github.io/styleguide/pyguide.html#22-imports

Screenshot 2023-08-01 at 12.48.53.png

So, according to the guidelines this is NOT RECOMMENDED:

Screenshot 2023-08-01 at 12.53.08.png
And the RECOMMENDED version is below:

Screenshot 2023-08-01 at 12.51.24.png


Note: other different Python guidelines are ok with importing classes, but not the one we are following, so let's stick to it.

sarmento (commented):
Happy to see this being included in the API :)

inductiva/structures/simulators/fenicsx.py
Joao97ribeiro: I agree, we can implement it this way for now, and when other cases arise, we can modify it as per your suggestion. Thank you, Fabio!

inductiva/structures/simulators/fenicsx.py
Joao97ribeiro: Thanks Luis, for the input! I will make the necessary changes to adhere the Google style and Hugo's recommendations

fabiocruz (approved):
LGTM!

PR #431, opened 2023-08-03
Fix urllib3 dependency

luispcunha:
This fixes a bug some of you faced (cc. @fabiocruz @david451lima). I was using a urllib3 v2 feature but our api client requires urllib3 v1. I updated the dependencies with the requirements of the generated API client and removed the usage of that v2 feature.


luispcunha:
> This seems to contain new features, and not just the fix you mention in the PR description. Did you mean to push those here, too?

Sorry.

PR #432, opened 2023-08-03
Simulate function

joaobernardo23:


inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -185,8 +189,40 @@ def from_curves_file(cls,
...
+            resource_pool_id: The resource pool to use for the simulation.
+        """
+        coil_coefficients = [coil.curve_coefficients for coil in self.coils]
+        coil_currents = np.array([coil.current for coil in self.coils])
+
+        coil_coefficients_filename = 'coil_coefficients.npz'
+        coil_currents_filename = 'coil_currents.npz'
+
+        np.savez(coil_coefficients_filename, *coil_coefficients)
+        np.savez(coil_currents_filename, coil_currents)
```
fabiocruz, line 213: Move this to a method `create_input_files`. See other scenarios to see exactly how this should be done.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -13,20 +20,17 @@ class StellaratorCoils(Scenario):
...
-        (independent coils).
         coils (list): List of `Coil` objects.
         num_field_periods (int): Number of magnetic field periods.
-        Refers to the number of complete magnetic field repetitions 
-        within a stellarator. Represents how many times the magnetic
-        field pattern repeats itself along the toroidal direction.
+          Refers to the number of complete magnetic field repetitions 
+          within a stellarator. Represents how many times the magnetic
+          field pattern repeats itself along the toroidal direction.
     """
```
fabiocruz, line 29: You're missing an attribute `valid_simulators = [Simsopt]` in this class.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -185,8 +189,40 @@ def from_curves_file(cls,
...
+
+        Args:
+            simulator: The simulator to use for the simulation.
+            resource_pool_id: The resource pool to use for the simulation.
+        """
+        coil_coefficients = [coil.curve_coefficients for coil in self.coils]
+        coil_currents = np.array([coil.current for coil in self.coils])
+
+        coil_coefficients_filename = 'coil_coefficients.npz'
+        coil_currents_filename = 'coil_currents.npz'
```
fabiocruz, line 210: Define these in global constants. See other scenarios to see how this should be done.

fabiocruz (changes requested):
Thanks @joaobernardo23! The code is clean, but it needs to be rearranged such that the files describing the coefficients and coils are created in a separate method.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -185,7 +196,35 @@ def from_curves_file(cls,
 
         return cls(coils, num_field_periods)
 
-    def simulate(self):
+    def simulate(
+        self,
+        simulator: Simulator = Simsopt(),
+        resource_pool_id: Optional[UUID] = None,
+        run_async: bool = False,
+        magnetic_field_filename: str = 'magnetic_field.npz',
```
fabiocruz, line 204: What is this parameter supposed to control? If it's the name of the file created by the simulator on the backend, then this shouldn't be controllable from the client. That file has a default name defined by the executer (or the underlying "simulator").

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -185,7 +196,35 @@ def from_curves_file(cls,
...
+        plasma_surface_filename: str = 'input.QA',
```
fabiocruz, line 205: I don't think this should have a default.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -185,7 +196,35 @@ def from_curves_file(cls,
...
+    ) -> tasks.Task:
+        """Simulates the scenario.
+
+        Args:
+            simulator: The simulator to use for the simulation.
+            resource_pool_id: The resource pool to use for the simulation.
```
fabiocruz, line 211: This is missing the description of some args, at least `plasma_surface_filename`.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -185,7 +196,35 @@ def from_curves_file(cls,
...
+        """
+
+        task = super().simulate(
+            simulator,
+            resource_pool_id=resource_pool_id,
+            run_async=run_async,
+            magnetic_field_filename=magnetic_field_filename,
```
fabiocruz, line 218: I don't think this should be here, see my comment above.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -237,3 +276,18 @@ def get_circular_curve_coefficients(toroidal_angle, major_radius, minor_radius):
...
 
     return curve_coefficients
+
+
+@StellaratorCoils.create_input_files.register
+def _(self, simulator: Simsopt, input_dir):  # pylint: disable=unused-argument
+    """Creates Simsopt simulation input files."""
+
+    coil_coefficients = [coil.curve_coefficients for coil in self.coils]
+    coil_currents = np.array([coil.current for coil in self.coils])
```
fabiocruz, line 286: Why are you converting one to `np.array` and not the other?

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -237,3 +276,18 @@ def get_circular_curve_coefficients(toroidal_angle, major_radius, minor_radius):
...
+
+    coil_coefficients_filename = os.path.join(input_dir,
+                                              COIL_COEFFICIENTS_FILENAME)
+    coil_currents_filename = os.path.join(input_dir, COIL_CURRENTS_FILENAME)
+
+    np.savez(coil_coefficients_filename, *coil_coefficients)
+    np.savez(coil_currents_filename, coil_currents)
```
fabiocruz, line 293: Why saving `*to_be_saved` in one case and not the other?

fabiocruz (changes requested):
A few more changes needed, @joaobernardo23, but this is looking almost ready!

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -1,8 +1,20 @@
...
 import random
 import numpy as np
+import os
+
+from typing import Optional
+from uuid import UUID
+
+from functools import singledispatchmethod
+from inductiva import tasks
+from inductiva.stellarators.simulators import Simsopt
```
sarmento, line 12: Please fix the imports to follow Inductiva's guidelines. Check Hugo's email.
You should not directly import classes.
Remember that note I sent about readability?

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -1,8 +1,20 @@
...
+from inductiva.simulation import Simulator
+from inductiva.scenarios import Scenario
+
+COIL_COEFFICIENTS_FILENAME = 'coil_coefficients.npz'
+COIL_CURRENTS_FILENAME = 'coil_currents.npz'
 
 
 class StellaratorCoils(Scenario):
```
sarmento, line 20: I am still not sure about the naming here...

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -13,20 +25,19 @@ class StellaratorCoils(Scenario):
...
-        num_coils (int): The number of coils per field period 
-        (independent coils).
         coils (list): List of `Coil` objects.
         num_field_periods (int): Number of magnetic field periods.
-        Refers to the number of complete magnetic field repetitions 
-        within a stellarator. Represents how many times the magnetic
-        field pattern repeats itself along the toroidal direction.
+          Refers to the number of complete magnetic field repetitions 
+          within a stellarator. Represents how many times the magnetic
+          field pattern repeats itself along the toroidal direction.
```
sarmento, line 32: Do you mind adding a bit more info here? What is the typical number? 
Try to be a bit more pedagogic, and maybe point to the literature (here or in the class description).

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -185,7 +196,35 @@ def from_curves_file(cls,
...
+            simulator: The simulator to use for the simulation.
+            resource_pool_id: The resource pool to use for the simulation.
+        """
+
+        task = super().simulate(
+            simulator,
+            resource_pool_id=resource_pool_id,
+            run_async=run_async,
+            magnetic_field_filename=magnetic_field_filename,
+            coil_coefficients_filename=COIL_COEFFICIENTS_FILENAME,
```
sarmento, line 219: I am struggling a bit with the naming of these variables. Their value seems to be simulator dependent, but they. have been declared as global variable for the scenario. Maybe this is common practice around scenarios but I would expected this to be named in a way that, at least, refers the corresponding simulator. E.g.

SIMOPT_DEFAULT_COIL_COEFFICIENTS_FILENAME

Ideally, these would belong to a structure or config file related with the simulator itself, but if we are including them in this file as global variable, there should be, at the very least, a mention to the simulator in their name.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -203,7 +242,7 @@ class Coil:
 
     Attributes: 
         curve_coefficients (np.ndarray): Array with Fourier coefficients 
-        defining the coil.
+          defining the coil.
```
sarmento, line 245: I am assuming is a 2D array (real + imaginary part). Do you want to be more explicit about this? Also, this has a size to? Has it been defined before (e.g. in the constructor) at some point so you can "check" that all is correct?

inductiva/stellarators/scenarios/coils/coils.py
joaobernardo23: Because coil_coefficients is a list of numpy arrays which I can immediately write to a file with numpy.savez whereas coil_currents is a list with the currents (list of floats), because each object of the `Coil` class has curve_coefficients, which is a numpy array, and the current which is merely a float.

inductiva/stellarators/scenarios/coils/coils.py
joaobernardo23: So I will remove it has a simulator argument and that means I will also need to remove self.args.magnetic_field_filenamen from the API, right?

inductiva/stellarators/scenarios/coils/coils.py
joaobernardo23: Because `coil_coefficients` is a list of numpy arrays and `coil_currents` is just one numpy array. That was the way we decided to construct the classes initially.

inductiva/stellarators/scenarios/coils/coils.py
joaobernardo23: I believe that it is fairly well described in the class Docstring just above that. There I have the whole explanation about the Fourier Series.

inductiva/stellarators/scenarios/coils/coils.py
joaobernardo23: I agree but it there a way to avoid the Pylint error then?

inductiva/stellarators/scenarios/coils/coils.py
fabiocruz: Yep. It can have a default name controlled by the API.

inductiva/stellarators/scenarios/coils/coils.py
fabiocruz: Got it thanks.

inductiva/stellarators/scenarios/coils/coils.py
joaobernardo23: Unfortunately the Scenario super class does not have `*args` in the simulate function so I think I will have to leave it with a default name.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -8,18 +8,37 @@
...
 
 
-class StellaratorCoils(Scenario):
+class StellaratorCoils(scenarios.Scenario):
     """Represents stellarator coils.
 
+    A stellarator is a magnetic fusion confinement device that possesses a 
+    set of external electromagnetic coils that have current running through 
+    them. These coils create a magnetic field that is able to confine a plasma, 
+    at extremely high temperatures, where the nuclear fusion reaction take
```
fabiocruz, line 23: Typo, "reactions"

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -8,18 +8,37 @@
...
     """Represents stellarator coils.
 
+    A stellarator is a magnetic fusion confinement device that possesses a 
+    set of external electromagnetic coils that have current running through 
+    them. These coils create a magnetic field that is able to confine a plasma, 
+    at extremely high temperatures, where the nuclear fusion reaction take 
+    place.
+
+    When designing a stellarator, one only needs to worry about a certain set of
+    NC independent coils, due to the symmetries involved. There is stellarator
```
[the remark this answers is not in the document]
joaobernardo23: But to convey the right idea, maybe I can write: When designing a stellarator, as it pretends to the number of coils used, one only needs ...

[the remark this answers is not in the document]
fabiocruz: Maybe add context: "Designing a stellarator involves defining a set of coils that... These coils are defined by a small set of NC independent coils, due to..."

Issue #412, opened 2023-07-27
URL connection pool warning


IvanPombo:
Lately, when running simulations synchronously we have been receiving the following warnings:

<img width="1424" alt="Screenshot 2023-07-27 at 12 23 16" src="https://github.com/inductiva/inductiva/assets/104431973/c0c9e65b-b8a7-4d75-838d-3aefc52c61e6">

The simulations run sucessfully, but the logs get full with it.

PR #433, opened 2023-08-03
Disable urllib3 warnings

luispcunha:
This disables the urllib3 retry warning. As it seems to not affect the behavior, I'll suppress the warning but not close the issue while we don't fix the root cause of the warning.

Issue #412.


begaiym-k (approved):
As a temporary solution this seems good. We can take a closer look later on.

PR #435, opened 2023-08-04
Allow default files to be downloaded per scenario

IvanPombo:
This PR adds the capability for each scenario to download a set of default files from the simulation outputs. The implementation was performed in a general sense and already applied to the WindTunnel scenario.

Now, when the task arises from a scenario simulation, `task.get_output()` gives these specific default files.

This connects with the backend implementation of the post-processing, and therefore it refactors a bit the WindTunnel scenario.

Next Step: Refactor the `WindTunnelOutput` to be able to handle these files or the general downloaded files.


inductiva/tasks/task.py
```diff
@@ -153,14 +153,36 @@ def set_output_class(self, output_class):
...
+            uncompress: Whether to uncompress the archive after downloading it.
+            rm_archive: Whether to remove the archive after uncompressing it.
+                If uncompress is False, this argument is ignored.
+
         Returns:
             The output of the task.
         """
         self.wait()
-        output_dir = self.download_outputs(output_dir)
+        output_dir = self.download_outputs(filenames=self._default_files_list,
```
cristianacarpinteiro, line 182: This is great! I will implement something like this for the protein solvation,

cristianacarpinteiro (approved):
LGTM! Thanks @IvanPombo

PR #436, opened 2023-08-04
scenarios - plate with holes - linear elasticity

Joao97ribeiro:


inductiva/structures/scenarios/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -0,0 +1,112 @@
...
+from geometry_utils import CircularHole
+from geometry_utils import EllipticalHole
+from geometry_utils import GeometricCase
+from geometry_utils import RectangularHole
+from geometry_utils import RectangularPlate
+from material_utils import IsotropicLinearElasticMaterial
+from mesh_utils import Mesh
+
+
+class PlateLinearElastic(scenarios.Scenario):
```
fabiocruz, line 23: I think we need an easier name for this, e.g. `DeformablePlate`.

inductiva/structures/scenarios/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -0,0 +1,112 @@
...
+    boundary conditions, and material properties. The plate and holes together
+    form the geometry, which is used to create the mesh for the simulation.
+    """
+
+    valid_simulators = [structures.simulators.FEniCSx]
+
+    def __init__(
+        self,
+        plate: RectangularPlate,
+        holes: List[CircularHole, RectangularHole or EllipticalHole],
```
fabiocruz, line 36: The right type annotation here would be `List[Union[CircularHole, RegangularHole, EllipticalHole]]`. Or, better yet, `List[Hole]`, where `Hole` would be the super class of the different hole classes.

inductiva/structures/scenarios/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -0,0 +1,112 @@
...
+        self.plate = plate
+        self.holes = holes
+        self.bcs = bcs
+        self.material = material
+
+        self.geometry = GeometricCase(plate_obj=plate, list_holes_objs=holes)
+
+    def simulate(
+        self,
+        simulator: simulation.Simulator = structures.simulators.FEniCSx,
```
fabiocruz, line 59: `()` missing at the end, `....FEniCSx()`.

inductiva/structures/scenarios/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -0,0 +1,112 @@
...
+        resource_pool_id: Optional[UUID] = None,
+        run_async: bool = False,
+        simulation_time=300,
+        output_time_step=10,
```
fabiocruz, line 63: This isn't relevant in this scenario, remove it.

inductiva/structures/scenarios/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+@PlateLinearElastic.create_input_files.register
+def _(self,
+      simulator: structures.simulators.FEniCSx,
+      input_dir: types.Path,
+      geometry_filename: str = "geometry.json",
+      mesh_filename: str = "mesh.xdmf",
+      bcs_filename: str = "bcs.json",
+      material_filemane: str = "material.json") -> None:
```
fabiocruz, line 94: These arguments can't be here. The names of these files should instead be defined by global constants in this file. See other scenarios for a reference.

inductiva/structures/scenarios/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -0,0 +1,112 @@
...
+        commands = self.get_commands()
+
+        task = super().simulate(simulator,
+                                resource_pool_id=resource_pool_id,
+                                run_async=run_async,
+                                commands=commands)
+        return task
+
+
+@PlateLinearElastic.create_input_files.register
```
fabiocruz, line 87: Remember to create a method `create_input_files` on the class. See other scenarios for a reference.

inductiva/structures/scenarios/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -0,0 +1,112 @@
...
+from inductiva import scenarios
+
+from bc_utils import BoundaryConditionsCase
+from geometry_utils import CircularHole
+from geometry_utils import EllipticalHole
+from geometry_utils import GeometricCase
+from geometry_utils import RectangularHole
+from geometry_utils import RectangularPlate
+from material_utils import IsotropicLinearElasticMaterial
+from mesh_utils import Mesh
```
fabiocruz, line 20: Remember that you shouldn't import classes directly. Import the submodules instead.

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,276 @@
+"""Utils to create the mesh."""
+
+import time
+from typing import List, Optional
+
+import dolfinx
+from dolfinx.io import gmshio
```
fabiocruz, line 7: This cannot depend on `dolfinx`, since that's the hard dependency that our API solves in this case. In other words, the code that uses `dolfinx` should all be on the API side.

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,276 @@
...
+import gmsh
+from mpi4py import MPI
+import numpy as np
+
+import geometry_utils
+
+
+class Mesh:
+    """DOLFINx mesh.
```
fabiocruz, line 16: Shouldn't have a reference to `dolfinx`.

inductiva/structures/scenarios/plate_linear_elastic/material_utils.py
fabiocruz, line 1: This file should also expose a few standard materials, e.g. `COPPER = IsotropicLinearElasticMaterial(...)` and others, such that we can then use them routinely without having to remember material properties.

inductiva/structures/scenarios/plate_linear_elastic/material_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    def to_dict(self) -> dict:
+        """Convert Material properties to a dictionary.
+
+        Returns:
+            material_dict: A dictionary containing the Material properties.
+        """
+        young_modulus_dict = {"young_modulus": self.young_modulus}
+        poisson_ratio_dict = {"poisson_ratio": self.poisson_ratio}
+
+        material_dict = {**young_modulus_dict, **poisson_ratio_dict}
```
fabiocruz, line 28: Why is this overcomplicating?

```python
material_dict = {
    "young_modulus": self.young_modulus,
    "poisson_ratio": self.poisson_ratio,
}
```

inductiva/structures/scenarios/plate_linear_elastic/material_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+import json
+
+
+class IsotropicLinearElasticMaterial:
+    """Linear elastic isotropic material.
+
+    Attributes:
+        young_modulus: A float representing the Young's modulus.
+        poisson_ratio: A float representing the Poisson's ratio.
```
fabiocruz, line 11: No need for the bits "A float representing". Add "... of the material" at the end, though.

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,413 @@
...
+import json
+import math
+
+
+class RectangularPlate:
+    """Rectangular plate.
+
+    Attributes:
+        plate_length: A float representing the plate length.
+        plate_width: A float representing the plate width.
```
fabiocruz, line 15: These are properties of the plate, so no need for the `plate_` prefix. Use `width` and `height` only.

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,105 @@
+"""Utils to create the different types of boundary conditions."""
+
+from typing import Optional, List
+
+import json
+
+
+class Dirichlet:
```
fabiocruz, line 8: Name this as `DirichletBC`, to emphasize that it represents boundary conditions. Same for Neumann.

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,105 @@
...
+        displacement_x: A float representing imposed displacement in
+          x-direction. Only fill it if want to resprtit the displamente. 
+          Fill this attribute if you want to specify the displacement.
+        displacement_y: A float representing imposed displacement in
+          y-direction. Only fill it if want to resprtit the displamente.
+          Fill this attribute if you want to specify the displacement.
+    """
+
+    def __init__(self,
+                 boundary_name: str,
```
fabiocruz, line 24: As per the available options, the type annotation here should be `Literal["left", "top", ...]`.

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,105 @@
...
+        self.boundary_name = boundary_name
+        self.tension_x = tension_x
+        self.tension_y = tension_y
+
+
+class BoundaryConditionsCase:
+    """Boundary conditions for a given case.
+
+    The boundary conditions for a given case are characterized by an ID, a
+      list of Dirichlet and Tension boundary conditions.
```
fabiocruz, line 56: No need for the ID.

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,105 @@
...
+                neumann_object_dict = vars(neumann_object)
+                tension_dicts_list.append(neumann_object_dict)
+            neumann_dict = {"neumann": tension_dicts_list}
+
+            # Merge dictionaries: boundary conditions dictionary
+            bcs_dict = {**bcs_dict, **neumann_dict}
+
+        # Write JSON file
+        with open(json_path, "w", encoding="utf-8") as write_file:
+            json.dump(bcs_dict, write_file, indent=4)
```
fabiocruz, line 105: This is still structuring BC based on their type. This is ok for now, but we should quickly aim at structuring them based on their name.

fabiocruz (changes requested):
Thank you @Joao97ribeiro! We're fast approaching a usable scenario ;) I do think there are a number of things to improve here still, so I am leaving a first round of comments below. Let me know if you have questions.

Joao97ribeiro:
Thank you for the feedback, @fabiocruz! I've incorporated all of your comments, and I believe it's now ready :D Let m know wdyt, thanks

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+
+        # Holes dictionary
+        if self.list_holes_objects is not None:
+
+            all_hole_properties = []
+            for hole_object in self.list_holes_objects:
+                hole_dictionary = vars(hole_object)
+                all_hole_properties.append(hole_dictionary)
+
+            holes_dict = {"holes": all_hole_properties}
```
fabiocruz, line 144: This line could be as simple as `holes_dict = {"holes": [hole.to_dict() for hole in self.holes]}`

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+
+            # Merge dictionaries: geometric case dictionary
+            geom_case_dictionary = {**plate_dict, **holes_dict}
+        else:
+            geom_case_dictionary = {**plate_dict}
```
fabiocruz, line 149: You can avoid this if-else block. Just populate the list of holes with `[]` if you get `None` for it on `__init__`. Then the dictionary of the geometry case always has a "holes" key, that may or not be empty.

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+
+    def write_to_json(self, json_path: str) -> None:
+        """Writes the geometric case to JSON file.
+
+        Args:
+            json_path (str): The JSON file path.
+        """
+
+        # Plate dictionary
+        plate_dict = {"plate": vars(self.plate_object)}
```
fabiocruz, line 134: The same as above applies here. The plate class should have a `to_dict` method.

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+
+        Raises:
+            ValueError: If there is no plate object attributed to the 
+              geometric case.
+
+        Returns:
+          tuple: The plate gmsh object and the IDs of the curves that constitute
+            the plate.
+        """
+        if self.plate_object is not None:
```
fabiocruz, line 166: Why would this be `None`?

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+                dy=self.plate_object.length)
+
+            gmsh.model.occ.synchronize()
+
+            plate_curves_id = []
+
+            for curve_id in gmsh.model.getBoundary([(2, plate_gmsh)]):
+                plate_curves_id.append([curve_id[1]][0])
+
+            return plate_gmsh, plate_curves_id
```
fabiocruz, line 183: This should be a method of the `RectangularPlate` class!

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+                            ry=hole_object.radius_x)
+                        gmsh.model.occ.rotate(
+                            dimTags=[(2, hole_gmsh)],
+                            x=hole_object.center_x,
+                            y=hole_object.center_y,
+                            z=0,
+                            ax=0,
+                            ay=0,
+                            az=1,
+                            angle=math.radians(hole_object.angle + 90))
```
fabiocruz, line 302: These should be methods of the hole classes. Then all you'd do would be `hole._to_opencascade_cad()`... (or `_to_occ()` for short).

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+                                     "circular, rectangular, or elliptical. "
+                                     "Please check.")
+
+                holes_mesh_offset.append(hole_mesh_offset)
+                holes_predefined_element_size.append(
+                    hole_predefined_element_size)
+
+            return holes_mesh_offset, holes_predefined_element_size
+
+    def to_gmsh(self) -> tuple:
```
fabiocruz, line 388: `to_gmsh_occ()`?

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+        """
+
+        # 1. Converts the plate object to OpenCASCADE CAD representation
+        plate_gmsh, plate_curves_id = self._plate_to_opencascade_cad()
+
+        # 2. Converts the hole objects to OpenCASCADE CAD representation
+        holes_gmsh, holes_curve_id = self._holes_to_opencascade_cad()
+
+        # 3. Removes holes from the plate in the OpenCASCADE CAD representation
+        for hole in holes_gmsh:
```
fabiocruz, line 414: `for hole_gmsh in holes_gmsh:`

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+        if self.plate_object is not None:
+
+            # Rectangular plate
+
+            plate_mesh_offset = min(self.plate_object.width,
+                                    self.plate_object.length) / 2
+            plate_predefined_element_size = (self.plate_object.width * 2 +
+                                             self.plate_object.length * 2) / 4
+
+            return plate_mesh_offset, plate_predefined_element_size
```
fabiocruz, line 218: Again, this should be a method of the `RectangularPlate` class.

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+                elif hole_object.hole_type == "elliptical":
+
+                    hole_mesh_offset = min(hole_object.radius_x,
+                                           hole_object.radius_y)
+                    hole_predefined_element_size = math.pi * (
+                        3 * (hole_object.radius_x + hole_object.radius_y) -
+                        math.sqrt(
+                            (3 * hole_object.radius_x + hole_object.radius_y) *
+                            (hole_object.radius_x + 3 * hole_object.radius_y))
+                    ) / 4
```
fabiocruz, line 375: Think about this. They could be methods of the corresponding hole classes...

inductiva/structures/scenarios/plate_linear_elastic/material_utils.py
```diff
@@ -0,0 +1,56 @@
...
+            json_path (str): JSON file path.
+        """
+        material_dict = self.to_dict()
+
+        # Write JSON file
+        with open(json_path, "w", encoding="utf-8") as write_file:
+            json.dump(material_dict, write_file, indent=4)
+
+
+# Define standard materials
```
fabiocruz, line 44: 👌

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,257 @@
+"""Utils to create the mesh."""
+
+from typing import List
+from typing import Optional
```
fabiocruz, line 4: Same line.

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,257 @@
...
+        glb_ref_fac (float): The refinement factor for global refinement of the
+          mesh. 
+        loc_ref_fac (float): The refinement factor for local refinement of  the
+          mesh.
+    """
+
+    def __init__(self,
+                 geometry: geometry_utils.GeometricCase,
+                 glb_ref_fac: Optional[float] = 1.0,
+                 loc_ref_fac: Optional[float] = 0.0) -> None:
```
fabiocruz, line 25: Be more expressive in the names of these attributes, `global_refinement_factor`, `local_refinement_factor`. I know they are lengthy, but that's fine. They are clear too!

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,257 @@
...
+
+class Mesh:
+    """Mesh.
+
+    Attributes:
+        geometry: A geometry_utils.GeometricCase object.
+        glb_ref_fac (float): The refinement factor for global refinement of the
+          mesh. 
+        loc_ref_fac (float): The refinement factor for local refinement of  the
+          mesh.
```
fabiocruz, line 19: Say a few more words about these refinement factors. What do they control in specific?

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,257 @@
...
+                                    [plate_curves_id[id_plate_boundary]], 100)
+
+            mesh_field_id += 1
+            add_mesh_field_threshold(mesh_field_id, mesh_field_id - 1,
+                                     global_mesh_size_max, global_mesh_size_max,
+                                     0, plate_mesh_offset)
+            mesh_field_threshol_id.append(mesh_field_id)
+
+        # Check if local refinement is required
+        if self.loc_ref_fac != 0.0:
```
fabiocruz, line 199: Check for `> 0` instead.

fabiocruz (changes requested):
Thanks @Joao97ribeiro. There's a bunch of methods that need to be moved to the classes representing the holes and plate. Other than that, the code is looking good!

sarmento:
> Thanks @Joao97ribeiro. There's a bunch of methods that need to be moved to the classes representing the holes and plate. Other than that, the code is looking good!

Thank you @Joao97ribeiro for this work and @fabiocruz for your detailed review.

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+class EllipticalHole(Hole):
+    """Elliptical hole.
+
+    Attributes:
+        center_x (float): x-coordinate of the center.
+        center_y (float): y-coordinate of the center.
+        radius_x (float): Radius of the hole in the x-direction.
+        radius_y (float): Radius of the hole in the y-direction.
```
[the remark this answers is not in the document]
Joao97ribeiro: Thanks, I can proceed to implement them right away; the changes are straightforward and uncomplicated!

Joao97ribeiro:
> Thanks @Joao97ribeiro. There's a bunch of methods that need to be moved to the classes representing the holes and plate. Other than that, the code is looking good!

Thanks @fabiocruz for the suggestions and comments. The major change is in the geometry utils. As you recommended, I had to move many functions, which significantly impacted the file:

https://github.com/inductiva/inductiva/blob/83df42f4569c65b0cec9292540399c67c308d3f8/inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py

Please share your thoughts :)

Joao97ribeiro:
> > Thanks @Joao97ribeiro. There's a bunch of methods that need to be moved to the classes representing the holes and plate. Other than that, the code is looking good!
> 
> Thank you @Joao97ribeiro for this work and @fabiocruz for your detailed review.

We're almost there, @sarmento! @fabiocruz has been a great help: providing excellent and rapid code reviews. Thanks!

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,614 @@
...
+
+    Attributes:
+        plate_type (str): Plate type.
+        length (float): Plate length.
+        width (float): Plate width.
+    """
+
+    def __init__(self, length: float, width: float) -> None:
+        """Initializes a RectangularPlate object."""
+        self.plate_type = "rectangular"
```
fabiocruz, line 51: I think this attribute was unintentionally left here. It isn't needed, so it should be removed.

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,614 @@
...
+        Returns:
+            List[int]: A list of boundary IDs corresponding to the plate's 
+              boundaries.
+        """
+        plate_boundaries = gmsh.model.getBoundary([(2, plate_gmsh)])
+
+        return [[boundary_id[1]][0] for boundary_id in plate_boundaries]
+
+    def get_mesh_metrics(self) -> Tuple[float, float]:
+        """Gets the metrics for mesh generation in Gmsh.
```
fabiocruz, line 106: The naming here is a bit odd, since you're not getting metrics of the "mesh" itself, but rather of the plate. You can leave it as is for now, and we can revisit this in the future.

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,614 @@
...
+            hole_gmsh (int): The Gmsh entity ID representing the hole's 
+              OpenCASCADE CAD representation.
+
+        Returns:
+            List[int]: A list of boundary IDs corresponding to the hole's 
+              boundaries.
+        """
+        hole_boundaries = gmsh.model.getBoundary([(2, hole_gmsh)])
+
+        return [[boundary_id[1]][0] for boundary_id in hole_boundaries]
```
fabiocruz, line 227: The same (about this being better a function in mesh_utils.py) applies here. It even is the same code!

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,614 @@
...
+            CAD representation for each hole.
+            - List of lists, where each inner list contains the boundary IDs 
+            corresponding to the hole's boundaries for each hole.
+        """
+
+        holes_gmsh = []
+        holes_boundary_ids = []
+
+        for hole in self.holes:
+            hole_gmsh = hole.to_occ()
```
fabiocruz, line 522: See how clean this has become?

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,268 @@
...
+            gmsh.model.mesh.field.setNumber(mesh_field_id, "InField",
+                                            ref_mesh_field_id)
+
+            # Set the size constraints for the elements.
+            gmsh.model.mesh.field.setNumber(mesh_field_id, "SizeMin", size_min)
+            gmsh.model.mesh.field.setNumber(mesh_field_id, "SizeMax", size_max)
+
+            # Set the distance constraints for the elements
+            gmsh.model.mesh.field.setNumber(mesh_field_id, "DistMin", dist_min)
+            gmsh.model.mesh.field.setNumber(mesh_field_id, "DistMax", dist_max)
```
fabiocruz, line 167: Why are these not external functions, instead of being defined in the scope of the function `_create_mesh_with_gmsh`?

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,268 @@
...
+        # Set the meshing algorithm: Quasi-structured Quad method
+        gmsh.option.setNumber("Mesh.Algorithm", 11)
+
+        # Set mesh smoothing to improve mesh quality
+        gmsh.option.set_number("Mesh.Smoothing", 1000)
+
+        # Generate the mesh in 2D
+        gmsh.model.mesh.generate(dim=2)
+
+    def write_to_msh(self, mesh_path: str) -> None:
```
fabiocruz, line 259: The argument could have the same name as the format, `msh_path`.

inductiva/structures/scenarios/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -0,0 +1,109 @@
...
+
+        Args:
+            simulator: The simulator to use for the simulation.
+            simulation_time: The simulation time, in seconds.
+            resource_pool_id: The resource pool to use for the simulation.
+            run_async: Whether to run the simulation asynchronously.
+        """
+        self.simulation_time = simulation_time
+
+        commands = self.get_commands()
```
fabiocruz, line 75: You don't need this.

fabiocruz (approved):
This is almost there. Only a few more changes needed. I am approving it though, go ahead and merge it when you fix the issues I identified! Thanks a good job @Joao97ribeiro!

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
Joao97ribeiro: yes, it is cleaner, thanks!

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,614 @@
...
+
+        for hole in self.holes:
+            mesh_offset, predefined_element_size = hole.get_mesh_metrics()
+
+            holes_mesh_offset.append(mesh_offset)
+            holes_predefined_element_size.append(predefined_element_size)
+
+        return holes_mesh_offset, holes_predefined_element_size
+
+    def generate_plate_with_holes_get_boundary_ids_and_mesh_metrics(
```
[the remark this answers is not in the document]
Joao97ribeiro: so, something like this: to_occ_and_get_boundary_ids_and_mesh_metrics?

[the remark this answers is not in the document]
fabiocruz: I meant simply `to_occ`. The metrics can either be returned by the method or obtained with a call to another method, right?

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,141 @@
+"""Utils to create the different types of boundary conditions."""
+
+from abc import ABC, abstractmethod
+from typing import Optional, List, Literal
+
+import json
+
+
+class BC(ABC):
```
sarmento, line 9: I would actually rename this to BoundaryCoundition, but it's ok as it is.

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    """Abstract base class for boundary condition.
+
+    Attributes:
+        boundary_name (float): The boundary name of the plate where the boundary
```
sarmento, line 13: Wouldn't it be more natural to have a reference to the boundary, which in case would have a name? Isn't there a Boundary class (not saying there should be but is you have a BoundaryCondition class, it would not be suprising to have a Boundary class).

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @abstractmethod
+    def to_dict(self):
+        """Abstract method to convert the bc properties to a dictionary."""
+        pass
+
+
+class DirichletBC(BC):
+    """Dirichlet boundary condition.
+
```
sarmento, line 32: Try to add a bit more info to why this is different from the Neumann BC

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        return {
+            "boundary_name": self.boundary_name,
+            "displacement_x": self.displacement_x,
+            "displacement_y": self.displacement_y
+        }
+
+
+class NeumannBC(BC):
+    """Neumann boundary condition.
+
```
sarmento, line 66: Yes: pleas give a little bit more info for the less informed :)

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,141 @@
...
+            json_path (str): The JSON file path.
+        """
+
+        # Divide the bcs_obj_list into two lists based on their types
+        dirichlet_bcs = [bc for bc in self.bcs if isinstance(bc, DirichletBC)]
+        neumann_bcs = [bc for bc in self.bcs if isinstance(bc, NeumannBC)]
+
+        # Dirichlet dictionary
+        if dirichlet_bcs is not None:
+            dirichlet_bcs_dict = []
```
sarmento, line 119: This is a bit confusing. Is this a list or a dict?
You initialize a list by []

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
```diff
@@ -0,0 +1,141 @@
...
+            for dirichlet_bc in dirichlet_bcs:
+                dirichlet_bc_dict = dirichlet_bc.to_dict()
```
sarmento, line 121: list_of_dirichlet_bc_dicts?

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,268 @@
+"""Utils to create the mesh."""
+
+from typing import List, Optional
+
+import gmsh
+
+import geometry_utils
+
+
+class Mesh:
```
sarmento, line 10: Mesh is a bit generic of a name...

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
sarmento: Agreed!

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
```diff
@@ -0,0 +1,268 @@
...
+    def write_to_msh(self, mesh_path: str) -> None:
+        """Writes the GMSH mesh to MSH file.
+
+        Args:
+            mesh_path (str): The mesh file path in MSH format.
+        """
+        self._create_mesh_with_gmsh()
+
+        gmsh.write(mesh_path)
+        gmsh.finalize()
```
sarmento, line 268: This is an enormous method. You certainly want to break this in smaller steps. Consider creating a gmsh_utils, or a specific class of objects that transforms whatever you need to transform. Observe that some of these gmesh related operations are also probably used in other scenarios, such as the WindTunnel and Heatsink. Maybe this is not yet the time, but later it would be great to see if you can have a class/utils that also helps those scenarios.

sarmento (commented):
This is looking good.
I think we are really almost there.
I think the only change needed has to do with that massive file/function related with gmesh.

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
Joao97ribeiro: is a list of dicts, i will change to not be confuse, Thanks Luis

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
Joao97ribeiro: yes, great suggestion, i will use it, thanks

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
Joao97ribeiro: i will rename it -> GmshMesh

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
Joao97ribeiro: i will change the name of the method to "get_plate_mesh_params" to make it clear that the method is providing parameters for meshing the plate in Gmsh

inductiva/structures/scenarios/plate_linear_elastic/geometry_utils.py
```diff
@@ -0,0 +1,614 @@
...
+
+        Returns:
+          hole_gmsh (int): The Gmsh entity ID representing the hole's 
+            OpenCASCADE CAD representation.
+        """
+        hole_gmsh = gmsh.model.occ.addRectangle(x=self.center_x - self.size_x,
+                                                y=self.center_y - self.size_y,
+                                                z=0,
+                                                dx=self.size_x * 2,
+                                                dy=self.size_y * 2)
```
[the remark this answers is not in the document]
Joao97ribeiro: the size_x and size_y is half the size of the rectangle to be more similar with other holes (radius and semi-axis), i will change the docstring

inductiva/structures/scenarios/plate_linear_elastic/mesh_utils.py
Joao97ribeiro: Thanks for the suggestion! I agree that breaking down this method into smaller steps is a good idea. Creating a 'gmsh_utils' module or a specific class for transformations sounds like a smart approach. I'll also keep in mind that some Gmsh operations might be reusable for other scenarios like 'WindTunnel' and 'Heatsink.' While I might not make these changes immediately, I'll definitely consider them for future improvements

inductiva/structures/scenarios/plate_linear_elastic/bc_utils.py
Joao97ribeiro: I see your point, but will the boundary be named "boundary" exclusively? Adding the "Boundary" class would provide more structure to the code, but wouldn't people need to initialize the "Boundary" class to generate the boundary condition?

Joao97ribeiro:
Thanks @fabiocruz and @sarmento for the comments and suggestions. I'll merge the code, and in the next iteration, I'll work on reducing it and applying the "GmshMesh2" class to other scenarios

PR #438, opened 2023-08-07
Tasks.delete_machine=True now deletes machine after task is finished

david451lima:
Task now have a new attribute, task.delete_machine. If task.delete_machine is true, the method task.wait() deletes the machine in which the task ran if task is in a terminal status. The value for this attribute is False by default, but can be specified by the user (which leads to some minor changes in a bunch of files)
Also added a get_machine method to get the name of the machine in which the task is running/ran

The method only works rn for ProteinSolvation. After feedback in this PR I will implement the necessary changes for the other scenarios.

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -8,6 +8,7 @@ import datetime
 
 from inductiva.client.models import TaskStatusCode
 from inductiva import api
+from inductiva import admin
 from inductiva.client.apis.tags.tasks_api import TasksApi
 from inductiva.utils import files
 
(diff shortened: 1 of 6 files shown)


cristianacarpinteiro (approved):
Looks good to me! Afterwards, we need to make these changes in the other scenarios. But maybe in a different PR.

inductiva/tasks/task.py
```diff
@@ -130,6 +132,12 @@ def wait(self, polling_period: int = 5) -> TaskStatusCode:
             prev_status = status
 
             if status in terminal_statuses:
+                if self.delete_machine:
+                    delete_logging_info = f"Deleting the machine \
+                                        {self.get_machine()}"
+
+                    logging.info(delete_logging_info)
+                    admin.kill_executer(self.get_machine())
```
IvanPombo, line 140: This will not work as you expect. You can only kill an executer if it does not belong to a group.

Moreover, this implementation only works for sync runs and I am not sure if that fulfils your purpose.

IvanPombo (changes requested):

PR #439, opened 2023-08-07
Resources class

begaiym-k:
Reopening #371


IvanPombo (approved):

cristianacarpinteiro (approved):
Looks good! Thanks Beggy :)

inductiva/resources/machines.py
```diff
@@ -0,0 +1,119 @@
...
+                raise e
+        return self._get_cost(instance_price)
+
+    def _get_cost(self, instance_price):
+        if self.spot:
+            return instance_price.body["on_demand"] * self.num_machines
+        else:
+            return instance_price.body["preemptible"] * self.num_machines
+
+    def _generate_instance_name(self):
```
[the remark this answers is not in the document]
sarmento: Passing this back to the server even makes the list of imports here simpler....

inductiva/resources/resources.py
```diff
@@ -0,0 +1,21 @@
+"""Methods related to managing executers and resource pools."""
+from uuid import UUID
+
+import inductiva
+from inductiva import api
+from inductiva.client import ApiClient, ApiException
+from inductiva.client.apis.tags.executers_api import ExecutersApi
+
+
+def create_machine_group_id() -> UUID:
```
sarmento, line 10: this runs on the server side or client side? 
ID's should be created on the serverside and returned as part of calls.

inductiva/resources/resources.py
```diff
@@ -0,0 +1,21 @@
+"""Methods related to managing executers and resource pools."""
+from uuid import UUID
+
+import inductiva
```
sarmento, line 4: just import inductiva

inductiva/resources/resources.py
```diff
@@ -0,0 +1,21 @@
...
+from inductiva import api
+from inductiva.client import ApiClient, ApiException
```
sarmento, line 6: Do not import classes directly.

inductiva/resources/machines.py
```diff
@@ -0,0 +1,119 @@
...
+                disk_size_gb=self.disk_size_gb,
+                zone=self.zone,
+            )
+            # body.resource_pool_id = self.machine_group_id
+            try:
+                logging.info("Creating a machine group. \
+                             This may take a few minutes.")
+                api_instance.create_instance_group(body=body)
+                logging.info(
+                    "Machine group with the specified settings successfully \
```
sarmento, line 63: Because you mentioned that this can take a few minutes, can you now tell the user how long it actually took?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,119 @@
...
+                        created.")
+                logging.info("Machine type: %s", self.machine_type)
+                logging.info("Number of machines: %s", self.num_machines)
+                logging.info("Spot: %s", self.spot)
+                logging.info("Disk size: %s GB", self.disk_size_gb)
+                logging.info("Machine group ID: %s", self.machine_group_id)
+                if self.name:
+                    logging.info("Name: %s", self.name)
+                logging.info("Machine group successfully created.")
+
```
sarmento, line 73: maybe you want to also print here the estimated price, right?

inductiva/resources/machines.py
```diff
@@ -0,0 +1,119 @@
...
+                instance_price = api_instance.get_instance_price(
+                    body=Instance(name=self.machine_type, zone=self.zone))
+                logging.info(
+                    "Estimated cost of a machine group is %s$ per hour.",
+                    self._get_cost(instance_price))
+            except ApiException as e:
+                raise e
+        return self._get_cost(instance_price)
+
+    def _get_cost(self, instance_price):
```
sarmento, line 110: is this _compute_cost? Also, why do you need this function? Why not directly integrate this in the previous function.

inductiva/resources/machines.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def estimate_cost(self):
+        """Returns an estimated cost per hour of a machine group."""
+        with ApiClient(self.api_config) as client:
+            api_instance = InstanceApi(client)
+
+            try:
+                instance_price = api_instance.get_instance_price(
+                    body=Instance(name=self.machine_type, zone=self.zone))
+                logging.info(
```
sarmento, line 103: I think this logging info make no sense here. It should be on the outside of this function.
What you could log here is something like: "Successfully retrieved price info".
Then you don't need to have the _get_cost function because you will only need to perform the computation once, i.e. for returning.

It's interesting to reflect on why you thought that this was the right place to log the price info to the user... :) It's not, it's *after* you return this call.

sarmento (commented):
Sorry for being late.
There are still a bunch of things that we need to fix here.
Little things: 

i) fix the imports (check Hugo's email)
ii) please add more comments for the user about what is happening.

2) there is the id issues we talked about. It makes no sense to generate id's for resources on the client side. Id's should be created on the server side and returned as part of the API calls.

inductiva/resources/machines.py
```diff
@@ -0,0 +1,119 @@
...
+                spot=self.spot,
+                resource_pool_id=self.machine_group_id,
+                disk_size_gb=self.disk_size_gb,
+                zone=self.zone,
+            )
+            # body.resource_pool_id = self.machine_group_id
+            try:
+                logging.info("Creating a machine group. \
+                             This may take a few minutes.")
+                api_instance.create_instance_group(body=body)
```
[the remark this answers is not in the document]
IvanPombo: Indeed, this should be the workflow, since we only log that id if we can actually launch the machine_group.

PR #440, opened 2023-08-07
[Bug] Fix output type in task class

cristianacarpinteiro:
The task class was only working when the output_dir variable was from type pathlib.Path. In this PR, I fixed it, now output_dir can be a str or a path like object.

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -14,6 +14,7 @@ from inductiva.client.apis.tags.tasks_api import TasksApi
 from inductiva.utils import files
 from inductiva.utils import data
 from inductiva.utils import output_contents
+from inductiva import types
 
 
 class Task:
@@ -160,7 +161,7 @@ class Task:
             The output of the task.
         """
         self.wait()
-        output_dir = self.download_outputs(output_dir)
+        output_dir = self.download_outputs(output_dir=output_dir)
 
         if self._output_class:
             return self._output_class(output_dir)
@@ -198,7 +199,7 @@ class Task:
     def download_outputs(
         self,
         filenames: Optional[List[str]] = None,
-        output_dir: Optional[pathlib.Path] = None,
+        output_dir: Optional[types.Path] = None,
         uncompress: bool = True,
         rm_archive: bool = True,
     ) -> pathlib.Path:
@@ -226,11 +227,10 @@ class Task:
         # implement our own download logic (with progress bar, first checking
         # the size of the file, etc.)
         response = api_response.response
-
+        output_dir = files.resolve_path(output_dir)
         if output_dir is None:
             output_dir = files.resolve_path(inductiva.output_dir).joinpath(
                 self.id)
-
         if output_dir.exists():
             shutil.rmtree(output_dir)
 
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -15,6 +15,7 @@ from inductiva.utils import files
 from inductiva.utils import data
 from inductiva.utils import output_contents
 from inductiva import types
+import warnings
 
 
 class Task:
@@ -228,14 +229,14 @@ class Task:
         # the size of the file, etc.)
         response = api_response.response
 
-        output_dir = files.resolve_path(output_dir)
         if output_dir is None:
             output_dir = files.resolve_path(inductiva.output_dir).joinpath(
                 self.id)
-        if output_dir.exists():
-            shutil.rmtree(output_dir)
+        output_dir = files.resolve_path(output_dir)
 
-        output_dir.mkdir(parents=True)
+        if output_dir.exists():
+            warnings.warn("Path already exists, files may be overwritten.")
+        output_dir.mkdir(parents=True, exist_ok=True)
 
         zip_path = output_dir.joinpath("output.zip")
 
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -55,7 +55,7 @@ class Task:
         """
         return self
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(self, exc_type, exc_value, traceback) -> bool:
         """Exit context manager killing the task if an exception was raised."""
         del traceback  # unused
         del exc_type  # unused
@@ -154,7 +154,8 @@ class Task:
         """Set the output class of the task."""
         self._output_class = output_class
 
-    def get_output(self, output_dir: Optional[types.Path] = None):
+    def get_output(self,
+                   output_dir: Optional[types.Path] = None) -> pathlib.Path:
         """Get the output files produced by the task. If the task 
         is not finished the function will wait until it is to download
         the output folder contents. 
@@ -174,7 +175,7 @@ class Task:
 
         return output_dir
 
-    def get_output_files_info(self):
+    def get_output_files_info(self) -> output_contents.OutputContents:
         """Get information of the output files of the task.
 
         Returns:
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -1,6 +1,5 @@
 """Manage running/completed tasks on the Inductiva API."""
 import pathlib
-import shutil
 import time
 import json
 from absl import logging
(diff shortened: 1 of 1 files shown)


IvanPombo (approved):

inductiva/tasks/task.py
```diff
@@ -14,6 +14,7 @@
 from inductiva.utils import files
 from inductiva.utils import data
 from inductiva.utils import output_contents
+from inductiva import types
```
sarmento, line 17: Check Hugo's emails about how to do import: avoid directly importing classes and functions.

Also, it's ok to just import inductiva and then have the full path so we can get the full package context in the code.

inductiva/tasks/task.py
```diff
@@ -153,14 +154,14 @@ def set_output_class(self, output_class):
...
         self._output_class = output_class
 
-    def get_output(self, output_dir=None):
+    def get_output(self, output_dir: Optional[types.Path] = None):
         """Get the output of the task.
 
         Returns:
             The output of the task.
         """
         self.wait()
```
sarmento, line 163: do you want to info the user that you are getting the input but you are now having to wait for the task to be finished?

Don't forget to provide as much feedback to the user as possible.

inductiva/tasks/task.py
```diff
@@ -153,14 +154,14 @@ def set_output_class(self, output_class):
         """Set the output class of the task."""
         self._output_class = output_class
 
-    def get_output(self, output_dir=None):
+    def get_output(self, output_dir: Optional[types.Path] = None):
         """Get the output of the task.
```
sarmento, line 158: Do you want to add something here about the type of output?
Also, this is a blocking task right?

inductiva/tasks/task.py
```diff
@@ -153,14 +154,14 @@ def set_output_class(self, output_class):
...
 
         Returns:
             The output of the task.
         """
         self.wait()
-        output_dir = self.download_outputs(output_dir)
+        output_dir = self.download_outputs(output_dir=output_dir)
```
sarmento, line 164: this operation can take forever, right?
Do you want to warn the user, or does donwload_outputs already inform the user?

inductiva/tasks/task.py
```diff
@@ -227,10 +228,10 @@ def download_outputs(
         # the size of the file, etc.)
         response = api_response.response
 
+        output_dir = files.resolve_path(output_dir)
```
sarmento, line 231: Are we informing the user of what is going on? Downloading can take forever, right?
How much data did we downloaded? Are we informing the user? Are we informing the user of how much space he has left on his local device?

Do you see my point here? :)

inductiva/tasks/task.py
```diff
@@ -227,10 +228,10 @@ def download_outputs(
...
         if output_dir is None:
             output_dir = files.resolve_path(inductiva.output_dir).joinpath(
                 self.id)
-
         if output_dir.exists():
```
sarmento, line 235: Are we informing the user that we are deleting this directory?

sarmento (approved):
LGTM!
But please have a look at my comments regarding informing the user of what is going on at every moment.

inductiva/tasks/task.py
cristianacarpinteiro: The download_outputs already gives this information: image

inductiva/tasks/task.py
cristianacarpinteiro: This already gives information about the space the dir occupies and the time it took to download. Do you want me to add the space left on the device? 
image

inductiva/tasks/task.py
cristianacarpinteiro: I think it is better not to delete it, actually, I am going to correct it.

inductiva/tasks/task.py
sarmento: Then add this to the comments. You can actually copy paste this info there, so that people know.

inductiva/tasks/task.py
```diff
@@ -198,7 +200,7 @@ def get_output_files_info(self):
     def download_outputs(
         self,
         filenames: Optional[List[str]] = None,
-        output_dir: Optional[pathlib.Path] = None,
+        output_dir: Optional[types.Path] = None,
         uncompress: bool = True,
         rm_archive: bool = True,
```
sarmento, line 205: just out of curiosity. is this removing the *remote* archive, i.e. the one on the bucket?
If so, please add "remote" to this flag, otherwise it is ambiguous.

sarmento (approved):

inductiva/tasks/task.py
```diff
@@ -153,14 +155,14 @@ def set_output_class(self, output_class):
...
 
         Returns:
             The output of the task.
         """
         self.wait()
-        output_dir = self.download_outputs(output_dir)
+        output_dir = self.download_outputs(output_dir=output_dir)
 
         if self._output_class:
             return self._output_class(output_dir)
```
sarmento, line 168: Quick question: if you don't uncompress, what happens here?
Does the output_class know how do deal with compressed formats?
Do you need to pass the information about things having been decompressed or not to the _outout_class so that is know that it has to decompress on demand.
Do you see my point?

@IvanPombo @begaiym-k wdyt?

What happens if we don't decompress the files and pass the output dir to one of these output classes?

inductiva/tasks/task.py
cristianacarpinteiro: Importing the full paths isn't going to be messy? For example doing inductiva.client.apis.tags.tasks_api.TasksApi

inductiva/tasks/task.py
IvanPombo: It won't work, the output classes do not have a notion to decompress the information if it's zipped.

From a user perspective, it does not make much sense to change the argument to keep compressed if he wants to perform their after post-processing. But either way, this possibility needs to be accounted for on the output class. @cristianacarpinteiro can you own this?

In my view, it may be preferable to create a general output class with this feature and the general initialisation with `sim_output_dir`. Thereafter, all output classes inherit from it and just perform the specific renders with things already unpacked.

inductiva/tasks/task.py
cristianacarpinteiro: Importing the full paths isn't going to be messy? For example doing `inductiva.client.apis.tags.tasks_api.TasksApi()` would be super long.

inductiva/tasks/task.py
sarmento: In that case, should we have that flag there?
Do you see my point?
Why are we letting a broken feature go to the hands of the user.

Now: having compressed files could make sense, but it just makes us have a lot more work. But if we are not having that work now, then we can't let this go to the hands of the user.

I need everyone to detect this issues.

inductiva/tasks/task.py
cristianacarpinteiro: Done.

inductiva/tasks/task.py
cristianacarpinteiro: I've updated the comments.

inductiva/tasks/task.py
IvanPombo: Yes, we can still have this flag here. The get_output() makes sure that the files are uncompressed. This is the way the user will fetch the Output class from the scenario. So it is not broken and can stay within the code. However, I misunderstood and was thinking the task.get_output() had the uncompressed flag as well.


The download_outputs is for more general stuff, when ones to download all files or other ones not considered in the task.get_output().

inductiva/tasks/task.py
cristianacarpinteiro: Yes, it is a blocking call, I've added info about that to the function. I've done  also the type annotations where it was missing.

inductiva/tasks/task.py
cristianacarpinteiro: Yes, I see your point! I will open an issue about it, since I think this PR is not the place to do that change.

inductiva/tasks/task.py
cristianacarpinteiro: no, from the code: 
`  if rm_archive:
        zip_path.unlink()`
which means that this is just removing the zip archive.

cristianacarpinteiro: but that feature makes sense.

sarmento: ok. so that should be a new feature.
But still, the naming is ambiguous. What name would you suggest to make this variable less ambiguous?

In general, you should try to spot this sort of ambiguities. Or, when someone asks a stupid question like i did, then you should try to see if this question was because something is not clear and you should try to correct the source of ambiguity.

cristianacarpinteiro: I think that rm_zip_archive can be a better name

sarmento: Then, you should do that. If you identify a source of ambiguity, then just fix it.

Also, you could even be more explicit.

rm_downloaded_zip_archive

PR #441, opened 2023-08-08
Clean machines code

IvanPombo:
This PR addresses the last comments done by @sarmento in a previous PR #439.

- It removes the imports with classes.
- Computes the start and termination time
- Estimates the price during start.
- Simplifies loggings

The next steps are to pass the machine_id and name completely to the backend and retrieve them via the API calls we are
already performing.


inductiva/resources/machines.py
```diff
@@ -32,20 +28,28 @@ def __init__(
...
         self.num_machines = num_machines
         self.spot = spot
         self.disk_size_gb = disk_size_gb
         self.name = self._generate_instance_name()
-        self.machine_group_id = resources.create_machine_group_id()
         self.zone = zone
-        self.api_config = api.validate_api_key(inductiva.api_key)
+        self.estimated_cost = -1  # The negative implies that the cost is unknown.
+
+        # Validate the specific user configurations/privileges from the API.
```
sarmento, line 41: It's still not clear what this does.
If we are validating the privileges, then there should be some exception thrown if the user does not meet the privileges required for this machine group, right? If there is a missing feature please write a TODO. 
Also, what is this api_config? It's not clear to me...

inductiva/resources/machines.py
```diff
@@ -32,20 +28,28 @@ def __init__(
...
-        with ApiClient(self.api_config) as client:
-            api_instance = InstanceApi(client)
+        """Starts a machine group."""
 
-            body = InstanceGroup(
+        with inductiva.client.ApiClient(self.api_config) as client:
+            api_instance = inductiva.client.apis.tags.instance_api.InstanceApi(
+                client)
+
+            self.machine_group_id = inductiva.resources.utils.create_machine_group_id(
```
sarmento, line 52: where is this id being generated? please comment.

inductiva/resources/machines.py
```diff
@@ -54,66 +58,76 @@ def start(self):
                 disk_size_gb=self.disk_size_gb,
                 zone=self.zone,
             )
-            # body.resource_pool_id = self.machine_group_id
             try:
                 logging.info("Creating a machine group. \
                              This may take a few minutes.")
-                api_instance.create_instance_group(body=body)
+                start_time = time()
+                api_instance.create_instance_group(body=instance_group_config)
```
sarmento, line 67: this is where you should get the id, if you need it.
If the id is still being created on the client, please leave a TODO so you can change it later to the server side.

inductiva/resources/machines.py
```diff
@@ -54,66 +58,76 @@ def start(self):
...
+                self.estimated_cost = self._compute_estimated_cost(api_instance)
```
sarmento, line 68: the cost computation is not part of the creation time.

inductiva/resources/machines.py
```diff
@@ -54,66 +58,76 @@ def start(self):
...
+                creation_time = (time() - start_time) / 60
```
sarmento, line 69: what unit?

inductiva/resources/machines.py
```diff
@@ -54,66 +58,76 @@ def start(self):
...
-
-            except ApiException as e:
-                raise e
-        return self.machine_group_id
+                self._log_machine_group_info()
+                logging.info("Machine group successfully created in %s mins.",
+                             creation_time)
+
+            except inductiva.client.ApiException as api_exception:
+                raise api_exception
```
sarmento, line 78: should we still return something?

inductiva/resources/machines.py
```diff
@@ -54,66 +58,76 @@ def start(self):
...
 
     def terminate(self):
+        """Terminates a machine group."""
 
-        with ApiClient(self.api_config) as client:
-            api_instance = InstanceApi(client)
+        with inductiva.client.ApiClient(self.api_config) as client:
```
sarmento, line 83: why are we constantly creating a client. Couldn't this be created at the time of the construction? 
Do clients have timeout? I really don't know.
But if a client is not something that has a time to live, why are we constantly creating a client?

inductiva/resources/machines.py
```diff
@@ -54,66 +58,76 @@ def start(self):
...
+            api_instance = inductiva.client.apis.tags.instance_api.InstanceApi(
```
sarmento, line 84: The same question for API instance: does this have a timeout?
If not, then we should just create one at construction time. But I don't know if this is true...

inductiva/resources/machines.py
```diff
@@ -54,66 +58,76 @@ def start(self):
...
         else:
-            return instance_price.body["preemptible"] * self.num_machines
+            estimated_price = instance_price.body[
+                "preemptible"] * self.num_machines
+
+        return estimated_price
 
     def _generate_instance_name(self):
+        """Generate instance name based on unique_id."""
+
```
sarmento, line 121: leave a todo, because you don't want to do this on the client side.

inductiva/resources/utils.py
```diff
@@ -0,0 +1,20 @@
...
+def create_machine_group_id() -> uuid.UUID:
+    """Request the creation of a machine group ID."""
+
+    api_config = inductiva.api.validate_api_key(inductiva.api_key)
+
+    with inductiva.client.ApiClient(api_config) as client:
+        api_instance = inductiva.client.apis.tags.executers_api.ExecutersApi(client)
+
+        try:
+            api_response = api_instance.create_resource_pool()
```
sarmento, line 17: I don't think this makes sense. I mean, I may be understanding something wrong, but what is create_resource_pool doing? Is it an empty resource pool? where are the parameters of the machines, etc..

inductiva/resources/machines.py
IvanPombo: The api_config carries the API configuration from the client to the backend. In particular, the api_url and api_key.

All the required authentications of the privileges are validated in the backend and the exception is thrown from there. Notice, at the moment we are only verifying if the user is admin or not.

inductiva/resources/machines.py
sarmento: Just to be clear. This with XX pattern is good. And should be kept.
But let's think a little about this later to simplify this repetition.

inductiva/resources/machines.py
IvanPombo: I am thinking about this... but if everything stays as a class attribute, then it does not make sense.

inductiva/resources/machines.py
sarmento: please add comments about exactly that. 
Also, I don't think this is the proper moment to do this and I think there are two different things here mixed. 

1) do I have access to API

2) can I launch this sort of resources. 

These are two different things.
It's ok for now, but we need to think about this later.

inductiva/resources/utils.py
IvanPombo: This will also be passed to the backend... but at the moment, it asks the backend to create a single string that will be used to unique identify the machine group in the DB.

inductiva/resources/machines.py
begaiym-k: This doesn't validate user privileges yet, it only validates the api key and makes sure that it exists in out database. And only when we send the start() request to the API that's when we do all the checks before launching the machine I will open a PR with that very soon.

IvanPombo: Indeed they are, so we can have two different checks for privileges. In all calls to the API, we need to perform the above validation, the second point will only occur for the resources.

inductiva/resources/machines.py
```diff
@@ -63,16 +64,18 @@ def start(self):
             try:
                 logging.info("Creating a machine group. \
                              This may take a few minutes.")
-                start_time = time()
+                start_time = time.time()
+                #TODO: Receive here the machine name and ID.
                 api_instance.create_instance_group(body=instance_group_config)
+                creation_time_mins = (time.time() - start_time) / 60
```
sarmento, line 72: is mins reasonable? should it be seconds? I don't have a clue...
Leave it as it is for now. Then we will know better :)

inductiva/resources/machines.py
```diff
@@ -117,13 +120,16 @@ def _compute_estimated_cost(self, api_instance):
         return estimated_price
```
sarmento, line 122: The naming is weird. The method is "cost" but the variable is "price".

sarmento (approved):
LGTM.
I think it is important now to get the id and name generation to the server side, and log things to file there.

inductiva/resources/machines.py
IvanPombo: I just referred mins to keep with the log above. With more testing we will see what is appropriate.

PR #446, opened 2023-08-08
Fixed imports

joaobernardo23:


inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -53,7 +54,7 @@ class StellaratorCoils(scenarios.Scenario):
           Typically varies between 1 and 3.
     """
 
-    valid_simulators = [stellarators.simulators.Simsopt]
+    valid_simulators = [inductiva.stellarators.simulators.Simsopt]
```
sarmento, line 57: this looks good. And you don't need to change anything.
But pls discuss later with Fabio the possibility of compressing the path here to:

```
inductiva.stellarators.Simsopt
```

We are looking for other opportunities to compress the hierarchy. For now, pls don't change anything.

sarmento (approved):
LGTM!

PR #451, opened 2023-08-08
Raise error for disk size on MachineGroup instantiation

cristianacarpinteiro:
In this PR, I implement an error logging for when the  disk_size set by the user in the MachineGroup is smaller than 20 GB.

### DIFF
--- a/inductiva/resources/machines.py
+++ b/inductiva/resources/machines.py
@@ -27,15 +27,13 @@ class MachineGroup():
             more information about machine types.
             num_machines: The number of virtual machines to launch.
             spot: Whether to use spot machines.
-            disk_size_gb: The size of the disk in GB, minimum is 20 GB.
+            disk_size_gb: The size of the disk in GB, recommended min. is 20 GB.
             zone: The zone where the machines will be launched.
         """
         #TODO: Check if machine type is valid.
         self.machine_type = machine_type
         self.num_machines = num_machines
         self.spot = spot
-        if disk_size_gb < 20:
-            raise ValueError("Disk size must be at least 20 GB.")
         self.disk_size_gb = disk_size_gb
         #TODO: Pass the name generation to the backend
         self.name = self._generate_instance_name()
@@ -68,8 +66,8 @@ class MachineGroup():
                     zone=self.zone,
                 )
             try:
-                logging.info("Creating a machine group."
-                             " This may take a few minutes.")
+                logging.info("Creating a machine group. \
+                             This may take a few minutes.")
                 start_time = time.time()
                 #TODO: Receive here the machine name and ID.
                 api_instance.create_instance_group(body=instance_group_config)
@@ -78,8 +76,8 @@ class MachineGroup():
                 self.estimated_price = self._compute_estimated_price(
                     api_instance)
                 logging.info(
-                    "Machine group with the specified settings successfully"
-                    " created.")
+                    "Machine group with the specified settings successfully \
+                        created.")
                 self._log_machine_group_info()
                 logging.info("Machine group successfully created in %s mins.",
                              creation_time_mins)
@@ -95,16 +93,16 @@ class MachineGroup():
                 client)
 
             try:
-                logging.info("Terminating machine group."
-                             " This may take a few minutes.")
+                logging.info("Terminating machine group. \
+                             This may take a few minutes.")
                 start_time = time.time()
                 api_instance.delete_instance_group(
                     body=inductiva.client.model.instance.Instance(
                         name=self.name))
                 termination_time_mins = (time.time() - start_time) / 60
                 logging.info(
-                    "Machine group of %s machines successfully "
-                    "terminated in % s mins.", self.num_machines,
+                    "Machine group of %s machines successfully \
+                    terminated in % s mins.", self.num_machines,
                     termination_time_mins)
 
             except inductiva.client.ApiException as api_exception:
--- a/inductiva/resources/machines.py
+++ b/inductiva/resources/machines.py
@@ -27,13 +27,15 @@ class MachineGroup():
             more information about machine types.
             num_machines: The number of virtual machines to launch.
             spot: Whether to use spot machines.
-            disk_size_gb: The size of the disk in GB, recommended min. is 20 GB.
+            disk_size_gb: The size of the disk in GB, minimum is 20 GB.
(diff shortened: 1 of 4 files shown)


inductiva/resources/machines.py
```diff
@@ -27,13 +27,15 @@ def __init__(
             more information about machine types.
             num_machines: The number of virtual machines to launch.
             spot: Whether to use spot machines.
-            disk_size_gb: The size of the disk in GB, recommended min. is 20 GB.
+            disk_size_gb: The size of the disk in GB. Minimum is 20 GB.
```
IvanPombo, line 30: Make this more generally configurable, since at least for now when we add more docker images to this will always need to be configured.

cristianacarpinteiro:
We discussed the disk size and it makes more sense that the user defines the disk space he wants to add to the default. The default disk size, should be defined in the backend. I will keep this open until the change happens on inductiva-web-api.

sarmento:
@cristianacarpinteiro : what is the problem we are trying to solve here? Could you please describe the situation that made you decide that this is a priority? I honestly don't see what we are trying to achieve here. Also: if you wanted to put a Maximum Limit, then I could understand that we are trying to protect the user (including us) of overspending, but I honestly don't see why we need this change. Can you pls clarify?

cristianacarpinteiro:
> @cristianacarpinteiro : what is the problem we are trying to solve here? Could you please describe the situation that made you decide that this is a priority? I honestly don't see what we are trying to achieve here. Also: if you wanted to put a Maximum Limit, then I could understand that we are trying to protect the user (including us) of overspending, but I honestly don't see why we need this change. Can you pls clarify?

I was testing the functions implemented by @begaiym-k. When I put a disk_size_gb less than 20 GB there is an error on the bakend that is not explicit or informative, so I wanted to raise the error to prevent the user from putting a disk size smaller than that value. But then we discussed and it would be more intuitive to the user to put the disk size he wants to run the simulations (additional to the default 20GB, the 20GB are there to be able to fit all the docker images, that's why there is the lower limit, a smaller disk size wouldn't fit the images). The default size of the disk would be defined on the backend and the user would just define what he wanted more. So the user can define a disk size of 2GB and on the backend we would do a disk with 22 GB. You see?  It wasn't a question of priority, it was just an easy fix to make the error more explicit, but we decided to alter it on the backend and change the disk_size definition here. But I agree that a cap limit would also make sense.

sarmento:
> > @cristianacarpinteiro : what is the problem we are trying to solve here? Could you please describe the situation that made you decide that this is a priority? I honestly don't see what we are trying to achieve here. Also: if you wanted to put a Maximum Limit, then I could understand that we are trying to protect the user (including us) of overspending, but I honestly don't see why we need this change. Can you pls clarify?
> 
> I was testing the functions implemented by @begaiym-k. When I put a disk_size_gb less than 20 GB there is an error on the bakend that is not explicit or informative, so I wanted to raise the error to prevent the user from putting a disk size smaller than that value. But then we discussed and it would be more intuitive to the user to put the disk size he wants to run the simulations (additional to the default 20GB, the 20GB are there to be able to fit all the docker images, that's why there is the lower limit, a smaller disk size wouldn't fit the images). The default size of the disk would be defined on the backend and the user would just define what he wanted more. So the user can define a disk size of 2GB and on the backend we would do a disk with 22 GB. You see? It wasn't a question of priority, it was just an easy fix to make the error more explicit, but we decided to alter it on the backend and change the disk_size definition here. But I agree that a cap limit would also make sense.

OK! But then let's think carefully about this. Does a "delta" type of parameter make sense? Why don't we just allow the user to set the disk. If the user sets a too small, yes, there should be an exception explaining that. So, your initial solution of just informing what happened is the correct one, and then we should let the user correct it. 

Having a parameter that is a plus change on a default base parameter is very confusing. We don;t have that pattern anywhere else. Users can SET values. We don't allow them to INCREMENT a value. That is not the pattern we are using.

Let's chat about this.

cristianacarpinteiro:
I did some more investigation on this and the exception that is raised on the client side is: 
> ApiException(status=500, reason='Internal Server Error', api_response=ApiResponseWithoutDeserialization(response=<urllib3.response.HTTPResponse object at 0x7feb4d15f310>, body=<inductiva.client.schemas.Unset object at 0x7feb48416b90>, headers=<inductiva.client.schemas.Unset object at 0x7feb48416b90>))

While  the GCP error that is happening on the backend is: 

>  raise core_exceptions.from_http_response(response)
> 2023-08-09T10:55:12.955818845Z google.api_core.exceptions.BadRequest: 400 POST https://compute.googleapis.com/compute/v1/projects/inductiva-api/global/instanceTemplates: Invalid value for field 'resource.properties.disks[0].initializeParams.diskSizeGb': '10000000'. Must be less than or equal to 65536
> 

This is much more informative and should be passed to the user. To this information be passed to the client side, we need to fix it on the backend.

sarmento:
> I did some more investigation on this and the exception that is raised on the client side is:
> 
> > ApiException(status=500, reason='Internal Server Error', api_response=ApiResponseWithoutDeserialization(response=<urllib3.response.HTTPResponse object at 0x7feb4d15f310>, body=<inductiva.client.schemas.Unset object at 0x7feb48416b90>, headers=<inductiva.client.schemas.Unset object at 0x7feb48416b90>))
> 
> While the GCP error that is happening on the backend is:
> 
> > raise core_exceptions.from_http_response(response)
> > 2023-08-09T10:55:12.955818845Z google.api_core.exceptions.BadRequest: 400 POST https://compute.googleapis.com/compute/v1/projects/inductiva-api/global/instanceTemplates: Invalid value for field 'resource.properties.disks[0].initializeParams.diskSizeGb': '10000000'. Must be less than or equal to 65536
> 
> This is much more informative and should be passed to the user. To this information be passed to the client side, we need to fix it on the backend.

OK! Let's try to pass this down to the client. It may be interesting to think if we want to do our own analysis of this sort of exceptions later.

Quick note: from the exception it is possible to see that users could potentially request 64TB of disk. 
How do we make sure that NEVER happens? @cristianacarpinteiro : do you mind checking how much that would actually cost?

@begaiym-k: this is where we certainly want to have an upper limit configured on the server side. It could even be a const for now. What about 30Gb for now?

cristianacarpinteiro:
Since this is solved in the backend, I will close this PR for now.

PR #454, opened 2023-08-09
Remove ID and Name gen from client

IvanPombo:
Remove the ID and Name generation from the client, based on the latest changes on the backend. Right now, we are generating the ID on the backend when machines are started.

This means that with the same class, we can start several machines groups of the same type.


inductiva/resources/machines.py
```diff
@@ -52,35 +50,28 @@ def start(self):
...
             try:
                 logging.info("Creating a machine group. \
                              This may take a few minutes.")
                 start_time = time.time()
-                #TODO: Receive here the machine name and ID.
-                api_instance.create_instance_group(body=instance_group_config)
+                instance_group = api_instance.create_instance_group(body=instance_group_config)
                 creation_time_mins = (time.time() - start_time) / 60
 
+                self.id = instance_group.body["id"]
```
sarmento, line 68: ok. This seems much better.

sarmento (approved):

IvanPombo:
I am going to merge this PR now, but I have commented the estimation of price for now, since it is not working at the moment.

I will dig into this now, for the client to have visibility of what is going on. (Maybe we won't even present this on the logs to the client).

PR #459, opened 2023-08-10
Abstract terrain generation from bathymetries

IvanPombo:
This PR passes the function used in the Bathymetries class to generate random seabeds. This function uses the diamond-square algorithm to general depths for all resolution types.

We move it into a place that can be easily used by other random terrain generators.


inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -134,38 +120,9 @@ def from_random_depths(
...
-
-        x, y = np.meshgrid(x, y, indexing="ij")
-
-        depths = scipy.interpolate.griddata(
-            (x_square.flatten(), y_square.flatten()),
-            depths.flatten(),
-            (x, y),
-            method="linear",
-        )
+        _, _, depths = inductiva.generative.generate_grid_elevation(
```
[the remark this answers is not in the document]
IvanPombo: Might as well just do it now!

inductiva/generative/procedural/terrain.py
```diff
@@ -6,10 +6,10 @@
 import numpy as np
 import scipy
 
-import inductiva
+from inductiva.generative import procedural
```
sarmento, line 9: ok. This looks good like this.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -120,7 +120,7 @@ def from_random_depths(
             random.uniform(-max_depth, 0),
         ]
 
-        _, _, depths = inductiva.generative.generate_grid_elevation(
+        _, _, depths = inductiva.generative.procedural.generate_random_terrain(
```
sarmento, line 123: see: here you are just storing the depths.
I really see no use case for the other two return values.
Please correct me if I am wrong.

sarmento (approved):

inductiva/fluids/scenarios/coastal_area/coastal_area.py
IvanPombo: The others are necessary for other types of terrains. The bathymetry file just needs the depths, but other terrain types generate the mesh from the three variables.

sarmento: Why is that? Can you be more specific?
What do those variable hold that you don't already know before hand?

IvanPombo: In this case, they are generated inside the procedural.generate_random_terrain from x_range, y_range and x_num, y_num. They are the coordinates of all the points in the mesh and say respect to z-values we are generating.

Mesh files will be generated by the x_grid, y_grid and z_values.

sarmento: But you know those from the parameters you pass to the method, right?
Those are only related with scales. They don't have to be produced inside this method, or am i wrong?

IvanPombo: Indeed, we do. We require them to generate the Bathymetry and other terrain meshes. For the Bathymetry object they won't be need to thereafter run simulations (at least for now), but for the others they will.

Since, if we read a terrain from a file we already have them by default, this seems like the appropriate place to create them.

IvanPombo:
Gonna merge this now, and the discussion we are having can be more easily seen with the next PR.

inductiva/generative/procedural/terrain.py
```diff
@@ -0,0 +1,89 @@
...
+    # Create elevation for a square grid with side resolution=size_square.
+    z_elevation = procedural.diamond_square.create_random_array(
+        size=size_square,
+        corner_values=corner_values,
+        initial_roughness=initial_roughness,
+        roughness_factor=roughness_factor)
+
+    # Adjust terrain to ensure that a given percentage of the terrain elevation
+    # is above or below the xy-plane.
+    percentile_translate = np.percentile(z_elevation,
```
sarmento, line 67: Taking into account my previous comment on another PR, can you please for create a separate function that takes care of this "physical scaling" part, so that we then decide better what goes where?

PR #461, opened 2023-08-10
Add Terrain object

IvanPombo:
In this PR, I add a Terrain class containing two initialisation methods - from a file or randomly generated. We assume that the Terrain is defined through a mesh - pyvista - for now and take advantage of the capabilities of pyvista to immediately generate the visualizations.

Render examples:
1Km x 1Km - (500x500 resolution) and  around 100m height
terrain

2Km x 1Km - (1000x1000 resolution) and around 400m height
terrain


inductiva/fluids/scenarios/wind_terrain/terrain.py
```diff
@@ -0,0 +1,147 @@
+"""Represent a complex terrain profile."""
```
sarmento, line 1: I think this should not live under fluid/scenarios.
I think we need to create another package just for this sort of classes.

inductiva/world

Let's open an issue on this, but for now please move it out of fluids and place it in world (or another package you see more fit.)

inductiva/fluids/scenarios/wind_terrain/terrain.py
```diff
@@ -0,0 +1,147 @@
...
+        Args:
+            x_range = [x_min, x_max]: Range of the x-axis in meters (m).
+            y_range = [y_min, y_max]: Range of the y-axis in meters (m).
+            x_num: Number of points in the x-axis.
+            y_num: Number of points in the y-axis.
+            height_factor: Factor to multiply the elevation by.
+            initial_roughness: Initial roughness of the terrain.
+            roughness_factor: Factor to multiply the roughness by.
+        """
+        corner_values = [
```
sarmento, line 63: the corner values max and min should become a parameter (which can have defaults). That is the way to "ensure" you can create things like hill sides, etc...

inductiva/fluids/scenarios/wind_terrain/terrain.py
```diff
@@ -0,0 +1,147 @@
...
+            random.uniform(0, 1),
+            random.uniform(0, 1),
+            random.uniform(0, 1),
+            random.uniform(0, 1)
+        ]
+
+        x_grid, y_grid, z_elevation = procedural.generate_random_terrain(
```
sarmento, line 70: OK! This is why I think we need to refactor this. 
There is one aspect which is getting a m_pixel by n_pixel grid. This grid does not depend on the final physical dimension, it just gives you a random z map of  m x n dimension. The physical "dimensions" are computed after by scaling. I think these should be two separate methods. Let me look again at the code that was submitted before.

sarmento (commented):
I am only happy when you show me a wider variety of terrain, including a hill side :)

inductiva/fluids/scenarios/wind_terrain/terrain.py
IvanPombo: Seems appropriate :)

inductiva/generative/procedural/terrain.py
```diff
@@ -70,22 +70,48 @@ def generate_random_terrain(
...
+                                            [size_square, size_square],
+                                            [x_num, y_num], z_elevation)
 
-    x_square, y_square = np.meshgrid(x_square, y_square, indexing="ij")
+    return z_elevation
 
-    x_grid = np.linspace(*x_range, x_num)
-    y_grid = np.linspace(*y_range, y_num)
+
+def create_grid(x_range, y_range, resolution):
```
sarmento, line 80: Observe that this method has *nothing* to do with terrain generation.
Should it be here? :)

inductiva/generative/procedural/terrain.py
```diff
@@ -70,22 +70,48 @@ def generate_random_terrain(
...
+    y_grid[i, j] = y_range[j].
+
+    Args:
+        x_range: The range of x values, in meters.
+        y_range: The range of y values, in meters.
+        resolution: Number of grid points in both
+            directions [x_res, y_res]
+    """
+
+    x_grid = np.linspace(*x_range, resolution[0])
```
sarmento, line 95: is this really "resolution"? A higher resolution means more detail. Does this parameter really mean resolution?
Does this correspond to the num parameter in linspace.
Could you please directly do the assignment num=resolution[0]

Are you starting to see the consistency problems that this code has?

inductiva/generative/procedural/terrain.py
```diff
@@ -70,22 +70,48 @@ def generate_random_terrain(
...
+    y_grid = np.linspace(*y_range, resolution[1])
 
     x_grid, y_grid = np.meshgrid(x_grid, y_grid, indexing="ij")
```
sarmento, line 98: Observe that the core of this method is this call to meshgrid. In fact, this whole method is just a wrapper around this meshgrid method. If that is the case, should this method be called get_meshgrid, which would return the same thing as np.meshgrid and would thus be much easier to understand?

Is this even the right place for this?

inductiva/generative/procedural/terrain.py
```diff
@@ -70,22 +70,48 @@ def generate_random_terrain(
...
 
-    z_elevation = scipy.interpolate.griddata(
-        (x_square.flatten(), y_square.flatten()),
+    return x_grid, y_grid
+
+
+def interpolate_between_grids(x_range, y_range, prev_resolution, new_resolution,
```
sarmento, line 103: again: nothing of this depends on this being a terrain.
Also, it does not require any "physical" dimensions. The core method should be able to map from m x n to p x q points. The physical dimensions are applied after you do the scaling.

inductiva/generative/procedural/terrain.py
IvanPombo: It can be considered resolution because with more res we also have more detail. But I will change to num_samples.

sarmento: use "num".
That way it becomes easier to understand since numpy is a sort of standard.

IvanPombo:
More terrains:

terrain_2

terrain_3

terrain_4

inductiva/utils/grids.py
```diff
@@ -0,0 +1,64 @@
+"""Utility functions to manipulate grid data."""
+import typing
+
+import scipy
+import numpy as np
+
+
+def get_meshgrid(x_range: typing.Sequence[float],
+                 y_range: typing.Sequence[float], num_list: typing.List[int]):
+    """Create a grid of x and y values.
```
sarmento, line 10: why do we need a num_list?
why not x_num e y_num?
Otherwise it would be fair to ask: why don't we have range_list?

inductiva/utils/grids.py
```diff
@@ -0,0 +1,64 @@
...
+    x_grid, y_grid = np.meshgrid(x_grid, y_grid, indexing="ij")
+
+    return x_grid, y_grid
+
+
+def interpolate_between_grids(x_range: typing.Sequence[float],
+                              y_range: typing.Sequence[float],
+                              prev_num_list: typing.List[int],
+                              new_num_list: typing.List[int],
+                              z_array: np.ndarray):
```
sarmento, line 36: can't you infer prev_num_list from z_array?

inductiva/world/terrain.py
```diff
@@ -0,0 +1,175 @@
...
+            num_list: List of number of points in both directions
+                [x_num, y_num].
+            initial_roughness: Initial roughness of the terrain.
+            roughness_factor: Factor to multiply the roughness by.
+            min_elevation: Minimum elevation of the terrain.
+            max_elevation: Maximum elevation of the terrain.
+            random_seed: Random seed to use for the terrain generation.
+        """
+        random.seed(random_seed)
+        corner_values = [
```
sarmento, line 68: We want to be able to set each of these *individually* so we can have more control on the type of landscape.
For example: a plain would have these ranging in equal interavals, but a slope would have two of these high and two low.

inductiva/world/terrain.py
```diff
@@ -0,0 +1,175 @@
...
+            random.uniform(min_elevation, max_elevation),
+            random.uniform(min_elevation, max_elevation),
+            random.uniform(min_elevation, max_elevation),
+            random.uniform(min_elevation, max_elevation)
+        ]
+
+        z_elevation = procedural.generate_random_terrain(
+            x_range=x_range,
+            y_range=y_range,
+            num_list=num_list,
```
sarmento, line 78: x_num and y_num

IvanPombo:
@sarmento I will address the third level (separate the terrain general grid structure from the pyvista mesh in another PR). Right now, I addressed the separation of the terrain generation from the physical structure.

inductiva/__init__.py
```diff
@@ -15,6 +15,7 @@
 from . import templates
 from . import utils
 from . import stellarators
+from . import world
 
 import absl
```
sarmento, line 20: this is one of the standard cases where you can do from
```
absl import logging
```

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -120,9 +120,11 @@ def from_random_depths(
             random.uniform(-max_depth, 0),
         ]
 
-        _, _, depths = inductiva.generative.procedural.generate_random_terrain(
-            x_range, y_range, x_num, y_num, corner_values, initial_roughness,
-            roughness_factor, percentile_above_water)
+        depths = inductiva.generative.procedural.generate_random_terrain(
+            x_num, y_num, corner_values, initial_roughness, roughness_factor)
+
+        depths = inductiva.generative.procedural.adjust_terrain_elevation(
```
sarmento, line 126: I would drop the words "terrain" and "elevation". I think we are still working at the abstract non-physical level. This should be something like adjust_depth_map_level.

I am not sure if we are going back, but I am even ok with merging this method with the previous, since it requires adding only one parameter and can be fully implemented at the "abstract" / "topological" level, i..e without requiring notion of the physical dimensions.

It's up to you, but I think we should drop "terrain" from the name.

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -120,9 +120,11 @@ def from_random_depths(
             random.uniform(-max_depth, 0),
         ]
 
-        _, _, depths = inductiva.generative.procedural.generate_random_terrain(
-            x_range, y_range, x_num, y_num, corner_values, initial_roughness,
-            roughness_factor, percentile_above_water)
+        depths = inductiva.generative.procedural.generate_random_terrain(
```
sarmento, line 123: Why don't you go more specific and call this:

```
inductiva.generative.procedural.generate_random_depth_map(
```

A terrain has physical properties. A depth map is an abstract thing on a grid.

inductiva/generative/procedural/__init__.py
```diff
@@ -1,2 +1,3 @@
 #pylint: disable=missing-module-docstring
 from .terrain import generate_random_terrain
+from .terrain import adjust_terrain_elevation
```
sarmento, line 3: see my previous comments.

inductiva/generative/procedural/terrain.py
```diff
@@ -4,20 +4,18 @@
 import typing
 
 import numpy as np
-import scipy
 
+import inductiva
 from inductiva.generative import procedural
 
 
 def generate_random_terrain(
```
sarmento, line 12: depth_map

inductiva/generative/procedural/terrain.py
```diff
@@ -33,10 +31,8 @@ def generate_random_terrain(
     `(x_num, y_num)`.
 
     Args:
-        x_range: The range of x values, in meters.
-        y_range: The range of y values, in meters.
-        x_num: Number of grid points in the x direction.
-        y_num: Number of grid points in the y direction.
+        x_num: The number of points in the x direction.
```
sarmento, line 34: See: much simpler.

inductiva/utils/grids.py
```diff
@@ -0,0 +1,43 @@
...
+
+    x_grid = np.linspace(*x_range, num=x_num)
+    y_grid = np.linspace(*y_range, num=y_num)
+
+    x_grid, y_grid = np.meshgrid(x_grid, y_grid, indexing="ij")
+
+    return x_grid, y_grid
+
+
+def interpolate_between_grids(x_num: int, y_num: int, z_array: np.ndarray):
```
sarmento, line 19: much cleaner. But the method name is not very friendly.
Why not 

```
reshape/rescale_map(z_array, target_x_num, target_y_num)
```

interpolation is the *strategy* we are using for this reshaping. It could be clipping, it could be mirroring, etc What we really want is the reshaping/rescaling. The method name should reflect that.

inductiva/utils/grids.py
```diff
@@ -0,0 +1,43 @@
...
+
+    x_grid_prev, y_grid_prev = np.meshgrid(np.arange(z_array.shape[0]),
+                                           np.arange(z_array.shape[1]),
+                                           indexing="ij")
+
+    x_grid_new, y_grid_new = np.meshgrid(np.arange(x_num),
+                                         np.arange(y_num),
+                                         indexing="ij")
+
+    new_z_array = scipy.interpolate.griddata(
```
sarmento, line 36: return here directly

inductiva/utils/grids.py
```diff
@@ -0,0 +1,43 @@
...
+def interpolate_between_grids(x_num: int, y_num: int, z_array: np.ndarray):
+    """Interpolate between two grid with different resolution.
+    
+    Args:
+        x_num: New number of grid points in x-direction.
+        y_num: New number of grid points in y-direction.
+        z_array: Values on the z-axis of the previous grid.
+    """
+
+    x_grid_prev, y_grid_prev = np.meshgrid(np.arange(z_array.shape[0]),
```
sarmento, line 28: out of curiosity: do you need to do this? Can you get x_grid_prev and y_grid_prev directly from z_array? (I am just curious to know)

sarmento (approved):
Much cleaner.
I think it's ok to submit it now after you do one more pass to make it more "topological", and then we can do the rest of the work later.

inductiva/utils/grids.py
IvanPombo: The z_array does not carry information about the specific values of the x_grid_prev...
The way to get it is the len of the array.

PR #462, opened 2023-08-11
Copy the plasma surface file to the inputs directory

joaobernardo23:
If the user provides the input file, it is copied to the inputs directory to be used for the simulation.
If the user does not provide a file, a default one is used (located in the templates directory).


inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -223,25 +229,34 @@ def simulate(
         simulator: simulation.Simulator = stellarators.simulators.Simsopt(),
         resource_pool_id: Optional[UUID] = None,
         run_async: bool = False,
-        plasma_surface_filename: str = 'input.QA',
+        plasma_surface_filepath: Optional[types.Path] = None,
```
IvanPombo, line 232: Do typing.Optional here and on top. Same with uuid.UUID and simplify the imports on top!

IvanPombo (approved):

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -2,17 +2,23 @@
...
 
+from absl import logging
 from functools import singledispatchmethod
 from typing import Optional
 from uuid import UUID
 
 import numpy as np
 
-from inductiva import scenarios, simulation, stellarators, tasks
+from inductiva import scenarios, simulation, stellarators, tasks, types, utils
```
sarmento, line 14: try to do just:

```
import inductiva
```

All the code below becomes much more readable.

IvanPombo: We are having problems with the structure of the imports. Some are working to do just that other don't...

joaobernardo23: I had that at first and I really wanted to use it, but in this case I think its impossible because it creates a circular import. This is due to the fact that in the inductiva `__init__.py` file I need to do `from . import stellarators`. Therefore, if here I do `import inductiva` it will create a circular import and the code will not run.

joaobernardo23: I do not know if there is another way to make the code more readable and avoid the circular imports.

sarmento (approved):
Later, let's think about how to fix these circular imports.

joaobernardo23:
> Later, let's think about how to fix these circular imports.

Noted, I believe it is an important thing because it also happens in other scenarios.

PR #464, opened 2023-08-11
[ProteinSolvation] Allow the user to costumize more the visualization

cristianacarpinteiro:
When testing the visualisation module with a bigger protein I was having some problems. These changes make the visualisation rendering more flexible, since I've exposed some parameters to the user. I also renamed some variables for the sake of readability.

### DIFF
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -132,7 +132,8 @@ class ProteinSolvationOutput:
         Args: 
             residue_attributes: The per residue values of the attribute you want 
             to visualize.
-            representation: The protein representation to use for the visualization.
+            representation: The protein representation to use for the 
+            visualization.
             use_compressed_trajectory: Whether to use the compressed trajectory 
             or the full precision trajectory.
             """
@@ -141,8 +142,8 @@ class ProteinSolvationOutput:
         protein = universe.select_atoms("protein")
         for residue, value in zip(protein.residues, residue_attributes):
             residue.atoms.tempfactors = value
-        view = nv.show_mdanalysis(protein)
-        view.add_representation(representation)
+        view = nv.show_mdanalysis(universe)
+        view.add_representation(representation, selection="protein")
         view.update_representation(color_scheme="bfactor")
         view.center()
         return view
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -96,7 +96,7 @@ class ProteinSolvationOutput:
         Args:
             nglview_visualization: Whether to return visualization of the 
             RMSF using nglview or not."""
-
+        start_time = time.time()
         universe = self.construct_universe(use_compressed_trajectory)
         topology = os.path.join(self.sim_output_dir, "solvated_protein.tpr")
 
@@ -118,6 +118,8 @@ class ProteinSolvationOutput:
         plt.title("RMSF per residue")
         plt.grid(True)
         plt.show()
+        duration = time.time() - start_time
+        print(f"RMSF calculation took {duration:.2f} seconds.")
         return rmsf_values
 
     def render_attribute_per_residue(
@@ -140,7 +142,7 @@ class ProteinSolvationOutput:
         for residue, value in zip(protein.residues, residue_attributes):
             residue.atoms.tempfactors = value
         view = nv.show_mdanalysis(protein)
-        view.update_representation(color_scheme="bfactor")
         view.add_representation(representation)
+        view.update_representation(color_scheme="bfactor")
         view.center()
         return view
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -7,12 +7,14 @@ import typing
 import matplotlib.pyplot as plt
 import MDAnalysis as mda
 import nglview as nv
+import numpy as np
 
 import inductiva
 
 FULL_TRAJECTORY_FILE = "full_trajectory.trr"
 COMPRESSED_TRAJECTORY_FILE = "compressed_trajectory.xtc"
 TOPOLOGY_FILE = "solvated_protein.tpr"
+
 
 class ProteinSolvationOutput:
     """Post process the simulation output of a ProteinSolvation scenario."""
@@ -54,7 +56,7 @@ class ProteinSolvationOutput:
         protein = universe.select_atoms("protein")
         view = nv.show_mdanalysis(protein)
         view.add_representation(representation, selection=selection)
-        if add_backbone: #hardcoding the backbone as a cartoon representation
+        if add_backbone:  #hardcoding the backbone as a cartoon representation
             view.add_representation("cartoon", selection="protein")
         view.center()
 
@@ -73,8 +75,7 @@ class ProteinSolvationOutput:
         Args:
             use_compressed_trajectory: Whether to use the compressed trajectory
             or the full precision trajectory."""
-        topology_path = os.path.join(self.sim_output_dir,
-                                     TOPOLOGY_FILE)
+        topology_path = os.path.join(self.sim_output_dir, TOPOLOGY_FILE)
         if use_compressed_trajectory:
             trajectory_path = os.path.join(self.sim_output_dir,
                                            COMPRESSED_TRAJECTORY_FILE)
@@ -104,8 +105,7 @@ class ProteinSolvationOutput:
             RMSF using nglview or not."""
         start_time = time.time()
         universe = self.construct_universe(use_compressed_trajectory)
-        topology_path = os.path.join(self.sim_output_dir,
-                                     TOPOLOGY_FILE)
+        topology_path = os.path.join(self.sim_output_dir, TOPOLOGY_FILE)
 
         aligned_trajectory_path = os.path.join(self.sim_output_dir,
                                                "aligned_traj.dcd")
@@ -116,19 +116,21 @@ class ProteinSolvationOutput:
         # Calculate RMSF for carbon alpha atoms
         c_alphas = align_universe.select_atoms("protein and name CA")
         rmsf = mda.analysis.rms.RMSF(c_alphas).run()
-        residue_number = c_alphas.resids
         rmsf_values = rmsf.results.rmsf
+        duration = time.time() - start_time
+        print(f"RMSF calculation took {duration:.2f} seconds.")
+        return rmsf_values
 
+    def plot_rmsf_per_residue(self, rmsf_values: np.array):
+        "Plot RMSF values per residue."
         # Plot the data
+        residue_number = np.arange(len(rmsf_values))
         plt.plot(residue_number, rmsf_values)
         plt.xlabel("Residue Number")
         plt.ylabel("RMSF")
         plt.title("RMSF per residue")
         plt.grid(True)
         plt.show()
-        duration = time.time() - start_time
-        print(f"RMSF calculation took {duration:.2f} seconds.")
-        return rmsf_values
 
     def render_attribute_per_residue(
             self,
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -1,11 +1,14 @@
 "Postprocessing steps for the MDWaterBox scenario."
 import os
+import numpy as np
 import nglview as nv
+import time
 import matplotlib.pyplot as plt
 import pathlib
 from typing import Literal
 import MDAnalysis as mda
-from inductiva.molecules.scenarios import utils 
+
+from inductiva.molecules.scenarios import utils
 
 
 class ProteinSolvationOutput:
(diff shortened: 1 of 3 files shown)


inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -1,18 +1,20 @@
 "Postprocessing steps for the MDWaterBox scenario."
 import os
+import numpy as np
 import nglview as nv
+import time
 import matplotlib.pyplot as plt
-from pathlib import Path
+import pathlib
 from typing import Literal
```
sarmento, line 8: just import typing.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -1,18 +1,20 @@
...
-from MDAnalysis.analysis import rms
 import MDAnalysis as mda
-from inductiva.molecules.scenarios.utils import unwrap_trajectory, align_trajectory_to_average
+
+from inductiva.molecules.scenarios import utils
```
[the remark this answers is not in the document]
sarmento: try this:

1) native python imports (alphabetically)

```
import os
import pathlib
import time
import typing
```
2) pip-installed packages (alphabetically)
```
import matplotlib.pyplot as plt
import nglview as nv
import numpy as np
```
3) our / internal packages (alphabetically)

```
import inductiva
```

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -26,24 +28,28 @@ def __init__(self, sim_output_path: Path = None):
 
     def render_interactive(self,
                            representation: Literal["cartoon", "ball+stick",
```
sarmento, line 30: with the right import this becomes 

typing.Literal

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -26,24 +28,28 @@ def __init__(self, sim_output_path: Path = None):
...
             trajectory or the full precision one.
             """
         universe = self.construct_universe(use_compressed_trajectory)
-        view = nv.show_mdanalysis(universe)
-        view.add_representation(representation,
-                                selection="not water and not ion")
+        protein = universe.select_atoms("protein")
+        view = nv.show_mdanalysis(protein)
+        view.add_representation(representation, selection=selection)
         if add_backbone:
```
sarmento, line 53: explain why you have fixed "cartoon" here.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -65,12 +71,12 @@ def construct_universe(self, use_compressed_trajectory: bool = False):
...
         if use_compressed_trajectory:
-            trajectory = os.path.join(self.sim_output_dir, "trajectory.xtc")
+            trajectory = os.path.join(self.sim_output_dir,
+                                      "compressed_trajectory.xtc")
         else:
             trajectory = os.path.join(self.sim_output_dir,
-                                      "solvated_protein.trr")
-
-        universe = unwrap_trajectory(topology, trajectory)
+                                      "full_trajectory.trr")
```
sarmento, line 78: should these filenames be defined as constants somewhere?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
david451lima, line 8: Is this import correct?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -112,15 +118,22 @@ def calculate_rmsf_trajectory(self,
         plt.title("RMSF per residue")
         plt.grid(True)
         plt.show()
+        duration = time.time() - start_time
+        print(f"RMSF calculation took {duration:.2f} seconds.")
```
david451lima, line 122: Why are you computing the time? Does it take that much to compute RMSF?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -65,12 +71,12 @@ def construct_universe(self, use_compressed_trajectory: bool = False):
             or the full precision trajectory."""
         topology = os.path.join(self.sim_output_dir, "solvated_protein.tpr")
         if use_compressed_trajectory:
-            trajectory = os.path.join(self.sim_output_dir, "trajectory.xtc")
+            trajectory = os.path.join(self.sim_output_dir,
```
sarmento, line 74: this is a trajectory_path, right?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -65,12 +71,12 @@ def construct_universe(self, use_compressed_trajectory: bool = False):
             or the full precision trajectory."""
         topology = os.path.join(self.sim_output_dir, "solvated_protein.tpr")
```
sarmento, line 72: this is a topology path, right?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -90,18 +96,18 @@ def calculate_rmsf_trajectory(self,
         Args:
             nglview_visualization: Whether to return visualization of the 
             RMSF using nglview or not."""
-
+        start_time = time.time()
         universe = self.construct_universe(use_compressed_trajectory)
         topology = os.path.join(self.sim_output_dir, "solvated_protein.tpr")
```
sarmento, line 101: path

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -26,24 +28,28 @@ def __init__(self, sim_output_path: Path = None):
...
             visualization.
+            selection: The selection to use for the representation. Check 
+            https://nglviewer.org/ngl/api/manual/usage/selection-language.html 
+            for details.
             add_backbone: Whether to add the protein backbone to the 
             visualization.
             use_compressed_trajectory: Whether to use the compressed 
             trajectory or the full precision one.
             """
         universe = self.construct_universe(use_compressed_trajectory)
```
david451lima, line 49: There is only need for a compressed trajectory here? Iirc before I was using a trajectory and a apdb file, why does it only need one rn?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -90,18 +96,18 @@ def calculate_rmsf_trajectory(self,
...
-        align_trajectory_to_average(universe, aligned_trajectory_path)
+        utils.align_trajectory_to_average(universe, aligned_trajectory_path)
         align_universe = mda.Universe(topology, aligned_trajectory_path)
 
         # Calculate RMSF for carbon alpha atoms
         c_alphas = align_universe.select_atoms("protein and name CA")
-        rmsf = rms.RMSF(c_alphas).run()
+        rmsf = mda.analysis.rms.RMSF(c_alphas).run()
         residue_number = c_alphas.resids
         rmsf_values = rmsf.results.rmsf
```
sarmento, line 113: should plotting be a different method?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -112,15 +118,22 @@ def calculate_rmsf_trajectory(self,
         plt.title("RMSF per residue")
         plt.grid(True)
         plt.show()
+        duration = time.time() - start_time
+        print(f"RMSF calculation took {duration:.2f} seconds.")
         return rmsf_values
```
sarmento, line 123: It seems plotting would take the output of this method and some other parameters (time) and plot this in whatever way we want. Plotting inside this method seems a bit weird.

david451lima (approved):
LGTM, just some questions to really understand the new way the viz works

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: I think so, why not?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: For the big protein it was taking like 3 min!

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -112,15 +118,22 @@ def calculate_rmsf_trajectory(self,
...
+            use_compressed_trajectory: bool = False):
         """Render a specific protein attribute in an interactive visualization.
         Args: 
             residue_attributes: The per residue values of the attribute you want 
             to visualize.
+            representation: The protein representation to use for the 
+            visualization.
             use_compressed_trajectory: Whether to use the compressed trajectory 
             or the full precision trajectory.
             """
```
sarmento, line 139: Observe that this method only uses class information in 
```
 self.construct_universe(use_compressed_trajectory)
```

everything else seems to be given as parameter input or taken from universe.

Do you really think this should be a class method?

inductiva/templates/protein_solvation/gromacs/commands.json
```diff
@@ -7,5 +7,5 @@
     {"cmd": "gmx grompp -f energy_minimization.mdp -c protein_solv_ions.gro -p topol.top -o em.tpr", "prompts":[]},
     {"cmd": "gmx mdrun -deffnm em -v yes", "prompts":[]},
     {"cmd": "gmx grompp -f simulation.mdp -c em.gro -r em.gro -p topol.top -o solvated_protein.tpr", "prompts":[]},
-    {"cmd": "gmx mdrun -deffnm solvated_protein -x trajectory.xtc -v yes", "prompts":[]}
+    {"cmd": "gmx mdrun -deffnm solvated_protein -o full_trajectory.trr -x compressed_trajectory.xtc -v yes", "prompts":[]}
```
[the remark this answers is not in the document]
cristianacarpinteiro: it returns the entire system, but I can look into a way of maybe just outputting the protein to have a smaller file!

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: Ah, inside the construct_universe function it reads the topology file!

sarmento (commented):
Thank you for this work.
I don't have a lot of context so my general comment is that we may need a bit more documentation to allow for other people to help you reviewing the code, otherwise you get stuck and nobody can really help you. Please help the others help you :)

My biggest concern is that is seems that the  ProteinSolvationOutput class is trying to provide too many functionalities: everything from visualization to doing some post-processing analysis is hanging from this class. In many cases, these methods almost don't use any information from the object except maybe one field, and then they do their things with several steps not involving the information of the object. I think ProteinSolvationOutput should be simpler and then passed as input to methods that do more complicated stuff (e.g. compute RMS or other interesting things.)

Anyway, I think we need one more round of cleaning up. There are still a number of little things that can be improved, besides what I believe is a major refactoring that may need to be done later.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: Because I'm representing the backbone of the protein and I want it to stand out from the rest. In a photo is easier to see.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: Yes, that makes sense. Maybe as global variables?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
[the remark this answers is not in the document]
cristianacarpinteiro: Thanks, I didn't know there was a order, but makes a lot of sense!

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: image
The blue part is the backbone and I want to highlight it!

sarmento: Great. But the reviewer may not know that. Please explain that because that's what you want, you are hardcoding "cartoon" there.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
sarmento: Yes. Pls do that and then we will figure out how to manage these better.
But are they constants of the scenario? It seems so...

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -65,12 +72,13 @@ def construct_universe(self, use_compressed_trajectory: bool = False):
             or the full precision trajectory."""
         topology = os.path.join(self.sim_output_dir, "solvated_protein.tpr")
         if use_compressed_trajectory:
-            trajectory = os.path.join(self.sim_output_dir, "trajectory.xtc")
+            trajectory = os.path.join(self.sim_output_dir,
+                                      "compressed_trajectory.xtc")
         else:
             trajectory = os.path.join(self.sim_output_dir,
```
sarmento, line 78: trajectory_path

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: I did it this way to avoid code duplication.

PR #476, opened 2023-08-21
Stellarators redefined simulate method

joaobernardo23:
This PR aims to include new arguments to the `simulate` method that will decide (when evaluated the API side) what is the simulation that will be performed. 

For the moment, I only included two cases, one where the user provides all the additional arguments and another where the user provides none of them. 

This seemed better for now rather than to check if every argument is None or not and then apply a default value to those that were, but it is easily extendible to that case, it would just make the code look a bit different with a lot of 'if' statements.

### DIFF
--- a/inductiva/stellarators/scenarios/coils/coils.py
+++ b/inductiva/stellarators/scenarios/coils/coils.py
@@ -230,8 +230,31 @@ class StellaratorCoils(scenarios.Scenario):
         resource_pool_id: typing.Optional[uuid.UUID] = None,
         run_async: bool = False,
         plasma_surface_filepath: typing.Optional[types.Path] = None,
+        num_iterations: int = None,
+        num_samples: int = None,
+        sigma_scaling_factor: float = None,
     ) -> tasks.Task:
         """Simulates the scenario.
+
+        A simulation can be performed in two ways simply by adding some more
+        arguments. 
+        
+        If `num_iterations`, `num_samples` and `sigma_scaling_factor`
+        are all not provided, the simulation merely uses the created device and 
+        performs the computation of the magnetic field produced by its coils
+        on the given plasma surface. Additionally, it produces an output of
+        the objective functions that assess the quality of the device.
+
+        Alternatively, if the arguments are provided, the simulation process
+        is different. It also uses the create device, but as a beginning step.
+        Using this initial device, the simulation adds noise to the coefficients
+        that describe the coils using a normal distribution function, generating
+        `num_samples` configurations. After this, the configuration with the
+        lowest value of the objective functions is used as an initial step for
+        the next iteration, being that this process is repeated `num_iterations`
+        times. All the generated configurations (their Fourier Series 
+        coefficients) and their corresponding objective functions are an output
+        of this process.
 
         Args:
             simulator: The simulator to use for the simulation.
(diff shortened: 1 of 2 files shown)


inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -230,15 +230,49 @@ def simulate(
...
         """Simulates the scenario.
 
+        A simulation can be performed in two ways simply by adding some more
+        arguments. 
+        
+        If `num_iterations`, `num_samples` and `sigma_scaling_factor`
+        are all not provided, the simulation merely uses the created device and 
+        performs the computation of the magnetic field produced by its coils
+        on the given plasma surface. Additionally, it produces an output of
+        the objective functions that assess the quality of the device.
```
fabiocruz, line 246: A simpler, more natural way to provide this functionality is by setting defaults to these parameters, namely:
```python
num_samples = 1
num_iterations = 1
sigma_scaling_factor = 0.1
```

Then explain what happens when `num_samples` and `num_iterations` > 1. WDYT?

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -230,15 +230,49 @@ def simulate(
...
+
+        Alternatively, if the arguments are provided, the simulation process
+        is different. It also uses the create device, but as a beginning step.
+        Using this initial device, the simulation adds noise to the coefficients
+        that describe the coils using a normal distribution function, generating
+        `num_samples` configurations. After this, the configuration with the
+        lowest value of the objective functions is used as an initial step for
+        the next iteration, being that this process is repeated `num_iterations`
+        times. All the generated configurations (their Fourier Series
```
fabiocruz, line 255: "... used as an initial step for the next iteration. This process is repeated \`num_iterations\` times."

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -250,14 +284,32 @@ def simulate(
...
+
+        if len(deciding_args) not in (0, 3):
+            raise ValueError('If you want to perform the stellarator noise '
+                             'search simulation please provide all 3 '
+                             'necessary arguments. These include:\n '
+                             '"num_iterations", "num_samples" and '
+                             '"sigma_scaling_factor".\nOtherwise, please do '
+                             'not provide any of these.')
+
+        else:
```
fabiocruz, line 300: See my comment above. I think we don't need this after all.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -230,15 +230,49 @@ def simulate(
...
+        arguments. 
+        
+        If `num_iterations`, `num_samples` and `sigma_scaling_factor`
+        are all not provided, the simulation merely uses the created device and 
+        performs the computation of the magnetic field produced by its coils
+        on the given plasma surface. Additionally, it produces an output of
+        the objective functions that assess the quality of the device.
+
+        Alternatively, if the arguments are provided, the simulation process
+        is different. It also uses the create device, but as a beginning step.
```
fabiocruz, line 249: We need to explain with 1/2 more sentences what this "search" process is about. Shall we use the word "optimization"?

fabiocruz (changes requested):
Thanks @joaobernardo23. I think that setting the right default values could simplify a lot the logic of the scenario. See my comments below and let me know what you think.

inductiva/stellarators/scenarios/coils/coils.py
joaobernardo23: I totally agree, much simpler and much easier to understand, thank you!

fabiocruz (commented):
Suggestion to improve the docstring below.

fabiocruz (approved):

PR #483, opened 2023-08-23
[ProteinSolvation]Remove option to plot with full precision trajectory 

cristianacarpinteiro:
In this pull request, I have removed the option to render the trajectory with the full precision file. After testing, I saw that plotting with either the full precision file or the compressed one does not yield significant differences in the visualization quality.
The full precision file contains velocity, coordinates, and forces data, so to do more in depth analysis is better to use this one. In this file everything is saved with full precision. 
The compressed file contains only the coordinates and saves them with only 3 decimal places by default. The size of it is much smaller (around 200 times smaller). 
I have chosen to use only the xtc format to render the trajectory. While for the computation of the RMSF uses the full precision file to have a more accurate measure. 
Finally, I have made adjustments to the mdp files. I was saving every timestep of the simulation, resulting in a unnecessarily large xtc file. Now I am saving coordinates every 1 ps. I also chose to only save the protein coordinates and discard the water molecules positions. These changes make the  resulting file smaller.

### DIFF
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -2,7 +2,7 @@
 import os
 import pathlib
 import time
-import typing
+from typing import Literal
 
 import matplotlib.pyplot as plt
 import MDAnalysis as mda
@@ -11,8 +11,10 @@ import numpy as np
 
 import inductiva
 
+FULL_TRAJECTORY_FILE = "full_trajectory.trr"
 COMPRESSED_TRAJECTORY_FILE = "compressed_trajectory.xtc"
-TOPOLOGY_FILE = "protein.gro"
+PROTEIN_TOPOLOGY_FILE = "protein.gro"
+SYSTEM_TOPOLOGY_FILE = "solvated_protein.tpr"
 
 
 class ProteinSolvationOutput():
@@ -29,19 +31,18 @@ class ProteinSolvationOutput():
             sim_output_path: Path to the simulation output directory."""
 
         self.sim_output_dir = sim_output_path
-        topology_path = os.path.join(self.sim_output_dir, TOPOLOGY_FILE)
+        topology_path = os.path.join(self.sim_output_dir, PROTEIN_TOPOLOGY_FILE)
         trajectory_path = os.path.join(self.sim_output_dir,
                                        COMPRESSED_TRAJECTORY_FILE)
         self.universe = inductiva.molecules.scenarios.utils.unwrap_trajectory(
             topology_path, trajectory_path)
 
-    def render_interactive(
-            self,
-            representation: typing.Literal["cartoon", "ball+stick", "line",
-                                           "point", "surface",
-                                           "ribbon"] = "ball+stick",
-            selection: str = "protein",
-            add_backbone: bool = True):
+    def render_interactive(self,
+                           representation: Literal["cartoon", "ball+stick",
+                                                   "line", "point", "surface",
+                                                   "ribbon"] = "ball+stick",
+                           selection: str = "protein",
+                           add_backbone: bool = True):
         """
         Render the simulation outputs in an interactive visualization.
         Args: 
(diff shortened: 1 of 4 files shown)


david451lima:
LGTM!
Just a question: How does the compressed file is so much smaller than the regular trajectory? If we have a bunch of parameters N, shouldn't the file be smaller by a factor of 3/N ?

david451lima (approved):
LGTM

inductiva/templates/protein_solvation/gromacs/simulation.mdp.jinja
```diff
@@ -4,11 +4,12 @@ nsteps                  = {{ nsteps }}       ; number of steps in simulation
 dt                      = 0.002              ; 2 fs
 
 ; Output control
-nstxtcout               = 1                   ; generate .xtc file 
-nstxout                 = 500                ; save coordinates every 500 iter
+nstxout                 = 500                ; save coordinates every 1.0 ps
```
fabiocruz, line 7: This should be configurable. We usually call this `output_time_step` (to make a distinction with the simulation `time_step`, which should also be configurable). In this case, should it be called `output_time_step_ps` to abide by our previous discussion on parameter units?

inductiva/templates/protein_solvation/gromacs/simulation.mdp.jinja
```diff
@@ -4,11 +4,12 @@ nsteps                  = {{ nsteps }}       ; number of steps in simulation
...
 nstvout                 = 500                ; save velocities every 1.0 ps
 nstenergy               = 500                ; save energies every 1.0 ps
 nstlog                  = 500                ; update log file every 1.0 ps
+nstxout-compressed      = 500                ; save coordinates every 1.0 ps in the compressed trajectory file
```
fabiocruz, line 11: The same would be used here.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -104,13 +84,16 @@ def calculate_rmsf_trajectory(self,
...
+        full_trajectory_path = os.path.join(self.sim_output_dir,
+                                            FULL_TRAJECTORY_FILE)
+        full_precision_universe = mda.Universe(topology_path,
+                                               full_trajectory_path)
 
         aligned_trajectory_path = os.path.join(self.sim_output_dir,
                                                "aligned_traj.dcd")
         inductiva.molecules.scenarios.utils.align_trajectory_to_average(
-            universe, aligned_trajectory_path)
+            full_precision_universe, aligned_trajectory_path)
```
fabiocruz, line 96: The PR description mentions that the representation is now done using the compressed trajectory, but this uses the "full precision universe". Can you elaborate on why?

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -13,10 +13,11 @@
 
 FULL_TRAJECTORY_FILE = "full_trajectory.trr"
 COMPRESSED_TRAJECTORY_FILE = "compressed_trajectory.xtc"
-TOPOLOGY_FILE = "solvated_protein.tpr"
+PROTEIN_TOPOLOGY_FILE = "protein.gro"
+SYSTEM_TOPOLOGY_FILE = "solvated_protein.tpr"
 
 
-class ProteinSolvationOutput:
+class ProteinSolvationOutput():
```
fabiocruz, line 20: No need for this `()`. I'm actually not sure that this works as is.

fabiocruz (approved):
Looking good, @cristianacarpinteiro!

inductiva/templates/protein_solvation/gromacs/simulation.mdp.jinja
cristianacarpinteiro: yes, I was thinking the same! I will expose it to the user then.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: Because in this function I am computing the RMSF measure, so I thought that it would be better to use the full precision trajectory to get a more precise measure.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
cristianacarpinteiro: It works that I tested it but I will correct it :)

inductiva/molecules/scenarios/protein_solvation/post_processing.py
fabiocruz: Gotcha, thanks.

PR #484, opened 2023-08-23
Change order of the parameters of plate class + typos

Joao97ribeiro:
This PR includes two changes:
1. `Reordered the parameters of the plate class`, so that the width comes first and then the length. This change is intended to enhance intuitiveness, aligning with the convention of specifying width before length (similar to x before y).
2. `Corrected typos`: Renamed the parameter 'holes' to 'list_holes'.


fabiocruz (approved):
LGTM.

PR #489, opened 2023-08-25
[Coastal area scenario] Add method to plot bathymetries

fabiocruz:
This PR adds a simple method to plot bathymetries using matplotlib, e.g.

```python
import numpy as np
import inductiva

bathymetry = inductiva.fluids.scenarios.coastal_area.Bathymetry(
    depths=np.random.uniform(-10, 10, size=(200)),
    x=np.random.uniform(0, 800, size=(200,)),
    y=np.random.uniform(0, 1000, size=(200,)),
)

bathymetry.plot(output_path)
```

bathymetry_random

I also moved the bathymetry class to its own file.

This is an important step toward being able to process standard bathymetry files before they can be consumed by simulators.


fabiocruz:
See #487 for context, if needed.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -0,0 +1,223 @@
+"""Bathymetry class and utilities."""
+
+import random
+from typing import Optional, Sequence, Tuple, Union
+
+import matplotlib
```
cristianacarpinteiro, line 6: why not do `from matplotlib.pyplot as plt `, does it go against the google style guide?

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -0,0 +1,223 @@
...
+        """Plots the bathymetry.
+
+        The bathymetry is represented as a 2D map of depths, with the x and y
+        coordinates of the points where the depths are defined in the axes.
+
+        The plot is produced with matplotlib.
+    
+        Args:
+            cmap: Colormap to use. Defaults to the matplotlib default colormap.
+            clim: Color limits.
```
cristianacarpinteiro, line 193: Maybe explain this better, the range of values you want to plot. Btw, why did you expose it to the user? Is there interest to show values in other ranges other than (min, max) ?

cristianacarpinteiro (approved):
LGTM!

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: I think either `import matplotlib` or `import matplotlib.pyplot as plt` are according to the Google Python style guide. In this file in particular I need to access `matplotlib` (to get the axes class, `matplotlib.axes.Axes`) and `matplotlib.pyplot` (to create and close the figure), so I decided to go with one import only.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: I'll add more info.

I think this parameter is useful to make sure that colors are the same in different plots, so I tend to add to all plots of this kind. Users will let us know, anyway :)

inductiva/fluids/scenarios/coastal_area/bathymetry.py
cristianacarpinteiro: Makes sense :)

Issue #478, opened 2023-08-22
[Coastal area scenario] Add ability to load bathymetry from standard file format


fabiocruz:
The following describes the problem with the current implementation of the coastal area scenario and the associated bathymetry class, as well as a tentative solution to solve this problem. I would like to collect feedback on this proposal before implementing anything. Thanks!

## Problem

The current implementation of the coastal area scenario requires that the user loads a text file with a grid of depth values defining the bathymetry to be used in the simulation.

While this in theory allows the user to use different bathymetries in simulations, it is not practical, since bathymetry data from *standard* sources / file formats cannot be used.

The main challenge in a user workflow thus seems to be loading standard bathymetry data and converting it to a format that can be consumed by simulators such as SWASH and XBeach.

The challenges of loading standard bathymetry data are:
1. There is no single *standard* source / file format. There are centralized repositories for this data, e.g. [European Marine Observation and Data Network  (EMODnet)](https://emodnet.ec.europa.eu/geoviewer/#!/), which follows an European policy and thus has some level of control over the format in which bathymetry data is stored. 
2. Bathymetry data is usually collected at a sample of locations that are not necessarily a) uniformly distributed in space, b) labeled with cartesian coordinates instead of latitude, longitude pairs, etc., as required by SWASH.

## Tentative solution

Ideally, we would be able to load bathymetries, inspect them and pre-process them before they can be passed to a simulator.

### File format

I suggest we start by adding support to bathymetries provided in the ASCII XYZ format defined [here](https://emodnet.ec.europa.eu/sites/emodnet.ec.europa.eu/files/public/20171127_DTM_exchange_format_specification_v1.6.pdf). This is a format provided in many of the datasets in EMODnet, see e.g. [this data from Algarve](https://sextant.ifremer.fr/record/SDN_CPRD_590_HR_Lidar_Algarve/).

### Bathymetry class

Due to the different definition of the bathymetry data in typical sources (scattered vs grid values, etc.), our `Bathymetry` class needs to be updated.

I suggest we allow loading a bathymetry with "no further questions asked", i.e.
```python
bathymetry = inductiva.....Bathymetry.from_ascii_xyz_file(file_path)
```

The `Bathymety` class should:
1. Store information as read from the file, i.e. a list of coordinates and corresponding depths, and not arbitrarily crop it/interpolate it to a grid.
2. Contain a visualization method that allows inspecting the loaded bathymetry.
3. Contain a cropping method that allows selecting only sub-regions.
[...]

PR #491, opened 2023-08-25
[Coastal area scenario] Add support to read XYZ ASCII bathymetry files

fabiocruz:
This PR adds a classmethod to the `Bathymetry` class that allows reading XYZ ASCII bathymetry files. See #478 for context.

Example code to read some bathymetry from the Sines area available [here](https://sextant.ifremer.fr/record/SDN_CPRD_590_HR_Sines/):
```python
import inductiva

bathymetry = inductiva.fluids.scenarios.coastal_area.Bathymetry.from_ascii_xyz_file("590_HR_Sines.emo")
bathymetry.plot(output_path)
```

bathymetry_demo

### DIFF
--- a/inductiva/fluids/scenarios/coastal_area/bathymetry.py
+++ b/inductiva/fluids/scenarios/coastal_area/bathymetry.py
@@ -5,6 +5,7 @@ from typing import Optional, Sequence, Tuple, Union
 
 import matplotlib
 import numpy as np
+import utm
 
 import inductiva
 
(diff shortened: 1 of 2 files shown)


inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -68,6 +69,60 @@ def from_bot_file(
...
+        # corresponding to longitude, latitude and average depth, respectively.
+        bathymetry_data = np.loadtxt(ascii_xyz_file_path,
+                                     usecols=(0, 1, 4),
+                                     delimiter=";")
+
+        x_lon = bathymetry_data[:, 0]
+        y_lat = bathymetry_data[:, 1]
+        depths = bathymetry_data[:, 2]
+
+        x, y, _, _ = utm.from_latlon(y_lat, x_lon)
```
IvanPombo, line 118: Can you comment for us this UTM a bit more. From a quick read it seems that is just a projection from the longitude and latitude to a plane. So it keeps the exact same tilling ("mesh") correct?

IvanPombo (approved):
LGTM!

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: Yes, it projects latitude and longitude to a x, y plane. It does keep the same tiling. However, because latitude and longitude are coordinates on a sphere, when projected to a x, y plane, a uniform grid in latitude, longitude does not translate to a uniform grid in x, y.

PR #494, opened 2023-08-28
[Coastal area scenario] Plot bathymetries on uniform grid

fabiocruz:
While the plotting method introduced in #489 works for bathymetries defined over regions with an approximately rectangular overall shape, it failed to accurately represent bathymetries defined over complex regions, e.g. defined over regions with concave borders (or hulls).

In this PR I introduce changes to the plotting method to circumvent this difficulty.

Below I use an example bathymetry (from the region of Algarve, taken from [here](https://sextant.ifremer.fr/record/SDN_CPRD_590_HR_Lidar_Algarve/)) to illustrate the problem and the solution I found to it.

Here is a representation of the x, y coordinates where the bathymetry is defined:

bathymetry_demo_scatter

As you can see, there is a lot of white space in the center of the figure.

When plotting this with the previous implementation, we would get something similar to the figure below:

bathymetry_demo_hull

I overlayed here a red scatter marking the locations where the bathymetry is defined (as shown in the previous figure) and a black line marking the [convex hull](https://en.wikipedia.org/wiki/Convex_hull) containing all those locations.

The problem is that matplotlib plotting methods such as [`tripcolor`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.tripcolor.html) determines the area to color using the convex hull that contains all the points.

The solution I implement interpolates the bathymetry data to a uniform grid and filters out of that grid any point that is at a distance from the locations where the bathymetry is defined larger than a given threshold.

bathymetry_demo_imshow

The size of the grid to which depths are interpolated and the threshold distance are parameters (with reasonable defaults, but which may require tuning in the future) in the plot method.

This allows the user to get a more accurate representation of the bathymetry data they are reading/exploring.

### DIFF
--- a/inductiva/fluids/scenarios/coastal_area/bathymetry.py
+++ b/inductiva/fluids/scenarios/coastal_area/bathymetry.py
@@ -285,8 +285,13 @@ class Bathymetry:
 
         # Filter out points that are far from bathymetry locations.
         tree = scipy.spatial.KDTree(np.c_[self.x, self.y])
+
+        # Obtain distance between each point on the uniform grid and the
+        # closest bathymetry location.
         distance, _ = tree.query(np.c_[x_grid.ravel(), y_grid.ravel()], k=1)
         distance = distance.reshape(x_grid.shape)
+
+        # Set depths to NaN for points that are far from bathymetry locations.
         depths_grid[distance > max_distance] = np.nan
 
         # Plot the bathymetry.
--- a/inductiva/fluids/scenarios/coastal_area/bathymetry.py
+++ b/inductiva/fluids/scenarios/coastal_area/bathymetry.py
@@ -5,6 +5,7 @@ from typing import Optional, Sequence, Tuple, Union
 
 import matplotlib
 import numpy as np
+import scipy
 import utm
 
 import inductiva
@@ -235,14 +236,26 @@ class Bathymetry:
         cmap: Optional[str] = None,
         clim: Optional[Tuple[float]] = None,
         path: Optional[str] = None,
+        grid_size: int = 500,
+        max_distance: float = 20,
     ) -> Union[matplotlib.axes.Axes, None]:
         """Plots the bathymetry.
 
         The bathymetry is represented as a 2D map of depths, with the x and y
         coordinates of the points where the depths are defined in the axes.
 
-        The plot is produced with matplotlib.
+        The bathymetry data is plotted with a color plot on a uniform grid,
+        defined by the size `grid_size`.
+        
+        The data is interpolated from the points where the bathymetry is defined
+        to the uniform grid using linear interpolation.
+
+        Points on the uniform grid at a distance to points where the bathymetry
+        is defined larger than a threshold distance `max_distance` are not
+        plotted.
     
+        The plot is produced with matplotlib.
+
         Args:
             cmap: Colormap to use. Defaults to the matplotlib default colormap.
             clim: Range of depth values to represent in colors. If `None`, the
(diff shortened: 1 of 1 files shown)


inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -253,16 +266,43 @@ def plot(
               the matplotlib `Axes` object is returned instead.
         """
 
+        # Create uniform grid for interpolation.
+        x_grid, y_grid = np.meshgrid(np.linspace(*self.x_range, grid_size),
+                                     np.linspace(*self.y_range, grid_size))
+
+        # Interpolate depths to uniform grid.
```
IvanPombo, line 273: I am not sure,  if this will be the right place to perform this. Given that this will be used by the simulator as well. Wdyt?

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -253,16 +266,43 @@ def plot(
...
+        depths_grid = interpolator(x_grid, y_grid)
+
+        # Filter out points that are far from bathymetry locations.
+        tree = scipy.spatial.KDTree(np.c_[self.x, self.y])
+        distance, _ = tree.query(np.c_[x_grid.ravel(), y_grid.ravel()], k=1)
+        distance = distance.reshape(x_grid.shape)
+        depths_grid[distance > max_distance] = np.nan
+
+        # Plot the bathymetry.
+        extent = (
```
IvanPombo, line 288: Extent or ranges?

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -253,16 +266,43 @@ def plot(
...
+        interpolator = scipy.interpolate.LinearNDInterpolator(
+            (self.x, self.y),
+            self.depths,
+        )
+
+        depths_grid = interpolator(x_grid, y_grid)
+
+        # Filter out points that are far from bathymetry locations.
+        tree = scipy.spatial.KDTree(np.c_[self.x, self.y])
+        distance, _ = tree.query(np.c_[x_grid.ravel(), y_grid.ravel()], k=1)
```
IvanPombo, line 283: I feel that this is very weird. What distance are you obtaining here?

IvanPombo (approved):
LGTM, I just left minor comments for readibility and one main question of where the interpolator should live.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: I agree that it doesn't feel right to have the interpolation code here. I am still figuring out the best place to put this... maybe under some interpolation utilities?

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: Extent, to follow matplotlib's nomenclature.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: I am obtaining the distance between all points defining the tree, (i.e. all points where the bathymetry is defined,`(self.x, self.y)`) and all points of the uniform grid, `(x_grid.ravel(), y_grid.ravel())`.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -250,9 +250,9 @@ def plot(
         The data is interpolated from the points where the bathymetry is defined
         to the uniform grid using linear interpolation.
 
-        Points on the uniform grid at a distance to points where the bathymetry
-        is defined larger than a threshold distance `max_distance` are not
-        plotted.
+        Points on the uniform grid at a distance larger than a threshold
+        distance `max_distance` from points where the bathymetry is defined are
```
IvanPombo, line 254: Then, shouldn't it be `distance_threshold`?

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -280,8 +285,13 @@ def plot(
 
         # Filter out points that are far from bathymetry locations.
         tree = scipy.spatial.KDTree(np.c_[self.x, self.y])
+
+        # Obtain distance between each point on the uniform grid and the
+        # closest bathymetry location.
         distance, _ = tree.query(np.c_[x_grid.ravel(), y_grid.ravel()], k=1)
         distance = distance.reshape(x_grid.shape)
+
+        # Set depths to NaN for points that are far from bathymetry locations.
```
IvanPombo, line 294: Top!!

IvanPombo (approved):

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: Right, but this plotting utility is meant to be used for arbitrary bathymetries, even those that can't be used to run SWASH simulations. The goal of the utility is exactly to visualize the bathymetry, so the user can inspect which region to crop and simulate.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: It could be, yes! Thanks. I'll update this in a new PR.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
IvanPombo: Yeah, I see that point. If needed we can adapt later!

PR #499, opened 2023-08-29
Add MachineGroup to the Simulators

IvanPombo:
This PR serves to remove the resource_pool_id from all types of simulators and substitute by the MachineGroup that is already deployed to the Cloud.

The changes descends the inheritance ladder up until the function run_simulation in tasks/run_simulation.py. Here, we keep the concept of resource_pool_id to still identify to the which machines the simulation should go.

Current deployment is under testing to find inconsistencies and bugs!

E.g. of usage now:

```python
machine_group = inductiva.resources.MachineGroup(
    machine_type="c2-standard-16",
    num_machines=2,
    disk_size_gb=30,
    zone="europe-west1-b")

machine_group.start()

scenario = inductiva.molecules.scenarios.ProteinSolvation(protein_pdb = "protein.pdb") 
task = scenario.simulate(simulation_time_ns=0.1,
                                           run_async=False,
                                           machine_group=machine_group)

task.download_outputs()
```

Verify if the scenarios and simulators you own work as expected.

### DIFF
--- a/inductiva/fluids/scenarios/coastal_area/coastal_area.py
+++ b/inductiva/fluids/scenarios/coastal_area/coastal_area.py
@@ -101,7 +101,7 @@ class CoastalArea(Scenario):
 
         Args:
             simulator: Simulator to use. Supported simulators are: SWASH.
-            machine_group: The MachineGroup to use for the simulation.
+            machine_group: The machine group to use for the simulation.
             simulation_time: Total simulation time, in seconds.
             time_step: Time step, in seconds.
             output_time_step: Time step for the output, in seconds.
--- a/inductiva/fluids/scenarios/coastal_area/coastal_area.py
+++ b/inductiva/fluids/scenarios/coastal_area/coastal_area.py
@@ -5,9 +5,8 @@ import math
 import os
 import shutil
 from typing import Literal, Optional
-from uuid import UUID
 
-from inductiva import tasks
+from inductiva import tasks, resources
 from inductiva.scenarios import Scenario
 from inductiva.simulation import Simulator
 from inductiva.fluids.simulators import SWASH
@@ -91,7 +90,7 @@ class CoastalArea(Scenario):
     def simulate(
         self,
         simulator: Simulator = SWASH(),
-        resource_pool_id: Optional[UUID] = None,
+        machine_group: Optional[resources.MachineGroup] = None,
         run_async: bool = False,
         simulation_time: float = 100,
         time_step: float = 0.1,
@@ -102,7 +101,7 @@ class CoastalArea(Scenario):
 
         Args:
             simulator: Simulator to use. Supported simulators are: SWASH.
-            resource_pool_id: Resource pool to use for the simulation.
+            machine_group: The MachineGroup to use for the simulation.
             simulation_time: Total simulation time, in seconds.
             time_step: Time step, in seconds.
             output_time_step: Time step for the output, in seconds.
@@ -115,7 +114,7 @@ class CoastalArea(Scenario):
 
         task = super().simulate(
             simulator,
-            resource_pool_id=resource_pool_id,
+            machine_group=machine_group,
             run_async=run_async,
             sim_config_filename=SWASH_CONFIG_FILENAME,
             n_cores=n_cores,
--- a/inductiva/fluids/scenarios/dam_break/dam_break.py
+++ b/inductiva/fluids/scenarios/dam_break/dam_break.py
@@ -2,11 +2,10 @@
 from typing import List, Literal, Optional
 from enum import Enum
 from dataclasses import dataclass
-from uuid import UUID
 
-from inductiva import tasks
+from inductiva import tasks, resources
 from inductiva.simulation import Simulator
-from inductiva.fluids.simulators import DualSPHysics
+from inductiva.fluids.simulators import SPlisHSPlasH
 from inductiva.fluids.scenarios.fluid_block import FluidBlock
 from inductiva.fluids.fluid_types import FluidType
 from inductiva.fluids.fluid_types import WATER
@@ -52,9 +51,9 @@ class DamBreak(FluidBlock):
     # pylint: disable=arguments-renamed
     def simulate(
         self,
-        simulator: Simulator = DualSPHysics(),
-        resource_pool_id: Optional[UUID] = None,
-        device: Literal["cpu", "gpu"] = "gpu",
+        simulator: Simulator = SPlisHSPlasH(),
+        machine_group: Optional[resources.MachineGroup] = None,
+        device: Literal["cpu", "gpu"] = "cpu",
         resolution: Literal["high", "medium", "low"] = "medium",
         simulation_time: float = 1,
         run_async: bool = False,
@@ -63,6 +62,7 @@ class DamBreak(FluidBlock):
 
         Args:
             simulator: Simulator to use.
+            machine_group: The MachineGroup to use for the simulation.
             device: Device in which to run the simulation.
             resolution: Resolution of the simulation.
             simulation_time: Simulation time, in seconds.
@@ -72,7 +72,7 @@ class DamBreak(FluidBlock):
         particle_radius = ParticleRadius[resolution.upper()].value
 
         task = super().simulate(simulator=simulator,
-                                resource_pool_id=resource_pool_id,
+                                machine_group=machine_group,
                                 device=device,
                                 particle_radius=particle_radius,
                                 simulation_time=simulation_time,
--- a/inductiva/fluids/scenarios/dam_break/dam_break.py
+++ b/inductiva/fluids/scenarios/dam_break/dam_break.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass
 
 from inductiva import tasks, resources
 from inductiva.simulation import Simulator
-from inductiva.fluids.simulators import SPlisHSPlasH
+from inductiva.fluids.simulators import DualSPHysics
 from inductiva.fluids.scenarios.fluid_block import FluidBlock
 from inductiva.fluids.fluid_types import FluidType
 from inductiva.fluids.fluid_types import WATER
@@ -51,9 +51,9 @@ class DamBreak(FluidBlock):
     # pylint: disable=arguments-renamed
     def simulate(
         self,
-        simulator: Simulator = SPlisHSPlasH(),
+        simulator: Simulator = DualSPHysics(),
         machine_group: Optional[resources.MachineGroup] = None,
-        device: Literal["cpu", "gpu"] = "cpu",
+        device: Literal["cpu", "gpu"] = "gpu",
         resolution: Literal["high", "medium", "low"] = "medium",
         simulation_time: float = 1,
         run_async: bool = False,
@@ -62,7 +62,7 @@ class DamBreak(FluidBlock):
 
         Args:
             simulator: Simulator to use.
-            machine_group: The MachineGroup to use for the simulation.
+            machine_group: The machine group to use for the simulation.
             device: Device in which to run the simulation.
             resolution: Resolution of the simulation.
             simulation_time: Simulation time, in seconds.
--- a/inductiva/fluids/scenarios/fluid_tank/fluid_tank.py
+++ b/inductiva/fluids/scenarios/fluid_tank/fluid_tank.py
@@ -4,7 +4,6 @@ from enum import Enum
 from functools import singledispatchmethod
 import os
 from typing import List, Literal, Optional
-from uuid import UUID
 
 from inductiva import tasks, resources
 from inductiva.scenarios import Scenario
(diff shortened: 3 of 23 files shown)


inductiva/tasks/run_simulation.py
```diff
@@ -1,16 +1,15 @@
...
-from inductiva import tasks
+from inductiva import tasks, resources
 from inductiva.api import methods
 
 
 def run_simulation(
     api_method_name: str,
     input_dir: pathlib.Path,
-    resource_pool_id: Optional[UUID] = None,
+    machine_group: resources.MachineGroup = None,
```
fabiocruz, line 12: You are missing the `Optional[...]` type annotation.

inductiva/structures/simulators/fenicsx.py
```diff
@@ -32,12 +31,12 @@ def run(
             mesh_filename: Mesh filename.
             bcs_filename: Boundary conditions filename.
             material_filename: Material filename.
-            resource_pool_id: Optional UUID of the resource pool to use.
+            machine_group: The MachineGroup to use for the simulation.
```
fabiocruz, line 34: No need to use the class name here, "The machine group to use...".

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -256,7 +256,7 @@ def simulate(
 
         Args:
             simulator: The simulator to use for the simulation.
-            resource_pool_id: The resource pool to use for the simulation.
+            machine_group: The MachineGroup to use for the simulation.
```
fabiocruz, line 259: "The machine group..."

inductiva/simulation/simulator.py
```diff
@@ -29,7 +25,7 @@ def run(
         self,
         input_dir: types.Path,
         *_args,
-        resource_pool_id: Optional[UUID] = None,
+        machine_group: resources.MachineGroup = None,
```
fabiocruz, line 28: `Optional[...]` type annotation missing.

sarmento:
I have a couple of question regarding the example, so I am leaving them there for further discussion with Hugo. Maybe these are non-issues but I am trying to think from the point of view of the user, and therefore pls take these as suggestions for future dev:

```
machine_group = inductiva.resources.MachineGroup(
    machine_type="c2-standard-16",
    num_machines=2,
    disk_size_gb=30,
    zone="europe-west1-b")

machine_group.start()
```

It's not obvious how this start() method is going to behave. Is this blocking? Should this be blocking/non-blocking? Should this return something (e.g. a summary of what is going on).

I would expect this to be blocking and behave in this fashion:

```
stats = machine_group.start()
logging.info("Starting %d machines took %s secs", stats{num_machines}, stats{completion_time})
```

or you could make it non-blocking and do something like

```
machine_group.start()
while(!machine_group.is_ready()):
    logging.info("Waiting...")
```

either way, I feel that just having start() leaves too much ambiguity.

Here:

```
scenario = inductiva.molecules.scenarios.ProteinSolvation(protein_pdb = "protein.pdb") 
task = scenario.simulate(simulation_time_ns=0.1,
                                           run_async=False,
                                           machine_group=machine_group)

```

we have another potential problem. As it is, it seems that the default mode is running async, so you need to explicitly set run_async to False. Not sure if this is what I would expect. Also, when does this unblock? 

It is also not clear why you have 2 machines in this machine group? How would the scenario use this in a blocking setting? Do we share the load?


There are also issues here:
```
task.download_outputs()
```
I am assuming this is blocking (the default should be), but should this return something? How can we also hint in the docs that this can take a long time....

Please see these as ways of helping all of you thinking from the point of view of the user and improving our documentation.

inductiva/resources/machines.py
```diff
@@ -124,4 +132,6 @@ def _log_machine_group_info(self):
         logging.info("Number of machines: %s", self.num_machines)
         logging.info("Spot: %s", self.spot)
         logging.info("Disk size: %s GB", self.disk_size_gb)
-        logging.info("Estimated cost per hour: %s $/h", self.estimated_price)
+        # We won't offer price values just yet, due to a bug.
+        # TODO: Fix the price computation.
+        # logging.info("Estimated cost per hour: %s $/h", self.estimated_price)
```
fabiocruz, line 137: Why is this change introduced here? Do it in a separate PR, to describe the bug, etc?

inductiva/resources/machines.py
```diff
@@ -84,17 +85,24 @@ def terminate(self):
...
-                    body=inductiva.client.model.instance.Instance(
-                        id=self.id, zone=self.zone))
-                termination_time_mins = (time.time() - start_time) / 60
+                api_instance.delete_instance_group(body=instance_group_config)
+                termination_time_mins = time.time() - start_time
                 logging.info(
-                    "Machine group of %s machines successfully"
-                    "terminated in %.2f mins.", self.num_machines,
+                    "Machine group of %s machines successfully "
+                    "terminated in %.2f s.", self.num_machines,
```
fabiocruz, line 105: You updated the string to indicate that the time is measured in seconds, but the variable is still `termination_time_mins`.

inductiva/resources/machines.py
```diff
@@ -84,17 +85,24 @@ def terminate(self):
...
+                    name=self.name,
+                    machine_type=self.machine_type,
+                    num_instances=self.num_machines,
+                    spot=self.spot,
+                    disk_size_gb=self.disk_size_gb,
+                    zone=self.zone,
+                )
             try:
                 logging.info("Terminating machine group."
                              "This may take a few minutes.")
```
fabiocruz, line 99: Since this may take a few minutes, maybe keep the log in minutes?

inductiva/resources/machines.py
```diff
@@ -64,13 +64,14 @@ def start(self):
...
-                creation_time_mins = (time.time() - start_time) / 60
+                creation_time_mins = time.time() - start_time
 
                 self.id = instance_group.body["id"]
+                self.name = instance_group.body["name"]
                 #self.estimated_price = self._compute_estimated_price(
                 #    api_instance)
 
-                logging.info("Machine group successfully created in %.2f mins.",
+                logging.info("Machine group successfully created in %.2f s.",
```
fabiocruz, line 74: The same comments apply here (logging in minutes vs seconds).

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -56,6 +55,7 @@ def simulate(
         """Simulate the solvation of a protein.
 
         Args:
+            machine_group: The MachineGroup to use for the simulation.
```
fabiocruz, line 58: "The machine group to..."

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -129,7 +128,7 @@ def simulate(self,
             n_cores: Number of cores to use for the simulation.
             resolution: Level of detail of the mesh used for the simulation.
                 Options: "high", "medium" or "low".
-            resource_pool_id: Id of the resource pool to use for the simulation.
+            machine_group: The MachineGroup to use for the simulation.
```
fabiocruz, line 131: "The machine group to..."

inductiva/fluids/scenarios/fluid_tank/fluid_tank.py
```diff
@@ -194,7 +193,7 @@ def simulate(
 
         Args:
             simulator: Simulator to use. Supported simulators are: SPlisHSPlasH.
-            resource_pool_id: Resource pool to use for the simulation.
+            machine_group: The MachineGroup to use for the simulation.
```
fabiocruz, line 196: "The machine group to..."

inductiva/fluids/scenarios/dam_break/dam_break.py
```diff
@@ -52,9 +51,9 @@ def __init__(
     # pylint: disable=arguments-renamed
     def simulate(
         self,
-        simulator: Simulator = DualSPHysics(),
-        resource_pool_id: Optional[UUID] = None,
-        device: Literal["cpu", "gpu"] = "gpu",
+        simulator: Simulator = SPlisHSPlasH(),
```
fabiocruz, line 54: Why are you updating the default simulator in this scenario?

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -102,7 +101,7 @@ def simulate(
 
         Args:
             simulator: Simulator to use. Supported simulators are: SWASH.
-            resource_pool_id: Resource pool to use for the simulation.
+            machine_group: The MachineGroup to use for the simulation.
```
fabiocruz, line 104: "The machine group to..."

fabiocruz (approved):
LGTM. Got minor comments only. There's a couple of things I think should be done in a separate PR, please see my comments below.

fabiocruz:
> It is also not clear why you have 2 machines in this machine group? How would the scenario use this in a blocking setting? Do we share the load?

I agree with @sarmento here in that it's not crystal clear that we are sending this task to be run by a group of machines, but that that group of machines operates "independently", i.e. not sharing the load. I feel that the previous nomenclature `ResourcePool` gives a more intuitive description of the actual underlying structure. Perhaps we could find a name in between the old and new nomenclatures, e.g. `MachinePool` (and later convert it to `ResourcePool`, when we can configure more than simply the machines?).

IvanPombo:
> I have a couple of question regarding the example, so I am leaving them there for further discussion with Hugo. Maybe these are non-issues but I am trying to think from the point of view of the user, and therefore pls take these as suggestions for future dev:
> 
> ```
> machine_group = inductiva.resources.MachineGroup(
>     machine_type="c2-standard-16",
>     num_machines=2,
>     disk_size_gb=30,
>     zone="europe-west1-b")
> 
> machine_group.start()
> ```
> 
> It's not obvious how this start() method is going to behave. Is this blocking? Should this be blocking/non-blocking? Should this return something (e.g. a summary of what is going on).

This start() method is going a blocking call to launch the machines and make sure that when the simulations are deployed they are already ready. After they are launched, we receive a set of information about the resources launched.
 
> I would expect this to be blocking and behave in this fashion:
> 
> ```
> stats = machine_group.start()
> logging.info("Starting %d machines took %s secs", stats{num_machines}, stats{completion_time})
> ```
> 
> or you could make it non-blocking and do something like
> 
> ```
> machine_group.start()
> while(!machine_group.is_ready()):
>     logging.info("Waiting...")
> ```
> 
> either way, I feel that just having start() leaves too much ambiguity.
> 
> Here:
> 
> ```
> scenario = inductiva.molecules.scenarios.ProteinSolvation(protein_pdb = "protein.pdb") 
> task = scenario.simulate(simulation_time_ns=0.1,
>                                            run_async=False,
>                                            machine_group=machine_group)
> ```
> 
> we have another potential problem. As it is, it seems that the default mode is running async, so you need to explicitly set run_async to False. Not sure if this is what I would expect. Also, when does this unblock?

The default is run_async=False, here I just wanted to show off that I was using it. It unblocks when the simulations ends running (no download).

> It is also not clear why you have 2 machines in this machine group? How would the scenario use this in a blocking setting? Do we share the load?

No, I just want to exemplify the launch of more than one machine. We have no load sharing at this moment in time. Two machines for two independent simulations.

> There are also issues here:
> 
> ```
> task.download_outputs()
> ```
> 
> I am assuming this is blocking (the default should be), but should this return something? How can we also hint in the docs that this can take a long time....

This just downloads things and informs the user about the specific size of the output and shows a progress bar.

> Please see these as ways of helping all of you thinking from the point of view of the user and improving our documentation.

We will add all of this information more explicitly to the documentation in this new version. Thanks, for the feedback!

inductiva/resources/machines.py
IvanPombo: This change is done here, since it is related with the new MachineGroup class being deployed and does not let machines to be launched.

inductiva/fluids/scenarios/dam_break/dam_break.py
IvanPombo: I will revert it, it was due to testing.

IvanPombo:
> > It is also not clear why you have 2 machines in this machine group? How would the scenario use this in a blocking setting? Do we share the load?
> 
> I agree with @sarmento here in that it's not crystal clear that we are sending this task to be run by a group of machines, but that that group of machines operates "independently", i.e. not sharing the load. I feel that the previous nomenclature `ResourcePool` gives a more intuitive description of the actual underlying structure. Perhaps we could find a name in between the old and new nomenclatures, e.g. `MachinePool` (and later convert it to `ResourcePool`, when we can configure more than simply the machines?).

I see your points about the nomenclature and thinking more carefully I agree with the idea that MachineGroup seems to refer to a cluster where a single simulation can run. I will rephrase everything to be MachinePool if everyone agrees (@hpenedones @sarmento @begaiym-k)?

IvanPombo:
I will merge this PR with the MachineGroup name, since the concept we wanted is achieved. Let's discuss the naming convention in the following issue #501.

PR #500, opened 2023-08-29
Add function to list and get user's tasks

luispcunha:
As part of improving managing tasks in the API, this PR introduces two new functions.
Usage:
```python
# list the last 5 tasks that were successful
inductiva.tasks.list(5, status="success")
ID                   Simulator    Status       Submitted            Started              Duration     VM Type
1691150776862178362  openfoam     success      04 Aug, 12:06:17     04 Aug, 12:06:18     0h 1m 53s    c2-standard-8
1691149904961476240  openfoam     success      04 Aug, 11:51:46     04 Aug, 11:51:46     0h 1m 28s    c2-standard-8
1691081881158823776  openfoam     success      03 Aug, 16:58:02     03 Aug, 16:58:02     0h 1m 20s    n2-standard-32
1691081409916619414  openfoam     success      03 Aug, 16:50:11     03 Aug, 16:50:11     0h 1m 20s    n2-standard-32
1691080520213617518  openfoam     success      03 Aug, 16:35:21     03 Aug, 16:35:21     0h 1m 23s    n2-standard-32
```
```python
# get the last 5 tasks that haven't started yet and kill them
tasks = inductiva.tasks.get(5, status="submitted")
for task in tasks:
  task.kill()
```

### DIFF
--- a/inductiva/__init__.py
+++ b/inductiva/__init__.py
@@ -18,6 +18,7 @@ from . import templates
 from . import utils
 from . import stellarators
 from . import world
+from inductiva.tasks import Task
 
 api_url = os.environ.get("INDUCTIVA_API_URL", "http://api.inductiva.ai")
 output_dir = os.environ.get("INDUCTIVA_OUTPUT_DIR", "inductiva_output")
--- a/inductiva/tasks/methods.py
+++ b/inductiva/tasks/methods.py
@@ -1,79 +1,17 @@
 """Methods to interact with the tasks submitted to the API."""
-import pathlib
-from typing import Dict, List, Optional, Type
-
-from absl import logging
+import json
+from typing import Dict, List, Optional
 
 import inductiva
 from inductiva import api
 from inductiva.client import ApiClient, ApiException
 from inductiva.client.apis.tags.tasks_api import TasksApi
-from inductiva.client.model.task_status_code import TaskStatusCode
-from inductiva.types import Path
-from inductiva.utils.data import unpack_output
+from inductiva.client import models
 
 
-def get_task_status(task_id: str):
-    """Get the status of a simulation."""
-
-    api_config = api.validate_api_key(inductiva.api_key)
-
-    with ApiClient(api_config) as client:
-        api_instance = TasksApi(client)
-
-        status = api.get_task_status(api_instance, task_id)
-    logging.info("Task status: %s", status)
-
-    return status
-
-
-def fetch_task_output(task_id: str,
-                      output_dir: Optional[Path] = None,
-                      return_type: Type = pathlib.Path) -> Type:
-    """Fetch the results of a simulation."""
-    logging.info("Fetching the simulation output for task %s", task_id)
-    api_config = api.validate_api_key(inductiva.api_key)
-
-    with ApiClient(api_config) as client:
-        api_instance = TasksApi(client)
-
-        status = api.get_task_status(api_instance, task_id)
-
-        logging.info(status)
-        if status == "started":
-            logging.info("The task is still being executed.")
-            return_type = None
-        elif status == "submitted":
-            logging.info("The task is waiting for resources...")
-            return_type = None
-        else:
-            result_list = api.download_output(api_instance, task_id, output_dir)
-
-            return unpack_output(result_list, output_dir, return_type)
-
-
-def get_task_info(task_id) -> Dict:
-    """Get info about a task by its ID."""
-    api_config = api.validate_api_key(inductiva.api_key)
-
-    with ApiClient(api_config) as client:
-        api_instance = TasksApi(client)
-
-        try:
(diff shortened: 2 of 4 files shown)


inductiva/__init__.py
```diff
@@ -18,6 +18,7 @@
 from . import utils
 from . import stellarators
 from . import world
+from inductiva.tasks import Task
```
IvanPombo, line 21: What is the reason to have this import here?

inductiva/tasks/methods.py
```diff
@@ -89,14 +27,104 @@ def get_tasks_info(status: Optional[str] = None,
...
 
         if status is not None:
-            query_params["status"] = TaskStatusCode(status)
+            query_params["status"] = models.TaskStatusCode(status)
 
         try:
             # Get User Tasks
-            api_response = api_instance.get_user_tasks(
-                query_params=query_params,)
+            resp = api_instance.get_user_tasks(
```
IvanPombo, line 34: With the description you provide below, why not `response_dict`?

Moreover, I don't really see a reason for this change, since it looked cleaner before and the output seems to be the same. Can you elaborate?

inductiva/tasks/methods.py
```diff
@@ -89,14 +27,104 @@ def get_tasks_info(status: Optional[str] = None,
...
+
+    id_header = "ID"
+    simulator_header = "Simulator"
+    status_header = "Status"
+    submitted_header = "Submitted"
+    started_header = "Started"
+    duration_header = "Duration"
+    vm_type_header = "VM Type"
+
+    print(f"  {id_header:<20} {simulator_header:<12} {status_header:<12}"
```
IvanPombo, line 86: This seems very hard-coded and error prone... Shall, we use pandas (we are installing with inductiva anyway)?

IvanPombo (approved):
Seems wonderful! Only left minor comments.

inductiva/__init__.py
luispcunha: I removed it now

inductiva/tasks/methods.py
luispcunha: not response_dict because resp is the urllib3 response object, not a dict. the reason for the change was to get rid of the annoying DynamicSchemas that the generated client uses when serializing. where not using the objects that the client generates so I'll rather just use the raw json data. at some point we'll probably serialize this to our own dataclass

IvanPombo: Alright, seems good! @david451lima look into this for your work, since that was also our concern at a certain point.

inductiva/tasks/methods.py
luispcunha: yup i'll change the way the table is currently being constructed, but  using pandas just to format a table seems a bit overkill

inductiva/tasks/methods.py
```diff
@@ -62,12 +123,12 @@ def list(num_tasks,
...
-        1691080520213617518  openfoam     success      03 Aug, 16:35:21     03 Aug, 16:35:21     0h 1m 23s    n2-standard-32
+                         ID       Simulator               Status            Submitted              Started        Duration            VM Type
+        1691150776862178362        openfoam              success     04 Aug, 12:06:17     04 Aug, 12:06:18       0h 1m 53s      c2-standard-8
+        1691149904961476240        openfoam              success     04 Aug, 11:51:46     04 Aug, 11:51:46       0h 1m 28s      c2-standard-8
+        1691081881158823776        openfoam              success     03 Aug, 16:58:02     03 Aug, 16:58:02       0h 1m 20s     n2-standard-32
+        1691081409916619414        openfoam              success     03 Aug, 16:50:11     03 Aug, 16:50:11       0h 1m 20s     n2-standard-32
+        1691080520213617518        openfoam              success     03 Aug, 16:35:21     03 Aug, 16:35:21       0h 1m 23s     n2-standard-32
 
     Args:
         num_tasks: The number of tasks to list.
```
hpenedones, line 134: The comment should say (if I understand correctly that these are the "the most recent num tasks" (btw, in terms of what, launch time?) that respect the status filter... Right?

luispcunha: If filtering criteria is provided, this will get the most recent N tasks that match that criteria. I'll change the docstring to clarify that

IvanPombo (approved):

david451lima (commented):
LGTM. A question: now if I want to query for the status of a specific task, how do I do it since get_task_info was deprecated?

luispcunha:
> LGTM. A question: now if I want to query for the status of a specific task, how do I do it since get_task_info was deprecated?

Use the Task object. If you have it, call `task.get_status()`, if you don't have it but have the task ID, create it with `task = Task("id....")`

david451lima (approved):
LGTM

PR #504, opened 2023-08-30
[Coastal area scenario] Add method to convert bathymetry to a uniform grid

fabiocruz:
This PR adds a method `to_uniform_grid` to the bathymetry class that allows converting it to a uniform grid.

This is another important step to support loading arbitrary bathymetry files and performing simulations on them. For context, see #478.

To illustrate the behaviour of this method, let me use the initial random bathymetry defined below:

```python
import numpy as np
import inductiva

np.random.seed(123)

size = 400
x = np.random.uniform(0, 800, size=(size,))
y = np.random.uniform(0, 1000, size=(size,))
depths=np.random.uniform(-10, 10, size=(size,))

bathymetry = inductiva.fluids.scenarios.coastal_area.Bathymetry(depths, x, y)
```

Plotting this bathymetry, we get:

bathymetry_random_without_edges

To attempt a conversion of this bathymetry to a uniform grid, we call:
```python
bathymetry.to_uniform_grid()
```

This yields an error:
```text
ValueError: The bathymetry cannot be converted to a uniform grid because depths are not defined in one or more edge regions of the domain.
```

This happens because, since the bathymetry is generated at random locations $(x, y)$, some corners do not have edges, so we can't perform interpolation there.

Here is an example for which interpolation works. We start by making sure the edges of the domain are covered by locations $(x, y)$ in the previous bathymetry.

```python
np.random.seed(123)

x = np.random.uniform(0, 800, size=(size,))
y = np.random.uniform(0, 1000, size=(size,))
depths=np.random.uniform(-10, 10, size=(size,))

# Make sure the edges exist.
x[0], y[0] = 0, 0
x[1], y[1] = 0, 1000
x[2], y[2] = 800, 0
x[3], y[3] = 800, 1000
 
bathymetry = inductiva.fluids.scenarios.coastal_area.Bathymetry(depths, x, y)
```

This yields the following bathymetry:

bathymetry_random_with_edges

It can be converted to a grid:

```python
bathymetry_on_grid = bathymetry.to_uniform_grid()
```

This yields the expected bathymetry interpolated to a uniform grid:

bathymetry_random_grid


inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -264,35 +265,40 @@ def plot(
               the minimum or maximum colors, respectively.
             path: Path to save the plot. If `None`, the plot is not saved, and
               the matplotlib `Axes` object is returned instead.
-            grid_size: Size of the uniform grid to which the bathymetry is
-              interpolated before being plotted.
-            max_distance: Threshold distance to filter out points on the
+            grid_resolution: Resolution, in meters, of the uniform grid to which
```
IvanPombo, line 268: I am having issues understanding how this resolution relates with the two coordinates x,y.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -264,35 +265,40 @@ def plot(
...
-            (self.x, self.y),
-            self.depths,
-        )
+        logging.info(
+            "Plotting the bathymetry on a uniform grid...\n"
+            "- grid resolution %f m \n"
+            "- grid size %d x %d", grid_resolution, x_num, y_num)
 
-        depths_grid = interpolator(x_grid, y_grid)
+        if x_num > 1000 or y_num > 1000:
```
IvanPombo, line 284: If it is 1000x10 is it still large?

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -329,3 +335,43 @@ def plot(
...
+            y_num=y_size,
+        )
+
+        depths_grid = inductiva.utils.interpolation.interpolate_to_uniform_grid(
+            x=(self.x, self.y),
+            values=self.depths,
+            x_grid=(x_grid, y_grid),
+        )
+
+        if np.sum(np.isnan(depths_grid)) > 0:
```
IvanPombo, line 369: Could it occur that there are holes in the middle of the bathymetry?

IvanPombo (approved):
Seems good to merge! I only left a couple of comments.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: The uniform grid in which we plot the bathymetry has a resolution, i.e. a given grid spacing. This sets that resolution.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: Not really. I'll update the check to `if x_num * y_num > 1000**2`.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: Yes, it could!

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: We could, yes. But I thought that setting a single resolution for $x$ and $y$ in this plotting method would suffice. Do you think it is critical to have variable resolution in $x$ and $y$ here?

inductiva/fluids/scenarios/coastal_area/bathymetry.py
IvanPombo: Then, the message below needs to reflect that :)

fabiocruz: Pardon. It couldn't! There can be holes in the arbitrary $x$ and $y$ coordinate, but we can always interpolate in regions like such in the middle of the bathymetry. The issue is with interpolating at the edges, where we don't have values on both sides of the target locations to interpolate from.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
IvanPombo: It would make it more uniform with the function .to_uniform_grid(). Wdyt?

IvanPombo (approved):

inductiva/fluids/scenarios/coastal_area/bathymetry.py
IvanPombo: Ah ok! Then, it is fine 👍

PR #505, opened 2023-08-30
Fix default values workflow in tasks module

IvanPombo:
This PR fixes two issues that were occurring in the tasks module:

- The attribute self._default_files_list was not being initialised to None, hence when using self.get_output() from a task initiliased with just an id, it would fail!
- The resource_pool_id that serves to identify the machines in run_simulation.py was not being correct when machine_group was not provided.


luispcunha (approved):
lgtm

PR #507, opened 2023-08-30
Add temp paths to current directory

IvanPombo:
This PR adds a path to where the temporary directories of the commands and scenario input files are prepared to send to the backend. The reason for this change is concerned with the Windows framework, where temporary files are created on:
`C:\Users\username\AppData\Local\Temp` and this folder may have weird access permissions for us to read and write files.

Hence, to fix this issue and simplify this approach, we write the files in the current working directory and they are removed automatically with the context manager when passed to the backend.

### DIFF
--- a/inductiva/fluids/scenarios/wind_terrain/wind_terrain.py
+++ b/inductiva/fluids/scenarios/wind_terrain/wind_terrain.py
@@ -137,7 +137,8 @@ class WindOverTerrain(scenarios.Scenario):
                                               OPENFOAM_TEMPLATE_INPUT_DIR,
                                               COMMANDS_TEMPLATE_FILE_NAME)
 
-        with tempfile.NamedTemporaryFile() as commands_file:
+        commands_path = inductiva.utils.files.resolve_path()
+        with tempfile.NamedTemporaryFile(dir=commands_path) as commands_file:
             inductiva.utils.templates.replace_params_in_template(
                 template_path=commands_template_path,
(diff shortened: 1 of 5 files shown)


fabiocruz:
Have you tried setting `tempfile.tempdir` when running the code on Windows? See [this](https://docs.python.org/3/library/tempfile.html#tempfile.tempdir).

This would be a better option IMO, because it wouldn't require changing the behaviour in all systems just because of the behaviour on one system.

IvanPombo:
> Have you tried setting `tempfile.tempdir` when running the code on Windows? See [this](https://docs.python.org/3/library/tempfile.html#tempfile.tempdir).
> 
> This would be a better option IMO, because it wouldn't require changing the behaviour in all systems just because of the behaviour on one system.

Do you mean, me as a user set tempfile.tempdir where I want it to be?

The goal of this change, is to generalize for all systems, and not keep the usability of the API restricted to Unix.

fabiocruz:
> > Have you tried setting `tempfile.tempdir` when running the code on Windows? See [this](https://docs.python.org/3/library/tempfile.html#tempfile.tempdir).
> > This would be a better option IMO, because it wouldn't require changing the behaviour in all systems just because of the behaviour on one system.
> 
> Do you mean, me as a user set tempfile.tempdir where I want it to be?
> 
> The goal of this change, is to generalize for all systems, and not keep the usability of the API restricted to Unix.

Nop, I mean set this on the base `__init__.py` of the package.

PR #509, opened 2023-08-30
Fix some scenarions not running on Windows: replace NamedTemporaryFile with StringIO

luispcunha:
This fixes the problem described in #507. The problem however wasn't related to permissions to access the filesystem but actually, it was related to differences in behavior of the `NamedTemporaryFile` class in Windows vs. Unix platforms. Quoted from [here](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile):

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows). 

My approach to fix this rather than directly using the file-like object provided when creating a `NamedTemporaryFile` was to use `StringIO` to create an "in memory file" to be used when using the templates to create the commands file.

### DIFF
--- a/inductiva/scenarios.py
+++ b/inductiva/scenarios.py
@@ -1,8 +1,9 @@
 """Base class for scenarios."""
 
 from abc import ABC, abstractmethod
+import io
 import tempfile
-from typing import Optional
+from typing import Optional, Union
 
 from inductiva import resources
 from inductiva.types import Path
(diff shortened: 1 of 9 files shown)


inductiva/scenarios.py
```diff
@@ -19,11 +20,15 @@ def create_input_files(self, simulator: Simulator, input_dir: Path):
         """To be implemented in subclasses."""
         pass
 
-    def read_commands_from_file(self, commands_path: str):
+    def read_commands_from_file(self, commands_file: Union[str, io.StringIO]):
         "Read list of commands from commands.json file"
-        with open(commands_path, "r", encoding="utf-8") as f:
-            commands = json.load(f)
-        return commands
+        if isinstance(commands_file, str):
```
IvanPombo, line 25: This if could just read the file and return be just in the end.

IvanPombo (approved):
Lgtm! Thanks for solving this @luispcunha!

inductiva/scenarios.py
luispcunha: But since this is using the context manager to open/close the file, it wouldn't look that well because afterwards we would need to write the conditional logic again to know that if the file was opened here, we need to close it

fabiocruz (approved):
LGTM. Thanks @luispcunha!

PR #510, opened 2023-08-30
[Coastal area scenario] Update method to write bathymetry to file format required by SWASH

fabiocruz:
This PR introduces a method to write a bathymetry to a bot file, the format required by SWASH. For context, see #478.

It converts the bathymetry to a uniform grid, and writes that grid to the file.

Here is an example snippet:

```python
import inductiva

bathymetry = inductiva.fluids.scenarios.coastal_area.Bathymetry.from_random_depths(
    x_range=(0, 1000),
    y_range=(0, 1000),
    x_num=250,
    y_num=250,
    max_depth=30,
    initial_roughness=0,
)

# To write the bathymetry to a bot file...
# bathymetry.to_bot_file(path)

# Create the scenario.
scenario = inductiva.fluids.scenarios.CoastalArea(bathymetry=bathymetry, ...)

# Run simulation.
# This calls .to_bot_file() internally and generates a bathymetry file compatible with SWASH.
task = scenario.simulate()
```

Running this with the random bathymetry above, we get the following simulation output. I am only representing a short time as a means of illustrating it works.

https://github.com/inductiva/inductiva/assets/11545632/ccd82ef7-8674-4040-b628-c9e11bd41bcd

### DIFF
--- a/inductiva/fluids/scenarios/coastal_area/bathymetry.py
+++ b/inductiva/fluids/scenarios/coastal_area/bathymetry.py
@@ -50,12 +50,12 @@ class Bathymetry:
         x_range: Sequence[float],
         y_range: Sequence[float],
     ):
-        """Creates a `Bathymetry` object from a text file.
+        """Creates a `Bathymetry` object from a bot file.
         
-        The depth values are read from a text file. The text file must contain
-        a 2D array with the depths, in meters. The first and second dimensions
-        of the array in the text file (i.e. rows and columns) correspond to the
-        x and y directions, respectively.
+        The depth values are read from a bot file, i.e. a text file with a 2D
+        array with the depths, in meters. The first and second dimensions of the
+        array in the text file (i.e. rows and columns) correspond to the x and y
+        directions, respectively.
 
         Args:
             text_file_path: Path to the text file.
@@ -180,14 +180,27 @@ class Bathymetry:
 
         return cls(depths.flatten(), x.flatten(), y.flatten())
 
-    def to_text_file(self, text_file_path: str):
-        """Writes the bathymetry to a text file.
+    def to_bot_file(self, bot_file_path: str):
+        """Writes the bathymetry to a bot file.
 
         Args:
             text_file_path: Path to the text file.
         """
 
-        np.savetxt(text_file_path, self.depths)
+        x_grid, y_grid = inductiva.utils.grids.get_meshgrid(
+            x_range=self.x_range,
+            y_range=self.y_range,
+            x_num=len(self.x_uniques()),
+            y_num=len(self.y_uniques()),
+        )
+
+        depths_grid = inductiva.utils.interpolation.interpolate_to_uniform_grid(
+            x=(self.x, self.y),
+            values=self.depths,
+            x_grid=(x_grid, y_grid),
+        )
+
+        np.savetxt(bot_file_path, depths_grid)
 
     @property
     def x_range(self) -> Tuple[float]:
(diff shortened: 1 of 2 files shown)


inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -50,12 +50,12 @@ def from_bot_file(
...
+        """Creates a `Bathymetry` object from a bot file.
         
-        The depth values are read from a text file. The text file must contain
-        a 2D array with the depths, in meters. The first and second dimensions
-        of the array in the text file (i.e. rows and columns) correspond to the
-        x and y directions, respectively.
+        The depth values are read from a bot file, i.e. a text file with a 2D
+        array with the depths, in meters. The first and second dimensions of the
+        array in the text file (i.e. rows and columns) correspond to the x and y
+        directions, respectively.
```
cristianacarpinteiro, line 58: why call this a bot file? It's the common name in this field?

cristianacarpinteiro (approved):
Looks great @fabiocruz!

inductiva/fluids/scenarios/coastal_area/coastal_area.py
```diff
@@ -162,10 +162,8 @@ def _(self, simulator: SWASH, input_dir):  # pylint: disable=unused-argument
     bathymetry_x_num = len(self.bathymetry.x_uniques())
     bathymetry_y_num = len(self.bathymetry.y_uniques())
 
-    bathymetry_x_delta = (self.bathymetry.x_range[1] -
-                          self.bathymetry.x_range[0]) / bathymetry_x_num
-    bathymetry_y_delta = (self.bathymetry.y_range[1] -
-                          self.bathymetry.y_range[0]) / bathymetry_y_num
+    bathymetry_x_delta = (self.bathymetry.x_ptp()) / bathymetry_x_num
```
IvanPombo, line 165: What is this x_ptp?

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -180,14 +180,40 @@ def from_random_depths(
...
+            y_num=len(self.y_uniques()),
+        )
+
+        depths_grid = inductiva.utils.interpolation.interpolate_to_uniform_grid(
+            x=(self.x, self.y),
+            values=self.depths,
+            x_grid=(x_grid, y_grid),
+        )
+
+        if np.sum(np.isnan(depths_grid)) > 0:
```
IvanPombo, line 210: To clarify, users will still need to be aware of this?

The idea is that they clip before where they want or to just interpolate those values as you explained before. Is this the workflow you are thinking?

IvanPombo (approved):
LGTM!

inductiva/fluids/scenarios/coastal_area/coastal_area.py
fabiocruz: It gets the difference between maximum and minimum values of $x$, in this case. It follows the nomenclature of NumPy (see [this](https://numpy.org/doc/stable/reference/generated/numpy.ptp.html)), for which ptp refers to "peak-to-peak" value.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: Yep. If they attempt to use an invalid bathymetry, they are informed. They should crop the bathymetry (with a method that I plan to add later today) until it is valid for simulations.

PR #511, opened 2023-08-31
Stellarators objectives weights

joaobernardo23:
This PR includes another argument in the `simulate` method of the `StellaratorCoils` class. This argument is a dictionary containing the weights of the objective functions to be used in the construction of the total objective for the optimization process. This data is dumped into a json file that will then be used by the API.

### DIFF
--- a/inductiva/stellarators/scenarios/coils/coils.py
+++ b/inductiva/stellarators/scenarios/coils/coils.py
@@ -1,4 +1,5 @@
 """Class for the coils creation."""
+import json
 import math
 import os
 import random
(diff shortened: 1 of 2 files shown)


inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -280,7 +297,31 @@ def simulate(
...
+
+        # Default weights for the objectives.
+        if objectives_weights is None:
+            self.objectives_weights = {
+                'squared_flux': 1,
+                'coils_length': 2e-03,
+                'mean_squared_curvature': 3e-04,
+                'arclength_variation': 5e-03,
+                'curvature': 3e-04
+            }
```
fabiocruz, line 310: Store this in a global constant.

inductiva/stellarators/scenarios/coils/coils.py
```diff
@@ -280,7 +297,31 @@ def simulate(
...
+
+        else:
```
fabiocruz, line 312: You don't need this else block. Just let the validation run on any `objectives_weights` provided by you or the user.

fabiocruz (approved):
LGTM. My comments address only minor issues with the code.

PR #513, opened 2023-08-31
Add streamlines features

IvanPombo:
This PR adds two small features that provide more generality to the Streamlines methods on the SteadyStateOutput.

### DIFF
--- a/inductiva/fluids/post_processing/openfoam.py
+++ b/inductiva/fluids/post_processing/openfoam.py
@@ -93,7 +93,7 @@ class SteadyStateOutput:
                         n_points: int = 100,
                         initial_step_length: float = 1,
                         source_radius: float = 0.7,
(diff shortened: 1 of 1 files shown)


inductiva/fluids/post_processing/openfoam.py
```diff
@@ -93,6 +93,7 @@ def get_streamlines(self,
                         n_points: int = 100,
                         initial_step_length: float = 1,
                         source_radius: float = 0.7,
+                        source_center: tuple = (0, 0, 0),
```
luispcunha, line 96: does this allow only 3d tuples? if so, you can type annotate tuple[float, float, float], this way you're telling that the tuple needs 3 elements

luispcunha (approved):
lgtm. left only a minor suggestion

PR #516, opened 2023-08-31
[Coastal area scenario] Adds flexibility to grid interpolation method

fabiocruz:
This PR adds flexibility to the grid interpolation method of the bathymetry class, which allows for other class methods to use it. For context on why we need an interpolation method in this class, see #478.

With this PR, the expected user workflow is now possible:

1 - Read the bathymetry

```python
import inductiva

bathymetry = inductiva.fluids.scenarios.coastal_area.Bathymetry.from_ascii_xyz_file(file_path)
bathymetry.plot()
```

This may read a large, complex bathymetry, for instance:

bathymetry_demo_imshow

The user then chooses a cropping region and crops the bathymetry.

```python
bathymetry = bathymetry.crop(x_range=(51000, 52000), y_range=(12150, 13000))
bathymetry.plot()
```

The result is the expected crop of the bathymetry.
bathymetry_demo_cropped

Note that cropping doesn't do any interpolation, so the bathymetry may end up with patches where it is undefined. See for instance the lower left corner of the image above.

To be used in SWASH simulations, a bathymetry must be converted to a uniform grid. To do so, the user may call:

```python
bathymetry = bathymetry.to_uniform_grid(x_resolution=5,
                                        y_resolution=5,
                                        fill_value="nearest")
bathymetry.plot()
```

This allows setting the grid resolution and the strategy used to fill grid values for which no linear interpolation is possible. For the example snippet, above one would get something similar to the following image.

bathymetry_demo_cropped_uniform

Note that the whole domain is now colored.

The final step is simulating a scenario with this bathymetry.

```python
scenario = inductiva.fluids.scenarios.CoastalArea(bathymetry=bathymetry,
                                                  wave_source_location="S",
                                                  wave_amplitude=5,
                                                  wave_period=5.5)

task = scenario.simulate(simulation_time=100)
```

The result is, as expected:

https://github.com/inductiva/inductiva/assets/11545632/1ac063fa-6963-4714-b6b1-4c1fd825a72e

NB: if the user doesn't convert the bathymetry to a uniform grid, the coastal area scenario attempts to do so by calling `bathymetry.to_uniform_grid()`. However, it does so without specifying a strategy to fill nan regions, since the validity of such operation depends strongly on the bathymetry in use.

### DIFF
--- a/inductiva/fluids/scenarios/coastal_area/bathymetry.py
+++ b/inductiva/fluids/scenarios/coastal_area/bathymetry.py
@@ -334,33 +334,20 @@ class Bathymetry:
               defined.
         """
 
-        # Determine grid size based on ranges and resolution.
-        x_size = int(self.x_ptp() / x_resolution)
-        y_size = int(self.y_ptp() / y_resolution)
+        logging.info("Plotting the bathymetry on a uniform grid...")
+        depths_grid, _ = self._interpolate_to_uniform_grid(
+            x_resolution,
+            y_resolution,
+            threshold_distance=threshold_distance,
+            nullable=True,
+        )
 
-        logging.info(
-            "Plotting the bathymetry on a uniform grid...\n"
-            "- grid resolution %f x %f (m x m) m \n"
-            "- grid size %d x %d", x_resolution, y_resolution, x_size, y_size)
+        x_size = self.x_ptp() / x_resolution
+        y_size = self.y_ptp() / y_resolution
 
         if x_size > 1000 and y_size > 1000:
             logging.warning(
                 "The plotting grid is large. It may take a while to plot.")
-
-        # Create uniform grid for interpolation.
-        (x_grid, y_grid) = inductiva.utils.grids.get_meshgrid(
-            x_range=self.x_range,
-            y_range=self.y_range,
-            x_num=x_size,
-            y_num=y_size,
-        )
-
-        depths_grid = inductiva.utils.interpolation.interpolate_to_uniform_grid(
-            x=(self.x, self.y),
-            values=self.depths,
-            x_grid=(x_grid, y_grid),
-            threshold_distance=threshold_distance,
-        )
 
         # Plot the bathymetry.
         extent = (
--- a/inductiva/fluids/scenarios/coastal_area/bathymetry.py
+++ b/inductiva/fluids/scenarios/coastal_area/bathymetry.py
@@ -389,16 +389,26 @@ class Bathymetry:
         self,
         x_resolution: float = 2,
         y_resolution: float = 2,
+        fill_value: Optional[Union[float, str]] = None,
     ):
         """Converts the bathymetry to a uniform grid.
+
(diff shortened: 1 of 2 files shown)


inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -421,16 +418,23 @@ def _interpolate_to_uniform_grid(
         self,
         x_resolution: float,
         y_resolution: float,
+        threshold_distance: Optional[float] = None,
```
IvanPombo, line 421: This whole function seems to be agnostic of the bathymetry, should it also leave in `inductiva.utils.interpolation`?

inductiva/fluids/scenarios/coastal_area/bathymetry.py
```diff
@@ -454,16 +458,33 @@ def _interpolate_to_uniform_grid(
             y_num=y_size,
         )
 
-        depths_grid = inductiva.utils.interpolation.interpolate_to_uniform_grid(
+        depths_grid = inductiva.utils.interpolation.interpolate_to_coordinates(
             x=(self.x, self.y),
             values=self.depths,
-            x_grid=(x_grid, y_grid),
-        )
+            x_interpolation=(x_grid, y_grid),
```
IvanPombo, line 464: Why `x_interpolation` and `x` above?

inductiva/utils/interpolation.py
```diff
@@ -1,51 +1,61 @@
...
 import scipy
 
 
-def interpolate_to_uniform_grid(x: Tuple[np.ndarray],
-                                values: np.ndarray,
-                                x_grid: Tuple[np.ndarray],
-                                threshold_distance=None):
-    """Interpolate values to a uniform grid.
+def interpolate_to_coordinates(
+    x: Tuple[np.ndarray],
```
IvanPombo, line 10: This is a very weird name, since in other cases we use x,y. What do you think of the more explicity `coordinates` and `coordinates_interpolation`?

IvanPombo (approved):
LGTM! Only left a comment regarding the naming convention. 
Looking forward to see a tutorial with this @fabiocruz.

inductiva/fluids/scenarios/coastal_area/bathymetry.py
fabiocruz: Hm, maybe! It has behaviour specific to the bathymetry (for now), like the fact that it can't (in some situations) be interpolated. We can move it in the future, when similar behaviour is required by some other scenario / utility. WDTY?

IvanPombo: > Hm, maybe! It has behaviour specific to the bathymetry (for now), like the fact that it can't (in some situations) be interpolated. We can move it in the future, when similar behaviour is required by some other scenario / utility. WDTY?

Those situations are not specific of the bathymetry, are more related with the non-existence of points, no?
That is perfectly fine!

PR #517, opened 2023-08-31
Add functions to retrieve resources from backend

begaiym-k:
This PR adds new functionalities to the resources, such as listing active machine groups, getting storage size, etc. 

The list function returns the names of the active machines and not the IDs. I decided to go with the names, because using the ID which is a UUID object brings a lot of complications because it has to be in UUID format all the time, so when it's needed in a string format (which happens most of the times), it always needs to be converted before proceeding. So I sticked with the name, and used it almost everywhere. 

Since the name needs to be passed when recreating the MachineGroup object, the parameter 'name' has to be in the init, but then that might be misleading for the user, because in the initial set up they might set a name (thinking that's what they need to do) but it is only necessary when recreating the object. So I am not sure what is the best way to go with in this case.


inductiva/resources/machines.py
```diff
@@ -17,73 +19,67 @@ def __init__(
...
+        try:
+            logging.info("Creating a machine group."
+                         "This may take up to a few minutes.")
+            start_time = time.time()
+            instance_group = self._api.create_instance_group(
+                body=instance_group_config)
+            creation_time_mins = (time.time() - start_time) / 60
+
+            self.id = instance_group.body["id"]
+            self.estimated_price = self.estimate_price()
```
IvanPombo, line 67: Is this working already? Can you verify?

Should we still pass it to the user at this moment?

inductiva/resources/machines.py
```diff
@@ -17,73 +19,67 @@ def __init__(
...
-            Check https://cloud.google.com/compute/docs/machine-resource for 
+            Check https://cloud.google.com/compute/docs/machine-resource for
             more information about machine types.
             num_machines: The number of virtual machines to launch.
             spot: Whether to use spot machines.
             disk_size_gb: The size of the disk in GB, recommended min. is 20 GB.
             zone: The zone where the machines will be launched.
         """
         self.id = None
+        self.name = name
```
IvanPombo, line 36: If the user passes it, then it needs to be used to create the instance...
Otherwise, it is unintuitive. What we could do is call it label/tag and be a way to identify the MachineGroup but not its ID.

Problems that may surge are related with the unique ID. Since, if they create two machine groups with the same tag, we need to tell them they have two and they need to select which one they want.

inductiva/resources/resources.py
```diff
@@ -0,0 +1,62 @@
+"""Functions to manage or retrieve user resources."""
+from absl import logging
+import inductiva
+from inductiva.client.apis.tags import instance_api
+from inductiva import resources
+
+
+def get_storage_size():
+    """Returns the total storage size in GB."""
```
IvanPombo, line 9: Does this mean the total storage size I have available or that I consumed?

inductiva/resources/resources.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        api = instance_api.InstanceApi(inductiva.api.get_client())
+        instance_price = api.get_storage_size()
+        return instance_price.body
+    except inductiva.client.ApiException as api_exception:
+        raise api_exception
+
+
+def list():
+    """Lists all active machine group names.
```
IvanPombo, line 19: tags?

inductiva/resources/machines.py
```diff
@@ -17,73 +19,67 @@ def __init__(
         spot: bool = False,
         disk_size_gb: int = 20,
         zone: typing.Optional[str] = "europe-west1-b",
+        name: typing.Optional[str] = None,
```
IvanPombo, line 22: Is the idea to use the name to fetch a machine?

After launching a machine, going about your life, returning and wanting to pick up that machine again just by the name, none of the functions `start` or `terminate` will work properly just with `machine_group=resources.MachineGroup(name="api-something")`.

So I am really not understanding the point of setting the name in the argument. You need the `get_machine_group` for this to make sense. Which is exactly what you have in the resources.py.

So it's not clear why should it be passed here.

IvanPombo (changes requested):
Left a couple of comments. My concern is obviously related with the name, since it will be very unintuitive to use the name on the creation of a machine, but then it won't be used.

My suggestion is to be a tag instead. In this way, users can find their machineGroups by this tag, but they know that this is not the actual name of the machineGroup.

inductiva/resources/machines.py
```diff
@@ -92,37 +88,37 @@ def terminate(self):
...
-                "on_demand"] * self.num_machines
-        else:
-            estimated_price = instance_price.body[
-                "preemptible"] * self.num_machines
+        Otherwise returns None"""
+        response = self._api.get_group_status({"name": self.name})
 
-        return estimated_price
+        if response is None:
+            logging.info("Machine group does not exist.")
```
fabiocruz, line 120: Maybe be explicit that this is using the name to fetch the machine group and log `f"Machine group '{self.name}' does not exist."`

inductiva/resources/resources.py
```diff
@@ -0,0 +1,62 @@
...
+    """Returns the total storage size in GB."""
+    try:
+        api = instance_api.InstanceApi(inductiva.api.get_client())
+        instance_price = api.get_storage_size()
+        return instance_price.body
+    except inductiva.client.ApiException as api_exception:
+        raise api_exception
+
+
+def list():
```
fabiocruz, line 18: `list` is a name too short and general, why not `list_machine_groups`?

inductiva/resources/resources.py
```diff
@@ -0,0 +1,62 @@
...
+            logging.info("No active machine groups found.")
+            return response.body
+
+        machine_group_names = []
+        for machine_group in response.body:
+            machine_group_names.append(machine_group["name"])
+            logging.info("Name: %s", machine_group["name"])
+            logging.info("Nubmer of machines: %s",
+                         machine_group["num_instances"])
+            logging.info("Created at: %s", machine_group["created_at"])
```
fabiocruz, line 37: Should all of these be printed in the same line, to be easy to identify the number of machine groups in the logs and the info (number of machines and creation time) associated with each one?

inductiva/resources/resources.py
```diff
@@ -0,0 +1,62 @@
...
+        response = api.list_active_user_instance_groups()
+        if len(response.body) == 0:
+            logging.info("No active machine groups found.")
+            return response.body
+
+        machine_group_names = []
+        for machine_group in response.body:
+            machine_group_names.append(machine_group["name"])
+            logging.info("Name: %s", machine_group["name"])
+            logging.info("Nubmer of machines: %s",
```
fabiocruz, line 35: Typo, "number".

inductiva/resources/resources.py
```diff
@@ -0,0 +1,62 @@
...
+
+    except inductiva.client.ApiException as api_exception:
+        raise api_exception
+
+
+def get_machine_group(name: str):
+    """Returns a 'MachineGroup' object.
+
+    Given the name of the machine group, returns a 'MachineGroup' object
+    with the attributes of that machine group."""
```
fabiocruz, line 48: Are we making sure that machine group names are unique (for all users or per user)?

fabiocruz (approved):
LGTM.

I see there is some discussion about "labeling" machine groups by ID or name. As a user, I think IDs are easier, since I don't have to "name" one more entity (if I don't want to).

inductiva/resources/resources.py
begaiym-k: Consumed. I will rephrase it.

inductiva/resources/machines.py
begaiym-k: That's exactly why I already added that function [here](https://github.com/inductiva/inductiva/blob/cd30d2b741913cbe58302e1122891ec3ceb8b709/inductiva/resources/resources.py#L44) it retrieves the machine by name and recreates it.

luispcunha: Why is the name optional? I was under the impression that the names were created in the backend by us. If it's optional here, can the user choose a name that we will actually use?

inductiva/resources/machines.py
luispcunha: `f"Machine group does not exist: {self.name}"` is better as it avoids ambiguities if the name of the machine is something that looks like it could be part of the sentence and is more easily automatically processed. Check in the [styleguide](https://google.github.io/styleguide/pyguide.html#3102-error-messages)

inductiva/resources/resources.py
luispcunha: this is how it was decided in a previous meeting. the goal is to call it with `inductiva.resources.list()`.

inductiva/resources/resources.py
luispcunha: i think at some point the concepts of ID, name, the user facing name vs. the name we actually use in gcloud got mixed up

inductiva/resources/machines.py
IvanPombo: Yes, but you can use that function and initialise a MachineGroup without setting the name through the init. If you want to pass the name to the client just do it afterwards. I will explain my self better in that part of the code.

inductiva/resources/resources.py
IvanPombo: It is not mixed up. The ID is a UUID and the name is `api-ID`. So they are technically the same, but to launch machines with google api you need a name that starts with letters, so this was the way.

inductiva/resources/resources.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        api = instance_api.InstanceApi(inductiva.api.get_client())
+        response = api.get_instance_group({"name": name})
+
+        return resources.MachineGroup(
```
IvanPombo, line 53: I suggest doing:

```
machine_group = resources.MachineGroup(
       machine_type=response.body["machine_type"],
       num_machines=response.body["num_instances"],
       spot=response.body["spot"],
       disk_size_gb=response.body["disk_size_gb"],
       zone=response.body["zone"])

machine_group.name = name 
```

In this way, we remove the complexity of the user passing a name that will not be used, when generating the MachineGroup at the start. However, if we want to still provide that ability to allow for an easier way to retrieve machines we could use a label/tag.

inductiva/resources/machines.py
begaiym-k: I removed it now, I found another way to integrate it later when we retrieve it from the backend. So it's not there anymore. And about the labels, yes I think it is very important, we will add them eventually, but I think it's another PR, because it needs to be implemented in the backend as well.

inductiva/resources/resources.py
begaiym-k: But also with the name list, pylint raises an error, so maybe it's better to rename it to `list_machine_groups`?

inductiva/resources/resources.py
begaiym-k: Yeah, I tried that but it looked messy that way. I will show it in a demo meeting, then we can decide maybe.

inductiva/resources/resources.py
begaiym-k: Done!

inductiva/resources/storage.py
```diff
@@ -0,0 +1,16 @@
+"""Methods to interact with the user storage resources."""
+from absl import logging
+import inductiva
+from inductiva.client.apis.tags import instance_api
+
+
+def get_used():
```
IvanPombo, line 7: get_used_storage_disk()?

IvanPombo (approved):
LGTM!

inductiva/resources/storage.py
begaiym-k: Since it is already under `inductiva.resources.storage` I don't think there is need to mention storage again in the function name. Disk is a little misleading. Maybe get_space_used?

IvanPombo: Yes, you are correct. I thought you wanted to do: `inductiva.resources.get_storage_space_used()`.

PR #518, opened 2023-08-31
[WindTunnel] Add visuals for backend post-process files

IvanPombo:
This PR adds the capability to visualize more easily the files that were post-processed by default on the backend.

In this way, it simplifies the user workflow and keeps the exact same structure to read and plot the outputs.
The only change is that one now passes the `post_process=True` argument in the simulate to use the appropriate `task.get_output()`. That is, the `post_process` arg just controls how the outputs will be processed on the Client. If True, only the default files are download and can be visualized by default - some args are ignored. If False, it downloads everything and it provides full post-processing capabilities on the client.

In the future, this arg can be used to turn on or off the post-process on the backend.


IvanPombo:
After our discussion, the last commit updates the framework so that no argument for post-processing is passed on the simulate. This decision is left single-handedly for the task.get_output().

E.g. 
```python
scenario = inductiva.fluids.scenarios.WindTunnel(
            flow_velocity=[30, 0, 0],
            domain={"x": [-7, 12], "y": [-5, 5], "z": [0, 8]})

task = scenario.simulate(
        machine_group=machine_group,
        num_iterations=50,
        object_path="meta_object.obj",
        resolution="low",
        n_cores=8)
```

For downloading full simulation outputs and running rendering tools:
```python
output=task.get_output()

output.get_object_pressure_field().render()
output.get_streamlines().render_frame()
```

For downloading only default outputs and running rendering tools:
```python
output=task.get_output(full_output=False)

output.get_object_pressure_field().render()
output.get_streamlines().render_frame()
```

Wdyt, @fabiocruz @hpenedones @cristianacarpinteiro ?

Ps. I am carrying some technical debt here, since I am introducing code duplication to define the default files.

inductiva/fluids/post_processing/openfoam.py
```diff
@@ -37,31 +39,56 @@ def __init__(self, sim_output_path: types.Path):
...
+
+        sim_output_files = glob.glob(os.path.join(self.sim_output_path, "**"),
+                                     recursive=True)
+
+        if sim_output_files == default_output_files_list:
+            return False
+
+        return True
 
     def get_output_mesh(self):  # pylint: disable=unused-argument
```
cristianacarpinteiro, line 70: can't you erase now the pylint warning? you are using the self argument inside the method.

inductiva/fluids/post_processing/openfoam.py
```diff
@@ -77,16 +104,23 @@ def get_object_pressure_field(self, save_path: types.Path = None):
             and to render it.
         """
 
-        _, object_mesh = self.get_output_mesh()
+        if self.full_output:
```
cristianacarpinteiro, line 107: maybe call this variable other thing like` backend_pressure_field`??

inductiva/fluids/post_processing/openfoam.py
IvanPombo: Yes!

inductiva/fluids/post_processing/openfoam.py
IvanPombo: No, this is a general parameter for all outputs and not a specific one just for pressure_field.

cristianacarpinteiro:
> After our discussion, the last commit updates the framework so that no argument for post-processing is passed on the simulate. This decision is left single-handedly for the task.get_output().
> 
> E.g.
> 
> ```python
> scenario = inductiva.fluids.scenarios.WindTunnel(
>             flow_velocity=[30, 0, 0],
>             domain={"x": [-7, 12], "y": [-5, 5], "z": [0, 8]})
> 
> task = scenario.simulate(
>         machine_group=machine_group,
>         num_iterations=50,
>         object_path="meta_object.obj",
>         resolution="low",
>         n_cores=8)
> ```
> 
> For downloading full simulation outputs and running rendering tools:
> 
> ```python
> output=task.get_output()
> 
> output.get_object_pressure_field().render()
> output.get_streamlines().render_frame()
> ```
> 
> For downloading only default outputs and running rendering tools:
> 
> ```python
> output=task.get_output(full_output=False)
> 
> output.get_object_pressure_field().render()
> output.get_streamlines().render_frame()
> ```
> 
> Wdyt, @fabiocruz @hpenedones @cristianacarpinteiro ?
> 
> Ps. I am carrying some technical debt here, since I am introducing code duplication to define the default files.

I think this makes sense! I am not sure about the name `full_output`... Maybe `download_post_processing` ?

inductiva/fluids/scenarios/wind_tunnel/output.py
```diff
@@ -36,26 +36,35 @@ def get_force_coefficients(self, save_path: types.Path = None):
...
-                    force_coefficients.append(line.split()[1:])
-                # Add the force coefficients for the simulation time chosen
-                elif index == num_header_lines + self.last_iteration + 1:
-                    force_coefficients.append(line.split())
-
-        if save_path:
-            with open(save_path, "w", encoding="utf-8") as csv_file:
-                csv_writer = csv.writer(csv_file)
-                csv_writer.writerows(force_coefficients)
+        if not self.full_output:
```
fabiocruz, line 39: Is `full_output` an attribute of `WindTunnelOutput`?

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
```diff
@@ -94,10 +94,9 @@ def set_default_output_files(self):
         """Set default files generated by backend post-processing."""
 
         return [
-            "pressure_field.vtk", "streamlines.vtk", "vertical_slice_0.vtk",
-            "vertical_slice_1.vtk", "vertical_slice_2.vtk",
-            "force_coefficients.csv", "stderr.txt", "stdout.txt",
-            "horizontal_slice.vtk"
+            "pressure_field.vtk", "streamlines.vtk", "stdout.txt", "stderr.txt",
+            "force_coefficients.csv", "xy_slice.vtk", "xz_flow_slice.vtk",
+            "yz_flow_slice.vtk", "constant/triSurface/object.obj"
```
fabiocruz, line 99: This function seems to *get* the default output files and not *set* them. Rename it?

fabiocruz: Also, it should be a dispatch method, since this will be different for different simulators.

inductiva/tasks/task.py
```diff
@@ -192,13 +192,16 @@ def set_default_output_files(self,
 
     def get_output(
         self,
+        full_output: bool = True,
```
fabiocruz, line 195: I'd rename this to simply `full`. The `_output` suffix is redundant.

inductiva/tasks/task.py
```diff
@@ -215,8 +218,14 @@ def get_output(
             100%|██████████| 1.64G/1.64G [00:32<00:00, 55.1MiB/s]
         """
         self.wait()
+
+        if full_output:
+            filenames = []
+        else:
+            filenames = self._default_files_list
```
fabiocruz, line 225: `filenames = self._default_files_list if full else []`

fabiocruz: Actually, shouldn't `filenames` be `None` when you want to download all files? Passing an empty list sounds like you want to download *none of the files*.

fabiocruz (commented):
Got a few comments. See below.

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
IvanPombo: I will keep this as a set_default... since it actually sets the default output files you want to download later on. It does not fetch them from anywhere, so get also does not make sense in terms of files.

IvanPombo: Moreover, I think it can be default for all simulators, since we are the ones deciding the names of the default files and what they look like. The only one here that does not abide to that is the object.obj, but later on can also be simplified.

IvanPombo:
To agree with the argument the approach is as follows:

- If one wants to just the default files: `output = task.get_output()`;
- If ones wants to download all of the output files: `output = task.get_output(full=True)`.

Then, the outputs can be processed and visualized in the same way with:
```python
output.get_object_pressure_field().render()
output.get_streamlines().render_frame()
```

@hpenedones wdyt?

hpenedones:
Seems reasonable... But can you illustrate here how the user would see the visualization that was generated on the server side?

IvanPombo:
> Seems reasonable... But can you illustrate here how the user would see the visualization that was generated on the server side?
> 
> Would the user call some method, or just browse the filesystem and open the video file that was downloaded?

It is exactly what I mention above. On the backend, we just do the post-processing of the general output files, since here we are just rendering images (steady state simulations).
```python
pressure_field = output.get_object_pressure_field()
pressure_field.render_frame()
```

```python
streamlines = output.get_streamlines()
streamlines.render_frame()
```

inductiva/fluids/scenarios/wind_tunnel/wind_tunnel.py
IvanPombo: However, I am not computing or deriving something from the object... Should I just set it as an attribute then?

fabiocruz: I think it should be a method (because you need it to be dispatched differently based on the simulator in use, for instance).

You are *getting* something unique to the object, which I think justifies naming the method as `get_...`.

IvanPombo: Alrighty then!

IvanPombo:
After finishing the documentation and fixing the issues here, I decided to keep with the current implementation with just a single get function for both the default files and post-processing tools. In the near future, this will take a different shape and I still feel comfortable with this current implementation.

inductiva/fluids/post_processing/openfoam.py
```diff
@@ -37,31 +39,67 @@ def __init__(self, sim_output_path: types.Path):
...
-            last_iteration: Last iteration of the simulation.
-                Obtained through the output folders.
+            default_output_files_list: List of default output files
+                that are computed on the backend and saved to the
+                simulation output directory.
         """
 
         self.sim_output_path = sim_output_path
+        self.full_output = self.inspect_output()
+        print(self.full_output)
```
fabiocruz, line 49: Remove the print statement.

inductiva/fluids/post_processing/openfoam.py
```diff
@@ -46,28 +46,44 @@ def __init__(self, sim_output_path: types.Path):
...
         self.full_output = self.inspect_output()
+        print(self.full_output)
 
     def inspect_output(self):
-        """Inspect the output of the simulation."""
+        """Inspect the output of the simulation.
+        
+        Return True if the simulation output contains more files than
+        the default ones. This is used to determine if the post-processing
+        was computed on the backend or not."""
```
fabiocruz, line 56: If the function returns this, then it should be named `is_full()` or something similar. `inspect_output` suggests that you want to print the output file names, plot them quickly or something of the sort.

inductiva/fluids/post_processing/openfoam.py
```diff
@@ -363,12 +374,12 @@ def __init__(self, mesh_data, scalar_name: str = None):
         self.mesh.point_data[scalar_name] = mesh_data.point_data[scalar_name]
         self.mesh.cell_data[scalar_name] = mesh_data.cell_data[scalar_name]
 
-    def render(self,
-               off_screen: bool = False,
-               background_color: str = "black",
-               scalars_cmap: str = "viridis",
-               virtual_display: bool = False,
-               save_path: types.Path = None):
+    def render_frame(self,
```
fabiocruz, line 377: Why renaming this? How is the frame to be rendered selected?

inductiva/tasks/task.py
```diff
@@ -192,15 +192,15 @@ def set_default_output_files(self,
 
     def get_output(
         self,
-        full_output: bool = True,
+        full: bool = False,
```
fabiocruz, line 195: I feel that the default here should be `True`, since most tasks won't be associated with a scenario that controls which output files should be downloaded by default.

fabiocruz (approved):
LGTM. Got minor comments and questions only this time.

inductiva/fluids/post_processing/openfoam.py
IvanPombo: I can set render for all. There are no time frames involved.

inductiva/tasks/task.py
IvanPombo: With default to true, the name needs to be changed, otherwise it does not make sense to call "full=False".

Issue #520, opened 2023-09-01
[Coastal area scenario] Add breakwater to a bathymetry


fabiocruz:
One of our API users wants to be able to add a breakwater structure to a bathymetry, more specifically vertical breakwaters (think of a parallelepiped going from the sea bottom to some elevation above the sea level).

Below I propose an interface to do so, for which I'd appreciate to receive some feedback before implementing it.

## Breakwater class

I suggest adding a `Breakwater` class, which would be initialized as follows:

```python
breakwater = inductiva.fluids.scenarios.coastal_area.Breakwater(
    x_center=..., # The x coordinate of the center of the breakwater.
    y_center=...,
    x_length=..., # The length of the breakwater along the x axis.
    y_length=...,
    rotation_angle=..., # The angle to rotate the breakwater relative to the positive direction of the x axis.
    crest_elevation=..., # How above the sea level the breakwater extends.
)
```

NB: the parameters describing the breakwater have been discussed with the user, and seem to provide a sufficient description of these structures.

## Bathymetries with breakwaters

I suggest supporting adding breakwaters to bathymetries.

When initializing a bathymetry from the class constructor, one could pass a list of breakwaters:
```python
bathymetry = inductiva.fluids.scenarios.coastal_area.Bathymetry(
    x=...,
    y=...,
    depths=...,
    breakwaters=[breakwater,],
)
```

When initializing a bathymetry from the available classmethods to read data from files (e.g. ASCII XYZ files), I don't think it makes much sense to support adding bathymetries on initialization.

I suggest instead that we have a method to add breakwaters to bathymetries:

```python
bathymetry = inductiva.fluids.scenarios.coastal_area.Bathymetry.from_ascii_xyz_file(...)
bathymetry.add_breakwaters([breakwater,] # -> None
```

When a bathymetry has a breakwater (or more than one), the depth in the region of the breakwater is overridden by the crest elevation of the breakwater.

Allowing adding breakwaters to bathymetries in this manner allows users to e.g. load arbitrary bathymetry files, add breakwater structures to them and export new bathymetry files to be used as they please (e.g. to run simulations).

IvanPombo:
This seems great @fabiocruz!!

hpenedones:
@sarmento and I were chatting about the right interface for this scenario, and he suggested having something like:

`bathymetry = inductiva.coastal.scenarios.Bathymetry(...)`

(instead of fluids, and  the longer class names.)
[...]

PR #523, opened 2023-09-04
[Coastal dynamics] Move files related to coastal dynamics simulations

fabiocruz:
This PR moves files in the package such that coastal dynamics utilities are accessible at a higher level than before, namely the scenario, simulator and other helper classes.

One can now access these classes via the `coastal` submodule, e.g.:

```python
import inductiva

bathymetry = inductiva.coastal.Bathymetry(...)
scenario = inductiva.coastal.CoastalArea(bathymetry=bathymetry, ...)
```

For further context, see #520.


IvanPombo (commented):
Overall the code is correct and I approve it! 
However, this change raises a question: the DualSPHysics is a simulator that can be used for coastal dynamics and other of fluids simulations. Another example, for the future will be FEniCS, that it is not just specific for structural mechanics. Even OpenFOAM has some capabilities for structural mechanics. Only Gromacs and SplishSplash are kinda correct at the moment.

So this change only thinks about SWASH and XBeach, and won't make much sense with the current nature of the package. An idea is to move all the simulators to a folder that is not specific of the nature (fluids, molecules, etc...). Hence, `fluids`, `molecules`,..., would be just for the scenarios. 

Wdyt @hpenedones @fabiocruz @sarmento ?

fabiocruz:
I agree, we need to move them out of use-case-specific directories. Finding a way to group them is a challenge, though. We could group them per "simulation method", e.g. FVM, FEM, SPH, for instance. Or at least those that are usable in more than one use-case and/or field.

IvanPombo:
But we could also just do it in an overall folder of simulators. 

My idea would be to have
'inductiva.simulators.openfoam()' and not 'inductiva.simulators.fvm.openfoam()'. 

Wdyt?

fabiocruz:
I think it should be fine for now to keep them all under one directory. However, it's hard to imagine that scaling well to hundreds of simulators... We can always find ways to group them when it becomes an issue, so I'm ok with merging them all into one directory for now.

IvanPombo:
> I think it should be fine for now to keep them all under one directory. However, it's hard to imagine that scaling well to hundreds of simulators... We can always find ways to group them when it becomes an issue, so I'm ok with merging them all into one directory for now.

Let's move with the one directory approach for now, since it is the one making the most sense and thereafter we check when hundreds of simulations are added.

fabiocruz:
> > I think it should be fine for now to keep them all under one directory. However, it's hard to imagine that scaling well to hundreds of simulators... We can always find ways to group them when it becomes an issue, so I'm ok with merging them all into one directory for now.
> 
> Let's move with the one directory approach for now, since it is the one making the most sense and thereafter we check when hundreds of simulations are added.

Sure, I can do that in a separate PR.

PR #524, opened 2023-09-05
Workflow to publish to PyPI

luispcunha:
This PR adds a workflow to publish the package to PyPI when a tag is created in GitHub. Note the caveat about the package version still needing to be defined in the `setup.cfg` file explained in the top of the workflow file.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,10 +1,10 @@
 # Workflow that publishes package to PyPI.
 # The workflow is triggered on push of a tag that starts with "v", e.g. v0.3.0.
-# The worflow will build the package and then publish it to PyPI. Note that the
+# The workflow will build the package and then publish it to PyPI. Note that the
 # version specified in the tag is not the used for the package version on PyPI.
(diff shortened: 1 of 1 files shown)


fabiocruz (approved):
LGTM! Great little add-on that simplifies deployment to PyPI, thanks @luispcunha!

hpenedones:
Awesome, @luispcunha 
Is the mechanisms to add tags, the one described here?
https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository

luispcunha:
> Awesome, @luispcunha Is the mechanisms to add tags, the one described here? https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository

We can either do it like that (creating a new tag when creating the release), or we can also create it locally via the command line and then push it (`git tag vx.y.z`, `git push --tags`). Currently the workflow is triggered on the tag, not on creating a release. But since a release is always associated to a tag, I think this is fine, so that we can push minor releases easier without going through the whole process of creating a "formal" release here

PR #530, opened 2023-09-05
[Protein Solvation] Allow user to only download the compressed trajectory in ProteinSolvation

david451lima:
Instead of downloading the full trajectory, the user can choose to download only the compressed trajectory file. This requires the name of the compressed trajectory file to be set as the default output file at simulate() time. The user controls if only wants the compressed trajectory setting compressed_trajectory to True

### DIFF
--- a/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
@@ -44,7 +44,7 @@ class ProteinSolvation(Scenario):
         self.temperature = temperature
 
     def set_default_output_files(self):
-        """Set the downloadable files as the compressed trajectory adjacent.
+        """Set the default output files for the Protein Solvation Scenario.
 
         Instead of downloading the full trajectory, the user can choose
         to download only the compressed trajectory file. The following files
--- a/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
@@ -43,6 +43,16 @@ class ProteinSolvation(Scenario):
         self.protein_pdb = files.resolve_path(protein_pdb)
         self.temperature = temperature
 
+    def set_small_files(self):
+        """Set the downloadable files as the compressed trajectory.
+
+        Instead of downloading the full trajectory, the user can choose
+        to download only the compressed trajectory file. This requires
+        the name of the compressed trajectory file to be set as the
+        default output file at simulate() time."""
+
+        return ["compressed_trajectory.xtc"]
+
     def simulate(
             self,
             simulator: Simulator = GROMACS("proteinsolvation"),
@@ -51,7 +61,8 @@ class ProteinSolvation(Scenario):
             simulation_time_ns: float = 10,  # ns
             output_timestep_ps: float = 1,  # ps
             integrator: Literal["md", "sd", "bd"] = "md",
-            n_steps_min: int = 5000) -> tasks.Task:
+            n_steps_min: int = 5000,
+            compressed_trajectory: bool = False) -> tasks.Task:
         """Simulate the solvation of a protein.
 
         Args:
@@ -90,6 +101,9 @@ class ProteinSolvation(Scenario):
                                 run_async=run_async)
 
         task.set_output_class(ProteinSolvationOutput)
+
+        if compressed_trajectory is True:
+            task.set_default_output_files(self.set_small_files())
 
         return task
 
--- a/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
@@ -52,7 +52,8 @@ class ProteinSolvation(Scenario):
         the user does not require the full trajectory at task.get_output()."""
 
         return [
-            "protein.gro", "compressed_trajectory.xtc", "solvated_protein.tpr"
+            "protein.gro", "compressed_trajectory.xtc", "solvated_protein.tpr",
+            "topol.top"
         ]
 
     def simulate(
@@ -96,6 +97,7 @@ class ProteinSolvation(Scenario):
         self.integrator = integrator
         self.n_steps_min = n_steps_min
         commands = self.get_commands()
+
         task = super().simulate(simulator,
                                 machine_group=machine_group,
                                 commands=commands,
--- a/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
@@ -43,15 +43,17 @@ class ProteinSolvation(Scenario):
         self.protein_pdb = files.resolve_path(protein_pdb)
         self.temperature = temperature
 
-    def set_small_files(self):
-        """Set the downloadable files as the compressed trajectory.
+    def set_default_output_files(self):
+        """Set the downloadable files as the compressed trajectory adjacent.
 
         Instead of downloading the full trajectory, the user can choose
-        to download only the compressed trajectory file. This requires
-        the name of the compressed trajectory file to be set as the
-        default output file at simulate() time."""
+        to download only the compressed trajectory file. The following files
+        are set as the default ones, and will be the only ones downloaded if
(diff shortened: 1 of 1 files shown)


inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -43,6 +43,16 @@ def __init__(self, protein_pdb: str, temperature: float = 300):
         self.protein_pdb = files.resolve_path(protein_pdb)
         self.temperature = temperature
 
+    def set_small_files(self):
```
cristianacarpinteiro, line 46: Why having this additional function if we are just downloading one file? To construct the universe with MDAnalysis you also need the .gro file and the .tpr. Shouldn't you add those files too here?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -51,7 +61,8 @@ def simulate(
             simulation_time_ns: float = 10,  # ns
             output_timestep_ps: float = 1,  # ps
             integrator: Literal["md", "sd", "bd"] = "md",
-            n_steps_min: int = 5000) -> tasks.Task:
+            n_steps_min: int = 5000,
+            compressed_trajectory: bool = False) -> tasks.Task:
```
cristianacarpinteiro, line 65: Maybe I would give other name to the variable. Because this will always be calculated. Maybe the variable could be download_full_trajectory and be set to false by default.

cristianacarpinteiro (commented):
Thanks for this David! We have to be carefull about this because the calculate_rmsf function in the post-processing needs the full trajectory file, so maybe we can raise a error there.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
fabiocruz, line 46: Call this `get_default_output_files`, for consistency with the wind tunnel scenario. See #518.

fabiocruz (changes requested):
Thanks @david451lima!

I don't think that allowing the user to set which files to _download_ when they _simulate_ the scenario is a pattern we want to use. The specific files to be downloaded should only be controlled via the task objects. See my specific comment about this below.

sarmento:
> Thanks @david451lima!
> 
> I don't think that allowing the user to set which files to _download_ when they _simulate_ the scenario is a pattern we want to use. The specific files to be downloaded should only be controlled via the task objects. See my specific comment about this below.

I agree with Fabio. Downloading should be controlled via the output objects. I think this issue was already raised wrt to the WindTunnel scenario. From the point of view of the user, we want to be consistent, and I think the decisions about downlaoding should all be taken from the same place (output object). Also, I think we should avoid fitting too many arguments in the simulate method.

In general, we should try to avoid "downloads" as much as we can, and keep the files on the server side until there is no other option. Even visualization routines should be, as much as possible, run on the server side. That way, the client package will be as lightweight as possible and easy to install, and we will be able to make use of powerful and expensive routines in the server side.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
sarmento, line 65: I don't think this decision should be made here but I would like to leave a comment on the naming.
The name of this parameter, as it is, is non-informative and ambiguous. For this kined of parameters, try to follow the pattern

VERB_OBJECT

for example something like:

generate_compressed_trajectory

would be much easier to understand.

david451lima:
I completely agree with the fact that the specification of the downloadable files should be done at task.get_outputs() and not at simulation time. However, the way the Tasks class is structured requires this definition at simulation time. (@fabiocruz i don't know if this was addressed in the wind tunnel scenario. I can also be wrong and there can be a smart way to circunvent this that I'm not aware).

IvanPombo:
> Thanks for this David! We have to be carefull about this because the calculate_rmsf function in the post-processing needs the full trajectory file, so maybe we can raise a error there.

Why not also bring a RMSF already pre-computed from the backend? Is it hard to configure a default one?

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -91,6 +102,9 @@ def simulate(
 
         task.set_output_class(ProteinSolvationOutput)
 
+        if compressed_trajectory is True:
```
IvanPombo, line 105: Remove this if and just keep the assignement you are doing below.

The idea is now that the task will handle everything! Then if you do:
`task.get_output(all_files=True)` it will download all files if with False, it will just download the default file.

Now the main idea is that you need to manage appropriately in the ProteinSolvationOutput. In the WindTunnelOutput (more precisely, on the SteadyStateOutput) we are inspecting if the simulation files we download match or not with the default files (in order to avoid passing any further arguments). Then, the functions of the class execute the appropriate steps to use the correct files. 

Notice that, you may need other files for the rendering and you need to be aware of the other functions as well. At least trow some errors for users to understand things are not available.

hpenedones:
> Thanks for this David! We have to be carefull about this because the calculate_rmsf function in the post-processing needs the full trajectory file, so maybe we can raise a error there.

I think this is the kind of post-processing steps that could be on the server side, because the full trajectory will tend to be huge and we can't assume that the user will download it...

sarmento:
> > Thanks for this David! We have to be carefull about this because the calculate_rmsf function in the post-processing needs the full trajectory file, so maybe we can raise a error there.
> 
> I think this is the kind of post-processing steps that could be on the server side, because the full trajectory will tend to be huge and we can't assume that the user will download it...

Totally agree!

david451lima:
@everyone If you could re-review I would appreciate.
Right now the compressed trajectory files to be downloaded (the ones specific for the post processing tool we provide) are specified as default, the user does not need to specify any download arguments at simulation time.
The only thing that remains to be done is to pass the post processing to the back-end, but that is beyond the scope of this particular PR.

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -43,6 +43,19 @@ def __init__(self, protein_pdb: str, temperature: float = 300):
         self.protein_pdb = files.resolve_path(protein_pdb)
         self.temperature = temperature
 
+    def set_default_output_files(self):
+        """Set the downloadable files as the compressed trajectory adjacent.
```
IvanPombo, line 47: Set the default output files for the Protein Solvation Scenario.

IvanPombo (approved):

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
```diff
@@ -43,6 +43,19 @@ def __init__(self, protein_pdb: str, temperature: float = 300):
...
+
+        Instead of downloading the full trajectory, the user can choose
+        to download only the compressed trajectory file. The following files
+        are set as the default ones, and will be the only ones downloaded if
+        the user does not require the full trajectory at task.get_output()."""
+
+        return [
```
cristianacarpinteiro, line 54: Cool, with all these files the post-processing will work.

cristianacarpinteiro (approved):
LGTM!

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
fabiocruz: This method gets something, so it should be named `get_…`

fabiocruz (approved):

inductiva/molecules/scenarios/protein_solvation/protein_solvation.py
david451lima: This method doesnt get something, it sets the default values that will be downloaded in the task.get_outputs

hpenedones: I agree with @fabiocruz here, this method is not setting anything: it is just returning a list with filenames, therefore it should be called "get_...". Later you will use it to set something on the task object :
```python
task.set_default_output_files(self.get_default_output_files())
```
and that's fine, but that is the moment when you are "setting" something.
When you invoke this method you are just "getting" a list...

PR #536, opened 2023-09-07
Cleanup function to list machine groups

luispcunha:
Introduced changes:
- Change name of method to show machine group cost: now it's called `estimate_cloud_cost` to clarify that it doesn't represent a possible charge, but an estimation of the cost of the requested resources in google cloud.
- Change name of the method that lists active machine groups to reflect that it only displays active machine groups.
- Cleanup display of list of active machine groups:
```
INFO:absl:Active machine groups:
                                    Name         VM Type   # machines         Created at
api-c9054524-7d0c-41ef-a575-58b54ac30479   c2-standard-4            2   07 Sep, 05:56:51
api-c9054524-7d0c-41ef-a575-58b54ac30479   c2-standard-4            2   07 Sep, 05:56:51
```

### DIFF
--- a/inductiva/utils/format_utils.py
+++ b/inductiva/utils/format_utils.py
@@ -1,0 +1,60 @@
+"""Util functions for formatting data for printing to console."""
+import datetime
+import pandas as pd
+import numpy as np
+
+
+def bytes_formatter(n_bytes: int) -> str:
+    """Convert bytes to human readable string."""
+    res = float(n_bytes)
+
+    for unit in ["B", "KiB", "MiB", "GiB", "TiB"]:
+        if res < 1024:
+            if unit == "B":
+                return f"{res:.0f} {unit}"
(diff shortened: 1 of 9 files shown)


inductiva/utils/format_utils.py
```diff
@@ -0,0 +1,60 @@
...
+def get_tabular_str(
+    rows,
+    columns,
+    default_col_space=15,
+    override_col_space=None,
+    formatters=None,
+) -> str:
+    """Converts a list of lists to a string table.
+
+    Temporary solution to display tables. `pandas` dependency to print tables
```
hpenedones, line 41: Ok, if we manage to remove the pandas dependency, maybe then we can consider to use something more lightweight like:

https://pypi.org/project/tabulate/

hpenedones (approved):

PR #538, opened 2023-09-07
[Fluid tank scenario] Remove mesh utilities

fabiocruz:
This PR removes the mesh utilities associated with the fluid tank scenario.

Meshing is now done as a pre-processing step on the API side. To allow for this, the properties of the fluid tank are written to a JSON file that is passed to the API.


fabiocruz:
Note that **this does not fully remove the dependency from gmsh**, since it is also used by the `structures` submodule under development.

To circumvent this, I have removed `structures` from the main `__init__.py` file, which prevents the cascade of imports that causes an error when running `import inductiva` on a machine without the libraries required to run gmsh.

sarmento (approved):
I think this should be ok for now!
A couple of things for the future:

1) fix import
2) move 3D Shapes to an "outer" package so it can be used elsewhere (e.g. closer to Terrain)

IvanPombo (approved):
LGTM! I just get a bit confused with all of the dict structures, but I think it looks great!

hpenedones (commented):
Awesome!!!
Can we already remove the magic line from the requirements.txt file?
Or is it pending on things related to the structures sub package (I saw that you commented it out here, so it should be fine?)

fabiocruz:
> Awesome!!! Can we already remove the magic line from the requirements.txt file? Or is it pending on things related to the structures sub package (I saw that you commented it out here, so it should be fine?)

See my comment above, https://github.com/inductiva/inductiva/pull/538#issuecomment-1710361373.

fabiocruz:
@hpenedones, the `import inductiva` statement doesn't fail because `gmsh` is in the requirements. It fails because we `import gmsh` in some file under `inductiva.submodule.subsubmodule....`. Commenting the `structures` import statement from the main `__init__.py` plus the changes proposed here should be enough to make the package run without trouble.

hpenedones:
But we can also remove the gmsh from the requirements so that it doesn't trigger the installation of gmsh and its dependencies, no?

fabiocruz:
> But we can also remove the gmsh from the requirements so that it doesn't trigger the installation of gmsh and its dependencies, no?

Technically we still have code that depends on gmsh, associated with the structural mechanics simulator. However, since that submodule is still under development, it should be fine to remove gmsh from the requirements. The people developing that submodule (@Joao97ribeiro @baribeiro) should be aware that they need to install gmsh anyway to keep developing.

sarmento:
> > But we can also remove the gmsh from the requirements so that it doesn't trigger the installation of gmsh and its dependencies, no?
> 
> Technically we still have code that depends on gmsh, associated with the structural mechanics simulator. However, since that submodule is still under development, it should be fine to remove gmsh from the requirements. The people developing that submodule (@Joao97ribeiro @baribeiro) should be aware that they need to install gmsh anyway to keep developing.

OK! Once we get the new version out in pypi I will test it.
Also: we need to start having a plan for 3D object manipulation.

fabiocruz:
I removed `gmsh` and `meshio` from the requirements.

luispcunha:
> I removed `gmsh` and `meshio` from the requirements.

To remove them from the dependencies that pip installs, you should remove them from setup.cfg

hpenedones:
> > I removed `gmsh` and `meshio` from the requirements.
> 
> To remove them from the dependencies that pip installs, you should remove them from setup.cfg, otherwise they will still be installed. To avoid confusion later, we can change the requirements.txt to just install "-e ."

Oh, wasn't aware of this... this is not respecting the DRY principle (in a quite important place), because we have two files saying the "same thing" and it's quite error prone... what do we need to do to fix this?

luispcunha:
> > > I removed `gmsh` and `meshio` from the requirements.
> 
> > 
> 
> > To remove them from the dependencies that pip installs, you should remove them from setup.cfg, otherwise they will still be installed. To avoid confusion later, we can change the requirements.txt to just install "-e ."
> 
> 
> 
> Oh, wasn't aware of this... this is not respecting the DRY principle (in a quite important place), because we have two files saying the "same thing" and it's quite error prone... what do we need to do to fix this?

Change the requirements.txt to install "-e .", and leave the setup.cfg with the dependencies of the package.

hpenedones:
> 

Ok, great, do you mind doing a small Pull Request with that change? It's very important to remove barriers of entry in the earliest possible interactions with the package, e.g. installation time, or import time.

hpenedones (approved):

luispcunha:
> > 
> 
> Ok, great, do you mind doing a small Pull Request with that change? It's very important to remove barriers of entry in the earliest possible interactions with the package, e.g. installation time, or import time. And if developers believe that removing the entry from requirements.txt is enough, we will be under the illusion of fixing something, but the problem is still there... Thanks!!

#538, but this one should be merged first because the branch I'm using in that PR was created from this one, so it includes all of these changes.

PR #543, opened 2023-09-07
[Coastal area scenario] Fixes issues with random bathymetry generator

fabiocruz:
This PR fixes a couple of issues with the random bathymetry generator:
- it adds a random seed for reproducibility.
- it reverses the sign of the offset applied when determining the percentage of point above water (because above water means subtracting depth).


cristianacarpinteiro (approved):
LGTM

sarmento (approved):

PR #547, opened 2023-09-08
[ProteinSolvation] Removed RMSF computation from frontend

david451lima:
Removed frontend calculation of the RMSF, which removes the necessity of downloading the full trajectory file as default. The RMSF data is downloaded as a .npy file

### DIFF
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -67,8 +67,19 @@ class ProteinSolvationOutput:
         print(f"Number of trajectory frames: {len(self.universe.trajectory)}")
         return view
 
-    def plot_rmsf_per_residue(self, rmsf_values: np.array):
-        "Plot RMSF values per residue."
+    def plot_rmsf_per_residue(self):
+        """Plot the root mean square fluctuation (RMSF) over a trajectory.  
+
+        It is typically calculated for the alpha carbon atom of each residue. 
+        These atoms make the backbone of the protein.The RMSF is the square root 
+        of the variance of the fluctuation around the average position:
+        &rhoi = √⟨(xi - ⟨xi⟩)²⟩
+        It quantifies how much a structure diverges from the average structure 
+        over time, the RSMF can reveal which areas of the system are the most 
+        mobile. Check 
+        https://userguide.mdanalysis.org/stable/examples/analysis/alignment_and_rms/rmsf.html 
+        for more details."""
+
         rmsf_values = np.load(RMSF_VALUES_FILE)
         # Plot the data
         residue_number = np.arange(len(rmsf_values))
--- a/inductiva/molecules/scenarios/protein_solvation/post_processing.py
+++ b/inductiva/molecules/scenarios/protein_solvation/post_processing.py
@@ -11,10 +11,10 @@ import numpy as np
 
 import inductiva
 
-FULL_TRAJECTORY_FILE = "full_trajectory.trr"
 COMPRESSED_TRAJECTORY_FILE = "compressed_trajectory.xtc"
 PROTEIN_TOPOLOGY_FILE = "protein.gro"
 SYSTEM_TOPOLOGY_FILE = "solvated_protein.tpr"
+RMSF_VALUES_FILE = "rmsf_values.npy"
 
 
 class ProteinSolvationOutput:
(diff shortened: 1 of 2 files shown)


inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -67,45 +67,9 @@ def render_interactive(self,
...
-
-        It is typically calculated for the alpha carbon atom of each residue. 
-        These atoms make the backbone of the protein.The RMSF is the square root 
-        of the variance of the fluctuation around the average position:
-        &rhoi = √⟨(xi - ⟨xi⟩)²⟩
-        It quantifies how much a structure diverges from the average structure 
-        over time, the RSMF can reveal which areas of the system are the most 
-        mobile. Check 
-        https://userguide.mdanalysis.org/stable/examples/analysis/alignment_and_rms/rmsf.html 
-        for more details.
```
cristianacarpinteiro, line 81: put this info on the the doc from the plot_rmsf function.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -67,45 +67,9 @@ def render_interactive(self,
...
-
-        # Calculate RMSF for carbon alpha atoms
-        c_alphas = align_universe.select_atoms("protein and name CA")
-        rmsf = mda.analysis.rms.RMSF(c_alphas).run()
-        rmsf_values = rmsf.results.rmsf
-        duration = time.time() - start_time
-        print(f"RMSF calculation took {duration:.2f} seconds.")
-        return rmsf_values
-
     def plot_rmsf_per_residue(self, rmsf_values: np.array):
```
cristianacarpinteiro, line 70: Here return the rmsf_values, so that the user can do the render_residue_attributes.

cristianacarpinteiro (approved):

inductiva/molecules/scenarios/protein_solvation/post_processing.py
sarmento, line 70: Just by glossing over this, it seems weird for you to remove one method and not the other. Would you mind adding in the method documentation for method plot_rmsf_per_residue where the data is supposed to come from?

hpenedones:
Does this mean we can remove the import of

inductiva/molecules/scenarios/protein_solvation/post_processing.py
david451lima: If you check plot_rmsf.. method you see that we load the data previously computed in the client side from a .npy file

sarmento: OK! But please try to explain that flow somewhere more explicitly.

david451lima:
> Does this mean we can remove the import: import MDAnalysis as mda
> 
> and also remove the MDAnalysis package from the requirements.txt @cristianacarpinteiro @david451lima ? That would be simply awesome, because it would simplify installation, and it will get rid of that warning message that we are getting when importing inductiva, I believe...

Sadly not yet. The visualization is still stuck to the client side and still requires MDAnalysis. I'm sorry for not being able to migrate that to the backend. It seems that we will need to abandon nglview and try other visualziation packages to be able to perform this migration.

cristianacarpinteiro:
@hpenedones No we cant remove the MDanalysis dependency yet because we need it to construct the universe object and plot it with nglview.

hpenedones:
> @hpenedones No we cant remove the MDanalysis dependency yet because we need it to construct the universe object and plot it with nglview.

Hmmm, ok, but when I search the code of the ProteinSolvation scenario, the import related to mda is never used (probably you will even get a pylint warning, no)?

sarmento:
> @hpenedones No we cant remove the MDanalysis dependency yet because we need it to construct the universe object and plot it with nglview.

Should we simply remove this capability for now? I think we should consider not shipping this now.

hpenedones:
In other words: it's only used in the utils.py file. So you need to keep the dependency. But you can remove the import of mda on the scenario file...

hpenedones:
> 

The problem with that is that we would be "unshipping" a feature from the previous release, that was already able to do the client side rendering of the protein. So, I wouldn't remove it now. But I think we can start thinking of ways to simplify this and keep the functionality, while simplifying the dependencies.

hpenedones (approved):
Feel free to merge after fixing pylint warning about the import. Thanks

PR #548, opened 2023-09-08
Fix sizes of images in Protein Solvation documentation

cristianacarpinteiro:


sarmento (approved):
I am assuming that just scaling works well (as opposed to regenerating the images at the right resolution)

cristianacarpinteiro:
The render looked fine, but with more time I can do it already with the right dimensions.

hpenedones:
Thanks @cristianacarpinteiro, that looks way better!

PR #549, opened 2023-09-08
Remove repeated info from requirements.txt

luispcunha:
`setup.cfg` is where dependencies of the package that are installed when the user does `pip install inductiva` are specified. `requirements.txt` is just an utility to setup the local dev environment, it now installs the package in editable mode (`-e .`) and other dev dependencies.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,8 +28,6 @@ install_requires =
     scipy
     matplotlib
     xarray
-    gmsh
-    meshio
     vtk
     imageio
     pyvista
@@ -38,6 +36,7 @@ install_requires =
     nglview
     MDAnalysis
     utm
+    dask
 
 [options.package_data]
 # Include .obj and .jinja files:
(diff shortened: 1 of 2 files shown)


hpenedones:
Awesome, thanks @luispcunha : can you just check if your branch is synced with the latest merges... this PR shows a lot of unrelated files...

hpenedones (approved):

setup.cfg
```diff
@@ -38,6 +36,7 @@ install_requires =
     nglview
     MDAnalysis
     utm
+    dask
```
hpenedones, line 39: is it a new dependency?

luispcunha: it was specified in the requirements.txt (by @fabiocruz, [here](https://github.com/inductiva/inductiva/commit/1f958923cb81d9cc54310ce3e57d43e97fe80fc2)), so i added it to the `setup.cfg` because probably it was added there by mistake

luispcunha: according to @fabiocruz it is needed

PR #551, opened 2023-09-08
Consolidate main readme

luispcunha:
This PR serves to further consolidate the main README. However, there were somethings I wasn't able to be 100% sure how they should be done. I'll list them here:
- [] The Protein Solvation should be simplified in the main README. I wanted to move it to a separate page, but I noticed that there's already a separate page with Protein Solvation documentation. Which one is the most recent?
- [] Is there any scenario missing a link in the main README?


sarmento (approved):
You can merge this now.

PR #552, opened 2023-09-08
[ProteinSolvation] plot_rmsf returns rmsf_values

david451lima:
plot_rmsf_values now returns rmsf_values for the user to further work with


sarmento (approved):
LGTM

PR #556, opened 2023-09-08
Pip install --upgrade inductiva

hpenedones:
By default the pip install will not install the latest version of the package, if there was already an older version of inductiva installed.


hpenedones:
One of you please merge, as I need to leave now. thanks!

cristianacarpinteiro (approved):

PR #557, opened 2023-09-11
[WindTunnel] Address testing errors with Documentation

IvanPombo:
This PR addresses a few errors on the documentation. In particular, when would copy the code snippets there available they would not run due to inconsistencies between the args passed and the args the functions need.

Moreover, I added an example object that now can be downloaded.

Notice that the large number of changes is due to the addition of this obj file.

### DIFF
--- a/inductiva/fluids/scenarios/wind_tunnel/README.md
+++ b/inductiva/fluids/scenarios/wind_tunnel/README.md
@@ -8,21 +8,21 @@ To initialize the scenario, the user can define the following parameters:
 
 Now, the user is ready to simulate the steady state. Here, the user chooses the object to be inserted inside the wind tunnel. This object is defined with a geometry file in STL or OBJ format. Notice that, the required meshing step will automatically be made before starting the simulation. The meshing is done with the [snappyHexMesh tool](https://www.openfoam.com/documentation/user-guide/4-mesh-generation-and-conversion/4.4-mesh-generation-with-the-snappyhexmesh-utility) of OpenFOAM.
 
-To test this scenario we have available one example of a vehicle geometry in OBJ format - [download it here](../../../../resources/geometry/test_vehicle.obj). The user can also use his own geometry file, as long as it is in STL or OBJ format.
+To test this scenario we have available one example of a vehicle geometry in OBJ format - [download it here](/resources/geometry/test_vehicle.obj). The user can also use his own geometry file, as long as it is in STL or OBJ format.
 
 The simulation parameters available for the user to configure are:
 - `num_iterations`: Set the maximum number of iterations for the iterative algorithm to converge.
 - `resolution`: Controls the resolution of the meshing that is done prior to the simulation. The higher the resolution, the finer the meshing. Possibilities: "high", "medium", "low".
 
 Moreover, the hardware and interaction are configured with the usual general parameters - `machine_group`, `run_async`, `n_cores`.
-Launching a simulation returns a task object, which can be used to verify the status of the simulation, get the simulation outputs and access post-processing tools. See more in the [Tasks section](../../../../README.md).
+Launching a simulation returns a task object, which can be used to verify the status of the simulation, get the simulation outputs and access post-processing tools. See more in the [Tasks section](inductiva/tasks/README.md).
 
 
 
 
 ### Example:
 
-To run this example download a vehicle geometry in OBJ format from [here](../../../../resources/geometry/test_vehicle.obj) and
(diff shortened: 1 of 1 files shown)


inductiva/fluids/scenarios/wind_tunnel/README.md
```diff
@@ -8,27 +8,34 @@ To initialize the scenario, the user can define the following parameters:
...
 
 Moreover, the hardware and interaction are configured with the usual general parameters - `machine_group`, `run_async`, `n_cores`.
 Launching a simulation returns a task object, which can be used to verify the status of the simulation, get the simulation outputs and access post-processing tools. See more in the [Tasks section](../../../../README.md).
 
+
+
+
 ### Example:
 
+Download a vehicle geometry in OBJ format from [here](../../../../resources/geometry/test_vehicle.obj).
```
AugustoPeres, line 25: And the API key?

inductiva/fluids/scenarios/wind_tunnel/README.md
```diff
@@ -8,27 +8,37 @@ To initialize the scenario, the user can define the following parameters:
 
 Now, the user is ready to simulate the steady state. Here, the user chooses the object to be inserted inside the wind tunnel. This object is defined with a geometry file in STL or OBJ format. Notice that, the required meshing step will automatically be made before starting the simulation. The meshing is done with the [snappyHexMesh tool](https://www.openfoam.com/documentation/user-guide/4-mesh-generation-and-conversion/4.4-mesh-generation-with-the-snappyhexmesh-utility) of OpenFOAM.
 
+To test this scenario we have available one example of a vehicle geometry in OBJ format - [download it here](../../../../resources/geometry/test_vehicle.obj). The user can also use his own geometry file, as long as it is in STL or OBJ format.
+
```
david451lima, line 12: Is it possible to have the path of the OBJ as absolute? 
If we change the folders in the future this can break

sarmento: I think that is better. Also, we may have a folder under www.inductiva.ai where we add resources. WDTY @hpenedones.

Finally, for a future demo, consider also using a randomly created object instead? Can we do it?

inductiva/fluids/scenarios/wind_tunnel/README.md
```diff
@@ -8,27 +8,37 @@ To initialize the scenario, the user can define the following parameters:
...
+To run this example download a vehicle geometry in OBJ format from [here](../../../../resources/geometry/test_vehicle.obj) and
+do not forget to insert your API Key (get one by filling this [form](https://docs.google.com/forms/d/e/1FAIpQLSflytIIwzaBE_ZzoRloVm3uTo1OQCH6Cqhw3bhFVnC61s7Wmw/viewform?usp=sf_link)).
+
 ```python
 import inductiva
 
+inductiva.api_key = "YOUR_API_KEY"
+
 # Initialize the scenario
 scenario = inductiva.fluids.scenarios.WindTunnel(
```
sarmento, line 34: Is this still this long? (just asking...)

sarmento (approved):

inductiva/fluids/scenarios/wind_tunnel/README.md
IvanPombo: Yes! But I think we can remove scenarios for now and do `inductiva.fluids.WindTunnel()`, since the simulators are not anymore under fluids specifically.

inductiva/fluids/scenarios/wind_tunnel/README.md
hpenedones: > I think that is better. Also, we may have a folder under [www.inductiva.ai](http://www.inductiva.ai) where we add resources. WDTY @hpenedones.
> 
> Finally, for a future demo, consider also using a randomly created object instead? Can we do it?

I think that storing the example files here in the git repo is fine (unless they are too big, in that case we don't want the git repo to get too bloated). Having them hosted on our website would mean that we also have to update the website repo and redeploy it, when we make changes to the documentation.
But one thing we need to verify is that these links work when the README is showed in pypi.
https://pypi.org/project/inductiva/

It seems that we have a lot of broken links there when the user is directed to other scenarios, etc. @IvanPombo @luispcunha

IvanPombo: On it! I believe it is just need to address the main README. I will merge this PR and open a new one with those changes.

PR #560, opened 2023-09-11
Molecules post process optional deps

luispcunha:
Make MDAnalysis and nglview optional


setup.cfg
```diff
@@ -15,28 +15,31 @@ classifiers =
...
+  vtk
+  imageio
+  pyvista
+  imageio-ffmpeg
+  ipython
+  utm
+  dask
+
+[options.extras_require]
+molecules =
```
hpenedones, line 40: Maybe call it molecules_viz. Your idea of splitting the visualization extra requirements separate by subfield (molecules, fluids) might make sense, such that users from different fields don't need to install visualization libraries only used for other fields... let's see if it works well / it's not too much fragmentation. And then in any case we might move more things to the server side.

IvanPombo (approved):
Nice work @luispcunha! Just left a comment to warn the users about the next steps if they real want to use this.

inductiva/molecules/scenarios/utils.py
```diff
@@ -1,15 +1,27 @@
...
+    from MDAnalysis.analysis import align
+except ImportError:
+    mda = None
+    transformations = None
+    align = None
+
+
+def _check_mdanalysis() -> None:
+    if mda is None or transformations is None or align is None:
+        raise ImportError("MDAnalysis is not installed.")
```
hpenedones, line 14: We need to tell the user how to install it. Either we are assuming that by saying this, the user will be doing "pip install MDAnalysis" or if we want the user to do pip install inductiva[molecules_viz] then we need to tell it explicitly in this error message.

hpenedones (commented):
This is cool, we are removing dependencies from the main package! Left some small suggestions!

setup.cfg
IvanPombo: I think `molecules` is good enough, since it installs related with the molecules vertical and avoids long names. On the documentation, we can specify that `pip install inductiva` just installs the barebones for running simulations.

Furthermore, I am not completely sure about the fragmentation as well. There are some packages that are just specific for the FluidTank scenario for example. So the naming for fluids may be more treaky.

inductiva/molecules/scenarios/protein_solvation/post_processing.py
```diff
@@ -22,18 +26,25 @@ def __init__(self, sim_output_path: pathlib.Path = None):
...
 
         self.sim_output_dir = sim_output_path
+        # universe will be loaded only when needed
+        self.universe = None
+
+    def _load_universe(self):
         topology_path = os.path.join(self.sim_output_dir, PROTEIN_TOPOLOGY_FILE)
         trajectory_path = os.path.join(self.sim_output_dir,
                                        COMPRESSED_TRAJECTORY_FILE)
-        self.universe = inductiva.molecules.scenarios.utils.unwrap_trajectory(
```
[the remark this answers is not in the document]
sarmento: @david451lima and @cristianacarpinteiro : I have mentioned this before. There are a lot of these assignments where the function name does not map the name of the variable. Also, always remember that other people will also have to understand these names of veriables so you may want to add some extra context about what a "universe" is...

inductiva/molecules/scenarios/utils.py
sarmento: I think messages need to be as helpful as possible. We should be explicit.

setup.cfg
IvanPombo: FluidTank is using xarrays to deal with the output data of another project we had. At the time it was merged into the API and we were still tinkering with having frameworks to read the data into appropriate structures. Thereafter, xarray are used in combination with matplotlib to render things.

In the other, SPH scenarios we are using pyvista to hold data and render things.

sarmento: OK! We need to think deeper about this. @david451lima : you will also be helping in this front, so have a look at this dependency too later.

inductiva/utils/optional_deps.py
```diff
@@ -0,0 +1,66 @@
...
+    # pylint: enable=import-outside-toplevel
+
+
+molecules_missing_deps_msg = (
+    "You can install this and other missing dependencies for 'molecules'"
+    "with `pip install 'inductiva[molecules_extra]'`.")
+
+# Apply the two first arguments to _needs_optional_deps function, creating
+# a new decorator function with those arguments already set.
+needs_molecolules_extra_deps = functools.partial(
```
hpenedones, line 62: Typo: molecules

inductiva/utils/optional_deps.py
```diff
@@ -0,0 +1,66 @@
...
+            raise RuntimeError(msg) from e
+
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def _check_molecules_optional_deps():
+    # pylint: disable=import-outside-toplevel
+    import MDAnalysis as mda
```
IvanPombo, line 47: Can you explain why we need to import inside and then del? I am confused by this?

inductiva/molecules/scenarios/md_water_box/post_processing.py
```diff
@@ -1,9 +1,15 @@
 "Postprocessing steps for the MDWaterBox scenario."
 import os
-import nglview as nv
+try:
+    import nglview as nv
+except ImportError:
+    nv = None
```
sarmento, line 6: If this is a global variable, should you "name" it as so? glb_nv_available

inductiva/molecules/scenarios/utils.py
```diff
@@ -1,11 +1,19 @@
 """Tools to analyze molecular dynamics simulations."""
-import MDAnalysis as mda
-from MDAnalysis import transformations
-from MDAnalysis.analysis import align
+try:
+    import MDAnalysis as mda
+    from MDAnalysis import transformations
+    from MDAnalysis.analysis import align
+except ImportError:
+    mda = None
```
sarmento, line 7: globals?

setup.cfg
```diff
@@ -15,28 +15,31 @@ classifiers =
...
+  tqdm
+  jinja2
+  numpy
+  scipy
+  matplotlib
+  xarray
+  vtk
+  imageio
+  pyvista
+  imageio-ffmpeg
```
sarmento, line 34: isn't this a highly specific package that requires external libs?

sarmento (commented):
I am happy we are making progress on this front.
A few more questions/comments

inductiva/utils/optional_deps.py
luispcunha: i rewrote this in a different way. Can you check if you think it's understandable?

setup.cfg
luispcunha: Yes, but since it is related with molecules, it'll be done in a different PR

inductiva/molecules/scenarios/utils.py
luispcunha: these were the imports that were already there. I just wrapped them in try except to catch the ImportError

inductiva/utils/optional_deps.py
```diff
@@ -19,48 +16,48 @@ def load_from_np():
...
-    # unused
-    del mda
-    del nv
-    # pylint: enable=import-outside-toplevel
+    return wrapper
 
 
 molecules_missing_deps_msg = (
-    "You can install this and other missing dependencies for 'molecules'"
+    "You can install this and other missing dependencies for 'molecules' "
```
IvanPombo, line 54: Just say here: "Install the missing dependencies for `molecules` with `...`", just to simplify the english variability. e.g. this.

inductiva/utils/optional_deps.py
IvanPombo: Much more! Thanks!

IvanPombo (approved):
Excellent! Much more understandable!

PR #565, opened 2023-09-12
Remove visuals dependencies

IvanPombo:
This PR removes all visualizations and post-processing dependencies. With it, users that just do `pip install inductiva` will be able to run simulations, perform the task.get_output() to retrieve the results, but won't be able to use the visualization classes that we have. A message error appears referring how to install those packages with inductiva as well.

Some further considerations:
- Scipy, numpy, pandas are still in the default installation. 
- To remove pandas, we need to change a few functions.
- To pass scipy into the common_extra it is better to remove the core package - linalg stuff.
- The numpy is used for type annotations and they trigger first than the decorator, which implies this approach won't work. Due to it's generality, we decided to keep it.

### DIFF
--- a/inductiva/fluids/post_processing/splishsplash.py
+++ b/inductiva/fluids/post_processing/splishsplash.py
@@ -46,7 +46,7 @@ def convert_vtk_data_dir_to_netcdf(
         xr_dataset.to_netcdf(os.path.join(netcdf_data_dir, f"{file_stem}.nc"))
 
 
-def read_vtk_file_to_xr_dataset(file_path: str, time: float):
+def read_vtk_file_to_xr_dataset(file_path: str, time: float) -> "xr.Dataset":
     """Reads a single simulation output file to an xarray Dataset.
 
     Args:
--- a/inductiva/fluids/post_processing/splishsplash.py
+++ b/inductiva/fluids/post_processing/splishsplash.py
@@ -17,7 +17,6 @@ except ImportError:
     np = None
     _vtk_data_to_numpy = None
 
-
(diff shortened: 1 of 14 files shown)


inductiva/fluids/post_processing/splishsplash.py
```diff
@@ -40,7 +47,7 @@ def convert_vtk_data_dir_to_netcdf(
         xr_dataset.to_netcdf(os.path.join(netcdf_data_dir, f"{file_stem}.nc"))
 
 
-def read_vtk_file_to_xr_dataset(file_path: str, time: float) -> xr.Dataset:
+def read_vtk_file_to_xr_dataset(file_path: str, time: float):
```
hpenedones, line 49: I see that this required to remove some type annotations in a couple of places.... but it might be fair enough for now....

hpenedones (commented):
This is awesome, great effort @IvanPombo ! 
Looking forward to try this new lightweight version of the inductiva package!

inductiva/fluids/post_processing/splishsplash.py
IvanPombo: I thought it was fair for now, since we still need to uniformize the outputs of SPH simulations (either xarray or pyvista).

luispcunha: you can use forward references for those type annotations, *i.e.*, specify the type annotation as a string literal (`'xr.Dataset'`). More info [here](https://peps.python.org/pep-0484/#forward-references)

begaiym-k (approved):
LGTM! Thanks @IvanPombo for this work!

PR #566, opened 2023-09-12
Remove core module and pass scipy to extra deps

IvanPombo:
This PR removes the core module from Inductiva package. The goal is to clean the package and remove some dependencies that are not being used at all from the begining of our user interactions.

In this way, we can simplify the scipy dependency, remove it from the default installation, and reduce the amount of conflicts that can occur with the overall dependencies and python.


inductiva/utils/optional_deps.py
```diff
@@ -50,14 +50,33 @@ def wrapper(*args, **kwargs):
...
+
+needs_fluids_extra_deps = functools.partial(
+    _needs_optional_deps,
+    ["pyvista", "vtk", "scipy", "xarray", "dask", "imageio", "matplotlib"],
+    _missing_deps_msg("fluids"),
+)
+
+needs_coastal_extra_deps = functools.partial(
+    _needs_optional_deps,
+    ["utm", "imageio", "scipy", "matplotlib"],
```
hpenedones, line 74: imageio and matplotlib are already in the common extra, if I understand correctly your mechanism they are not needed here, right?

hpenedones: Or is it for the warning when the import fails?

IvanPombo: It is for the purpose of the warning, and mention that this will be the ones needed to install.

PR #567, opened 2023-09-12
Remove core module and scipy in default installation

IvanPombo:
This PR removes the core module from Inductiva package. The goal is to clean the package and remove some dependencies that are not being used at all from the begining of our user interactions.

In this way, we can simplify the scipy dependency, remove it from the default installation, and reduce the amount of conflicts that can occur with the overall dependencies and python.


hpenedones (commented):
Cool, thanks Ivan, keep testing locally with:
https://stackoverflow.com/questions/46205648/can-pip-install-from-setup-cfg-as-if-installing-from-a-requirements-file

luispcunha (approved):
thanks @IvanPombo

PR #569, opened 2023-09-13
Remove scenarios folder from modules

IvanPombo:
This PR removes the folder of scenarios from their respective modules (`fluids`, `molecules`, etc...). The reason for this change is due to moving the simulators to a single isolated folder, hence the structure `fluids/scenarios` becomes redundant, since there is no need to differentiate.

I will add every owner as reviewer to look into this change!


inductiva/fluids/__init__.py
```diff
@@ -9,17 +9,19 @@
...
-from .scenarios import DamBreak
-from .scenarios import FluidBlock
-from .scenarios import WindTunnel
-from .scenarios import (
+from .dam_break import DamBreak
+from .fluid_block import FluidBlock
+from .wind_tunnel import WindTunnel
+from .wind_terrain import WindOverTerrain
+from .fluid_tank import (
     FluidTank,
```
sarmento, line 17: This is a bit of a weird import.
Let's fix it later, in a subsequent PR.

inductiva/molecules/__init__.py
```diff
@@ -1,3 +1,4 @@
 #pylint: disable=missing-module-docstring
-from .scenarios import ProteinSolvation
-from .scenarios import MDWaterBox
+from .protein_solvation import ProteinSolvation, ProteinSolvationOutput
```
sarmento, line 2: This is also a weird import...
How would the code look like in this case?

inductiva/molecules/protein_solvation/README.md
```diff
@@ -39,7 +39,7 @@ To test this scenario we have available one example of a protein PDB file - [dow
 inductiva.api_key = "YOUR_API_KEY"
 
  # Initialize the scenario
- scenario = inductiva.molecules.scenarios.ProteinSolvation(
+ scenario = inductiva.molecules.ProteinSolvation(
```
sarmento, line 42: Is this how the code would look like with that previous import?

inductiva/stellarators/__init__.py
```diff
@@ -1,2 +1,2 @@
 #pylint: disable=missing-module-docstring
-from .scenarios import StellaratorCoils
+from .coils import StellaratorCoils
```
sarmento, line 2: Check with @hpenedones if this is where we want to have this.
I am not sure.

inductiva/fluids/__init__.py
IvanPombo: I agree... and I think it is more related with the general structure of imports that needs to be discussed.

inductiva/molecules/__init__.py
IvanPombo: The point of this is for us to be able to do:
`inductiva.molecules.ProteinSolvation` instead of `inductiva.molecules.protein_solvation.ProteinSolvation`.

I don't understand your point about weirdness.

inductiva/molecules/protein_solvation/README.md
IvanPombo: Yes! It fits the coastal area scenarios that were moved by Fábio recently.

inductiva/stellarators/__init__.py
IvanPombo: The general idea that I want to pass is that the modules `stellarators`, `fluids`, `molecules` are already specific to scenarios. So stating scenarios is redudant.

hpenedones: It can be in a separate PR if you prefer, but the package for stellarators should be called plasmas. For consistency with WindTunnel (which is a simulated machine, like Stellarators) is under fluids, not under a package called "windtunnel".

IvanPombo: I am gonna do it next :)

inductiva/fluids/__init__.py
sarmento: ok! Let's fix it later.

inductiva/molecules/__init__.py
sarmento: ok then!

inductiva/molecules/protein_solvation/README.md
sarmento: ok!

sarmento (approved):

PR #572, opened 2023-09-13
Fix possible issues when running scenarios with preprocessing

luispcunha:
I ran into an issue when running the FluidTank scenario.
The following:
```python
scenario = FluidTank()
scenario.simulate(simulator=SPlisHSPlasH())
```
It would not work because we need to call splishsplash with a different executer script due to needed additional preprocessing. This was being passed in the constructor of the default simulator, but if I passed my simulator as an argument it wouldn't work. This fixes that so that the scenario configures the simulator with scenario-specific things inside the simulate method.

### DIFF
--- a/inductiva/simulators/simulator.py
+++ b/inductiva/simulators/simulator.py
@@ -1,6 +1,6 @@
 """Base class for low-level simulators."""
 from typing import Optional
-from abc import ABC, abstractmethod
+from abc import ABC
 
 from inductiva import types, tasks, resources
 from inductiva.utils import files
@@ -9,10 +9,24 @@ from inductiva.utils import files
 class Simulator(ABC):
     """Base simulator class."""
 
-    @property
-    @abstractmethod
-    def api_method_name(self) -> str:
-        pass
+    def __init__(self):
+        self.api_method_name = ""
+
+    def override_api_method_prefix(self, prefix: str):
+        """Override the API method prefix.
+
+        Example:
+            # prefix = "protein_solvation"
+            "md.gromacs.run_simulation" becomes
+              "protein_solvation.gromacs.run_simulation"
+
+        Args:
+            prefix: The new prefix to use.
+        """
+        last_elements = self.api_method_name.split(".")[1:]
+        all_elements = [prefix] + last_elements
+
+        self.api_method_name = ".".join(all_elements)
 
     def _setup_input_dir(self, input_dir: types.Path):
         """Setup the simulator input directory."""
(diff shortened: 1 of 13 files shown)


inductiva/simulators/simulator.py
```diff
@@ -9,10 +9,24 @@
...
+        Example:
+            # prefix = "protein_solvation"
+            "md.gromacs.run_simulation" becomes
+              "protein_solvation.gromacs.run_simulation"
+
+        Args:
+            prefix: The new prefix to use.
+        """
+        last_elements = self.api_method_name.split(".")[1:]
+        all_elements = [prefix] + last_elements
```
sarmento, line 27: extend()?

sarmento (commented):
What is the minimal script that would test this?
Could you have a test script at least for one scenario?

IvanPombo (approved):
Great simplification! LGTM!

inductiva/simulators/simulator.py
sarmento: Just for readibilty but not suoer important

david451lima (changes requested):
I think that here we need to distinguish between the scenarios that only work with the default executer and the ones that work with both of them.
For example ProteinSolvation both works with GROMACS() and GROMACS("proteinsolvation").
It is ok if we always force the scenario specific executer, but that should be decided right now.

luispcunha:
> I think that here we need to distinguish between the scenarios that only work with the default executer and the ones that work with both of them. For example ProteinSolvation both works with GROMACS() and GROMACS("proteinsolvation"). It is ok if we always force the scenario specific executer, but that should be decided right now.

Ok.. I wasn't aware of that. Anyway passing GROMACS("proteinsolvation") or GROMACS() in order to decide if we do some kind of pre processing is a bit obscure. If for any other reason it should not use the scenario specific executer, we should control it with an optional argument that clearly identifies the difference.

IvanPombo:
> > I think that here we need to distinguish between the scenarios that only work with the default executer and the ones that work with both of them. For example ProteinSolvation both works with GROMACS() and GROMACS("proteinsolvation"). It is ok if we always force the scenario specific executer, but that should be decided right now.
> 
> Ok.. I wasn't aware of that. Anyway passing GROMACS("proteinsolvation") or GROMACS() in order to decide if we do some kind of pre processing is a bit obscure. If for any other reason it should not use the scenario specific executer, we should control it with an optional argument that clearly identifies the difference.

This notion is to run the pre-processing steps and post-processing steps obligatory, otherwise it does not make sense to have them. If they are optional then they should be posed in a different class, like the Bathymetries or so. Otherwise, we can also provide arguments to turn them on and off, but that change will only come later since it is of a different nature.

david451lima (approved):
My previous doubts weere addressed by @IvanPombo

PR #575, opened 2023-09-13
Remove n_cores param

begaiym-k:
Reflecting the latest change in the background, I am removing the n_cores parameter.

### DIFF
--- a/demos/py_scripts/swash.py
+++ b/demos/py_scripts/swash.py
@@ -22,7 +22,6 @@ flags.DEFINE_string("output_dir", None,
 
 flags.DEFINE_string("working_dir", None,
                     "Directory to which paths are relative to.")
-flags.DEFINE_integer("n_cores", 1, "Number of cores to use.")
 
 
 def main(_):
@@ -36,7 +35,6 @@ def main(_):
     task = swash_sim.run(
         input_dir=FLAGS.sim_dir,
         sim_config_filename=FLAGS.sim_config_filename,
-        n_cores=FLAGS.n_cores,
     )
 
     output = task.get_output(output_dir=FLAGS.output_dir)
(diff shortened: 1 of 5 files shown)


sarmento:
I am not sure if this is what we want to do. I think this parameters has an ambiguous name -- maybe it should be named n_cores -- but it is actually an important parameter that we may want to keep for the users to define, and also have ways to automatically assign it if the users does not define it.

Question to @IvanPombo : is this parameter related to the number of mpi threads / processes that are going to be run? If so, then the parameter should be named mpi_threads or execution_threads.

If the users does not define it (which is fair), then we should automatically set this. And it could be automatically set to the number of cores of the machine. 

So: can anybody clarify if this actually refers to the number of mpi threads?

IvanPombo:
> I am not sure if this is what we want to do. I think this parameters has an ambiguous name -- maybe it should be named n_cores -- but it is actually an important parameter that we may want to keep for the users to define, and also have ways to automatically assign it if the users does not define it.
> 
> Question to @IvanPombo : is this parameter related to the number of mpi threads / processes that are going to be run? If so, then the parameter should be named mpi_threads or execution_threads.
> 
> If the users does not define it (which is fair), then we should automatically set this. And it could be automatically set to the number of cores of the machine.
> 
> So: can anybody clarify if this actually refers to the number of mpi threads?

*SWASH*, *XBEACH*: they use number of cores to launch simulations
*OpenFOAM*: At least on the cloud, it can only use the number of physical cores... When trying to launch more processes it fails. Reading the stderr.txt of the simulation, it mentions that the number of slots available can be changed by creating a machine_file for it.

However, the whole point of this idea is that users will only have to configure the MachineGroup to decide the number of cores - I think @hpenedones has mentioned this in one of the last meetings. We can keep this on the client and still provide that option anyway.

PR #576, opened 2023-09-13
Move the rest of template files to general templates folder

IvanPombo:
This PR moves some of the template files that were still in their respective scenario folder into the general templates folder. Namely, the template files changed correspond to the scenarios: `FluidBlock` and `FluidTank`.

Test performed for FluidBlock (done for one of the children - DamBreak):
```python
import inductiva

inductiva.api_key = "API_KEY"

scenario = inductiva.fluids.DamBreak()
task = scenario.simulate()
task.get_output()
```

Test performed for FluidTank:
```python
import inductiva

inductiva.api_key = "API_KEY"

scenario = inductiva.fluids.FluidTank()
task = scenario.simulate()
task.get_output()
```


inductiva/fluids/dam_break/dam_break.py
```diff
@@ -53,7 +53,7 @@ def simulate(
         self,
         simulator: Simulator = DualSPHysics(),
         machine_group: Optional[resources.MachineGroup] = None,
-        device: Literal["cpu", "gpu"] = "gpu",
+        device: Literal["cpu", "gpu"] = "cpu",
```
sarmento, line 56: ok! I suspect that we are having problems with the GPU, right?
Should there be an issue?

inductiva/fluids/fluid_block/fluid_block.py
```diff
@@ -11,17 +11,20 @@
 from inductiva.fluids.fluid_types import FluidType
 from inductiva.simulators import SPlisHSPlasH
 from inductiva.simulators import DualSPHysics
-from inductiva.utils.templates import replace_params_in_template
+from inductiva.utils import templates
```
sarmento, line 14: here we either do just 

```
import inductiva
```
and then expand the whole thing below or, if we assumie that inductiva.utils is going to be something "standard" across we can do semething such as

```
import inductiva.utils as U
```

Please check with @hpenedones but I think moving into a world that we only have high level imports, will make it more obvious when we need to refactor the code (as we already did with scenarios).

Also, please take the chance to fix some class imports if possible.

inductiva/fluids/fluid_block/fluid_block.py
```diff
@@ -90,7 +93,7 @@ def simulate(
         simulator: Simulator = DualSPHysics(),
```
sarmento, line 93: no class imports...

luispcunha (approved):
lgtm. thanks @IvanPombo

inductiva/fluids/fluid_block/fluid_block.py
```diff
@@ -204,8 +209,10 @@ def _(cls, simulator: DualSPHysics):  # pylint: disable=unused-argument
 def _(self, simulator: DualSPHysics, input_dir):  # pylint: disable=unused-argument
     """Creates DualSPHysics simulation input files."""
 
-    replace_params_in_template(
-        template_path=os.path.join(os.path.dirname(__file__),
+    template_files_dir = os.path.join(SCENARIO_TEMPLATE_DIR,
+                                      DUALSPHYSICS_TEMPLATE_INPUT_DIR)
+    templates.replace_params_in_template(
```
sarmento, line 214:
```
templates.replace_params
```
no need to have "_in_template". 
The in_template suffix is only needed when you import the function, thereby loosing all the context.
Don't change it now, because it would require changes in many places, but let's discuss this with @hpenedones and how to systematically do this sort of changes across all code.

sarmento (approved):
LGTM!
Thank you for showing how you tested it :)
But there are a number of other issues that need to be fixed later.

inductiva/fluids/dam_break/dam_break.py
IvanPombo: We are moving to fix this, but there should be some introspection on the backend to understand if the requests we are asking here are being fulfilled!

inductiva/fluids/fluid_block/fluid_block.py
IvanPombo: I feel that doing that here, will be a big a pain for testing all of the things. Moreover, I think we should first clear out the importing structure we want to provide, and only afterwards move into these changes.

inductiva/fluids/fluid_block/fluid_block.py
hpenedones: Yes, if we follow the Google python guidelines we won't be importing classe or methods directly, therefore you will have the module "templates" and will call `templates.replace_params` (like you just did), so there's no need for the redundancy in the method name... Cool

PR #580, opened 2023-09-13
Add a function to get a list of MachineGroup objects

begaiym-k:
Here, I am adding the functionality that @hpenedones mentioned the other day. Also, did some changes to the doctoring and documentation.


inductiva/resources/resources.py
```diff
@@ -81,3 +99,13 @@ def get_machine_group(name: str):
 
     except inductiva.client.ApiException as api_exception:
         raise api_exception
+
+
+def get_machine_groups():
```
hpenedones, line 104: The pattern is typically the opposite to what : 
- the get method returns a list with MachineGroup objects
- the `list` calls the `get` method and does some pretty printing of its contents.

Similar to what is implemented for the tasks:
https://github.com/inductiva/inductiva/blob/20c315dad7cc180136d211ff5e179283569ad702/inductiva/tasks/methods.py#L131

inductiva/resources/resources.py
```diff
@@ -33,11 +37,25 @@ def _machine_group_list_to_str(machine_groups) -> str:
...
+    The name of the machine group can be used to retrieve a MachineGroup object
+    with the 'get_machine_group' function.
 
     Returns:
         List of machine group names.
     """
+    # pylint: enable=line-too-long
     try:
         api = instance_api.InstanceApi(inductiva.api.get_client())
         response = api.list_active_user_instance_groups()
```
hpenedones, line 61: In this list method, I think using logging.info is a bit ugly and clutters the output, consider just printing to stdout. Again consistent with:

https://github.com/inductiva/inductiva/blob/20c315dad7cc180136d211ff5e179283569ad702/inductiva/tasks/methods.py#L133

inductiva/resources/resources.py
```diff
@@ -33,11 +37,25 @@ def _machine_group_list_to_str(machine_groups) -> str:
 
 
 def list_active_machine_groups():
```
hpenedones, line 39: you can just call it `list_machine_groups` (consistent with the `get_machine_groups`).
Potentially even better would be `inductiva.resources.machine_groups.list()`  and `inductiva.resources.machine_groups.get()` which would be more consistent with the tasks...

Wdyt @begaiym-k and @luispcunha ?

hpenedones (changes requested):
Thanks @begaiym-k , please chat with @luispcunha on how to make this the most consistent possible with the tasks interface. I left some suggestions. Keep up the great work!

hpenedones:
Ah, now I see the problem and where my confusion came from. Actually the "list_machine_groups" method is also returning a list of things, in addition to printing things. In other words, it's working both as a "get" function AND a "print" method, all in one. Try to split those two functionalities, like in the tasks module...

sarmento:
Going fwd don't forget to add a section "How did I test this?" to your PR :)

begaiym-k:
> Thanks @begaiym-k , please chat with @luispcunha on how to make this the most consistent possible with the tasks interface. I left some suggestions. Keep up the great work!

Thank @hpenedones! I addressed most of your comments. 
Regarding the `get()` functions, I put two functions there `get()` (return a single MahcineGroup) and `get_all()` (return a list of MachineGroup objects). I understand we are trying to make it similar to the tasks, but in tasks one may want to proceed with tasks in batches since that's the whole idea, but with the machines I don't think that one will have such large batches of machine groups, so I though it would be more useful to give a user an option to get a single machine group.

I changes the path to the function as you suggested, so now it is `inductiva.resources.machine_groups.list()`

inductiva/resources/machine_groups.py
```diff
@@ -0,0 +1,106 @@
...
+    INFO:absl:Active machine groups:
+                                        Name         VM Type   # machines    Disk Size in GB       Spot         Created at
+    api-1b1f724c-5cfe-4d87-8439-9689aa139723   c2-standard-4            1                 40      False   13 Sep, 07:38:50
+    api-8e6bf7d8-4888-4de9-bda5-268484b46e6f   c2-standard-4            1                 40      False   13 Sep, 07:37:49
+
+    The name of the machine group can be used to retrieve a MachineGroup object
+    with the 'get' function."""
+    # pylint: enable=line-too-long
+
+    machine_groups = _fetch_machine_groups_from_api()
```
luispcunha, line 72: here I think it would be better to have a call to `get_all()`, and then the method `_machine_group_list_to_str()` would use the machine group objects to print, instead of parsing the response body (which you already do in MachineGroup.from_api_response().

inductiva/resources/machine_groups.py
```diff
@@ -0,0 +1,106 @@
...
+        api = instance_api.InstanceApi(inductiva.api.get_client())
+        response = api.get_instance_group({"name": name})
+
+        return resources.MachineGroup.from_api_response(response.body)
+
+    except inductiva.client.ApiException as api_exception:
+        raise api_exception
+
+
+def get_all():
```
hpenedones, line 95: I think that in the task we don't have get and get_all, and also: list is listing all but is not called "list_all" . I think it would be enough to have a get method that return a list with all the  machine group objects, like we did in tasks, no?

luispcunha (approved):

inductiva/resources/machine_groups.py
begaiym-k: I am not sure about it, but I can comment out get() function for now and see based on the use cases  what type of functions make the most sense and are useful, wdyt?

inductiva/resources/machine_groups.py
begaiym-k: I implemented that!

inductiva/resources/README.md
```diff
@@ -87,11 +87,13 @@ You can also list your active machine groups, and use the names that appear in t
...
+#                                     Name         VM Type   # machines    Disk Size in GB       Spot         Created at
+# api-1b1f724c-5cfe-4d87-8439-9689aa139723   c2-standard-4            1                 40      False   13 Sep, 07:38:50
+# api-8e6bf7d8-4888-4de9-bda5-268484b46e6f   c2-standard-4            1                 40      False   13 Sep, 07:37:49
 
-# Create a MachineGroup object by using its name and resume using it
+# Create a MachineGroup object by using its name and resume it using
 mg = inductiva.resources.get_machine_group("api-24e497af-d135-4a59-bdd1-854bf0176cbf")
 
+# Or get a list of all the MachineGroup objects (for example, if you want to terminate it at once)
+mg_list = inductiva.resources.get_machine_groups()
```
hpenedones, line 98: Outdated example?

hpenedones (approved):
Thanks Begaiym! Left a minor comment

PR #581, opened 2023-09-13
[FluidBlock] Fix template name

IvanPombo:
This fixes the template name of splishsplash for the fluid block scenario. It was missing splishsplash in the name.


inductiva/fluids/fluid_block/fluid_block.py
```diff
@@ -62,15 +62,20 @@ def __init__(self,
         """Initializes the fluid block scenario.
 
         Args:
-            density: The density of the fluid in kg/m^3.
-            kinematic_viscosity: The kinematic viscosity of the fluid, in m^2/s.
+            density: The density of the fluid in kg/m^3. Valid
+                range is [400, 2000] Kg/m3.
```
hpenedones, line 66: Will those limits work? The error I got from DualSPhysics was:

```
*** Exception (JSphCpuSingle::LoadConfigCommands) at DualSPHysics/source/JSph.cpp:865
Text: The reference density value 1.000000 is outside the defined limits [700.000000,1300.000000].
```

inductiva/fluids/fluid_block/fluid_block.py
```diff
@@ -62,15 +62,20 @@ def __init__(self,
...
+            kinematic_viscosity: The kinematic viscosity of the fluid, in m^2/s. 
+                Reference value for water is 1e-6 m^2/s.
             dimensions: The fluid block dimensions (in x, y, z), in meters.
             position: The position of the fluid block in the tank (in x, y, z),
               in meters.
             initial_velocity: The initial velocity of the fluid block (in x, y,
               z), in m/s.
         """
 
+        if not (400 <= density <= 2000):
```
cristianacarpinteiro, line 76: maybe you can do the assert instead    ` assert 400 <= density <= 1000 `

cristianacarpinteiro (approved):

inductiva/fluids/fluid_block/fluid_block.py
IvanPombo: Yes, these will work for now - we look into some of the physical aspects of fluids and SpliSHPlasH to make further sense of the values.

However, I want to highlight that DualSPH performance is hindered by density values further way from those of water, which is 1000 Kg/m^3. However, since this is more like a toy scenario I think we can keep this values and it will still work!

inductiva/fluids/fluid_block/fluid_block.py
IvanPombo: assert can be deactivated and are more used for debugging, I believe. This needs to be performed at all run times to warn users there values are unfeasible!

Issue #561, opened 2023-09-11
[ProteinSolvation Post-Processing] Incorrect path to file of RMSF


IvanPombo:
We passed the RMSF values generation to the backend and created an interface on the client to visualise the results. 

It is not working properly, since the path is not joined correctly with the output folder on the client side.

Currently:
```python
RMSF_VALUES_FILE = "rmsf_values.npy"
rmsf_values = np.load(RMSF_VALUES_FILE)
```

How it should be:
```python
RMSF_VALUES_FILE = "rmsf_values.npy"
rmsf_values=np.load(os.path.join(self.sim_output_path, RMSF_VALUES_FILE))
```

This implies that the visualizations presented on the documentation don't work on Colab by copy-paste.

PR #582, opened 2023-09-13
[ProteinSolvation] Fix path for RMSF values

cristianacarpinteiro:
Adressing issue #561 and fixing path for RMSF values.


IvanPombo (approved):
Thanks for this! LGTM!

PR #587, opened 2023-09-14
[SPH] Clean Simulators and Imports

IvanPombo:
This PR starts simplifying the naming convention for the Simulators - just one capital letter at the start - and takes the time to fix all of the imports in the respective scenarios according to Google style guide.

To verify everything was running smoothly I perform the next three tests:
```python
import inductiva

inductiva.api_key = "API_KEY"

scenario = inductiva.fluids.DamBreak()
task = scenario.simulate()
output = task.get_output()

output.render()
```
Also tested with the non-default simulator `inductiva.simulators.Splishsplash()`.

```python
import inductiva

inductiva.api_key = "API_KEY"

scenario = inductiva.fluids.FluidBlock(
    density=1000,
    kinematic_viscosity=1e-6,
    dimensions=[0.2,0.2,0.1]
)

task = scenario.simulate()

output = task.get_output()

output.render()
```
Also tested with the non-default simulator `inductiva.simulators.Splishsplash()`.

```python
import inductiva

inductiva.api_key = "API_KEY"

scenario = inductiva.fluids.FluidTank()

task = scenario.simulate()

output = task.get_output()

output.render()
```


luispcunha:
Why Splishsplash and not SplishSplash?

IvanPombo:
> Why Splishsplash and not SplishSplash?

Because than you can raise the question: Dualsphysics or DualSPHysics, OpenFOAM, OpenFoam or Openfoam... Here, the objective is to just simplify the naming convention and follow just the class structure. Considering that the name of the simulators are a single name - a brand name if you wish - and not an aggregate of things.

luispcunha (approved):
this seems quite a philosophical change, but I don't oppose. 
also, thanks for fixing some of the imports

IvanPombo:
> this seems quite a philosophical change, but I don't oppose. also, thanks for fixing some of the imports

I believe we should follow the philosophy of simplifying the life of the users - even myself have struggled with some of the names.

hpenedones:
> > this seems quite a philosophical change, but I don't oppose. also, thanks for fixing some of the imports
> 
> I believe we should follow the philosophy of simplifying the life of the users - even myself have struggled with some of the names.

This is a "philosophical" question only if we were trying to define the guidelines ourselves, but we have taken that decision long ago, we follow the Google Python guidelines, that already specifies this. 
Class names are CamelCase, which means capital letter in the beginning of each word.

https://google.github.io/styleguide/pyguide.html#3163-file-naming

This means that we should use: SplishSplash (splish and splash are two words).
The only thing that is somewhat under-specified in the Google python guidelines is what to do with acronyms.

In this rare cases of ambiguity, let's from now on fall back to the other super common python guideline, which is:
https://peps.python.org/pep-0008/#descriptive-naming-styles

they say
"Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError."

Therefore, OpenFOAM would be the right form, because FOAM is an acronym that stands for "Field Operation and Manipulation".

IvanPombo:
> > > this seems quite a philosophical change, but I don't oppose. also, thanks for fixing some of the imports
> > 
> > 
> > I believe we should follow the philosophy of simplifying the life of the users - even myself have struggled with some of the names.
> 
> This is a "philosophical" question only if we were trying to define the guidelines ourselves, but we have taken that decision long ago, we follow the Google Python guidelines, that already specifies this. Class names are CamelCase, which means capital letter in the beginning of each word.
> 
> https://google.github.io/styleguide/pyguide.html#3163-file-naming
> 
> This means that we should use: SplishSplash (splish and splash are two words). The only thing that is somewhat under-specified in the Google python guidelines is what to do with acronyms.
> 
> In this rare cases of ambiguity, let's from now on fall back to the other super common python guideline, which is: https://peps.python.org/pep-0008/#descriptive-naming-styles
> 
> they say "Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError."
> 
> Therefore, OpenFOAM would be the right form, because FOAM is an acronym that stands for "Field Operation and Manipulation".

Yes Hugo, but that is considering you are joining separate words. Here Splishsplash is to refer to the name of the simulator itself, so I am not sure if we should capitalize the two words like that. Then, we have DualSPhysics?

In this sense, it starts getting weirder with the names, for example SWASH is due to: ["Simulating WAve till SHore"](https://swash.sourceforge.io/home.htm) should it be: SWaSh?

My idea with all of this is that the name of the simulator is already a brand name - like Inductiva, McDonalds - so either we keep the name as it is usually used by the creators or just simplify it to first capital letter.

hpenedones:
SWASH is an acronym.

"My idea with all of this is that the name of the simulator is already a brand name - like Inductiva, McDonalds - so either we keep the name as it is usually used by the creators or just simplify it to first capital letter."  - This is against the standards that we are following. The answer is no.

PR #588, opened 2023-09-14
[Proteins] Added IPython to molecules_extra extended package

david451lima:
Added Ipython to the requirements in the molecules_extra package. It was missing from the stup.cfg file and its used in the protein post-processing


IvanPombo (approved):
Thanks! Nice catch!

cristianacarpinteiro (approved):
Thank you :)

PR #589, opened 2023-09-14
[CoastalArea] Check if bathymetry is uniform at plot()

david451lima:
First in a serie of micro PRs, each to address a specific issue with CoastalArea.
In this one, in the plot() method, we check if the bathymetry is uniform at the start of the method, in order for the _interpolate_to_uniform_grid to not be raised all the times, but only when in fact the grid is not uniform.

### DIFF
--- a/inductiva/coastal/bathymetry.py
+++ b/inductiva/coastal/bathymetry.py
@@ -155,7 +155,7 @@ class Bathymetry:
         boundary (East) are assumed to be below water (i.e. have 0 < depths <
         `max_depth`). The corners on the upper x boundary (West) are assumed to
         be above water (i.e. have - `max_depth` < depths < 0).
-        
+
         Args:
             x_range: The range of x values, in meters.
             y_range: The range of y values, in meters.
@@ -251,7 +251,7 @@ class Bathymetry:
     @optional_deps.needs_coastal_extra_deps
     def x_uniques(self, sort: bool = False) -> np.ndarray:
         """Returns the unique x values.
-        
+
         Args:
             sort: Whether to sort the unique values.
         """
@@ -263,7 +263,7 @@ class Bathymetry:
     @optional_deps.needs_coastal_extra_deps
     def y_uniques(self, sort: bool = False) -> np.ndarray:
         """Returns the unique y values.
-        
+
         Args:
             sort: Whether to sort the unique values.
         """
@@ -278,10 +278,10 @@ class Bathymetry:
              y_range: Sequence[float],
              remove_offset=True):
         """Crops the bathymetry to a given range of x and y values.
-        
+
         Args:
             x_range: The range of x values, in meters.
-            y_range: The range of y values, in meters.    
+            y_range: The range of y values, in meters.
         """
         x_min, x_max = x_range
         y_min, y_max = y_range
@@ -336,14 +336,14 @@ class Bathymetry:
         The bathymetry data is plotted with a color plot on a grid with uniform
         spacing in the x and y directions. The spacing in the x and y directions
         is controlled by the `x_resolution` and `y_resolution` arguments.
-        
+
         The data is interpolated from the points where the bathymetry is defined
         to the uniform grid using linear interpolation.
 
         Points on the uniform grid at a distance larger than a threshold
         distance `threshold_distance` from points where the bathymetry is
         defined are omitted.
-    
+
         The plot is produced with matplotlib.
 
         Args:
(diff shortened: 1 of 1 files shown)


inductiva/coastal/bathymetry.py
```diff
@@ -363,13 +363,14 @@ def plot(
...
 
-        logging.info("Plotting the bathymetry on a uniform grid...")
-        depths_grid, _ = self._interpolate_to_uniform_grid(
-            x_resolution,
-            y_resolution,
-            threshold_distance=threshold_distance,
-            nullable=True,
-        )
+        if not self.is_uniform_grid():
+            logging.info("Plotting the bathymetry on a uniform grid...")
```
IvanPombo, line 367: I agree with this, but I am just thinking if we should keep the x_resolution, y_resolution at all or some better default values...

Moreover, can you comment if it is taking to much time to plot?

IvanPombo (approved):
Clean! Just a minor question and a consideration, but I think we can move forward and address it later if confusing.

inductiva/coastal/bathymetry.py
david451lima: Is plotting quite quickly, it doesnt take a lot of time

IvanPombo: Nice!

inductiva/coastal/bathymetry.py
```diff
@@ -363,13 +363,14 @@ def plot(
...
         """
 
-        logging.info("Plotting the bathymetry on a uniform grid...")
-        depths_grid, _ = self._interpolate_to_uniform_grid(
-            x_resolution,
-            y_resolution,
-            threshold_distance=threshold_distance,
-            nullable=True,
-        )
+        if not self.is_uniform_grid():
```
sarmento, line 366: Can you add a comment on why would this work? How does the bathymetry "know" that is uniform_grid?
What does that method do?

sarmento:
Can you show that piece of code that I added to the issue now runs as expected?

david451lima:
I implemented a refactor, since the interpolation was still required if the user wanted to implement a x or y resolution.
Now, the code:
-Does no interpolation and plots with the bathymetry's resolution if the user doesn't specify anything;
-Does interpolation if the user choses a non-uniform bathymetry OR if a resolution is specified.


Test:
 - No logging for .plot() call:
image


- Logging if the resolution is specified:

image

inductiva/coastal/bathymetry.py
```diff
@@ -334,11 +346,14 @@ def plot(
...
-        spacing in the x and y directions. The spacing in the x and y directions
-        is controlled by the `x_resolution` and `y_resolution` arguments.
+        spacing in the x and y directions. If the user does not specify the
+        resolution, the plotted mesh has the bathymetry resolution. If
+        necessary, the user can control the spacing in the x and y directions
+        using the `x_resolution` and `y_resolution` arguments.
 
-        The data is interpolated from the points where the bathymetry is defined
-        to the uniform grid using linear interpolation.
+        For non-uniform grids of if the resoltuion is specified, the data is
```
IvanPombo, line 354: typo *or

inductiva/coastal/bathymetry.py
```diff
@@ -355,30 +370,40 @@ def plot(
...
+            x_size = self.x_num()
+            y_size = self.y_num()
 
-        depths_grid, _ = self._interpolate_to_uniform_grid(
-            x_resolution,
-            y_resolution,
-            threshold_distance=threshold_distance,
-            nullable=True,
-        )
+            depths_grid = self.depths.reshape((x_size, y_size))
```
IvanPombo, line 388: Why you need this reshape? Isn't the bathymetry already like this?

david451lima: No, the bathymetry is a one dimensional array

sarmento: 1D array. That was unexpected...
OK.

inductiva/coastal/bathymetry.py
```diff
@@ -355,28 +370,40 @@ def plot(
...
+                x_resolution,
+                y_resolution,
+                threshold_distance=threshold_distance,
+                nullable=True,
+            )
+
+            x_size = self.x_ptp() / x_resolution
+            y_size = self.y_ptp() / y_resolution
 
         if x_size > 1000 and y_size > 1000:
```
sarmento, line 403: I don't think you need this....

sarmento (approved):

inductiva/coastal/bathymetry.py
david451lima: You mean the warning for large x and y sizes?

sarmento: yes... why not 800 x 800 or 2000 x 2000?
If we don't have a good number, we don't do it.
If we want to educate the user, then we measure the time a log.info how long it took so that the user then decides...

PR #591, opened 2023-09-14
[Protein Solvation] Download examples from RSCB 

cristianacarpinteiro:
In this PR, I've added a submodule called examples to the molecules module. It consists of a set of functions to download proteins directly from the RSCB database. 
I've also updated the example in the README to use the lysozyme, since alanine is too small and the resulting visualization doesn't have a backbone. I think a larger protein is best to showcase what we can do. 
In another PR I will update the images in the README. 

## How did I test this? 
Normal behaviour: 
image

Incorrect pdb_id: 
image


inductiva/molecules/examples.py
```diff
@@ -0,0 +1,36 @@
+"Set of functions to load examples of molecules."
```
IvanPombo: This is super cool! Please just notice if it works on Colab due to drive mounts - we may still need to do them, but it is important just to be aware.

inductiva/molecules/protein_solvation/README.md
```diff
@@ -38,17 +38,19 @@ To test this scenario we have available one example of a protein PDB file - [dow
 
 inductiva.api_key = "YOUR_API_KEY"
 
- # Initialize the scenario
- scenario = inductiva.molecules.ProteinSolvation(
-     protein_pdb = "alanine.pdb",
+# Initialize the scenario
+lysozyme = inductiva.molecules.examples.load_lysozyme()
```
IvanPombo, line 42: How long does it take this simulation in the default machines?

setup.cfg
```diff
@@ -39,6 +39,7 @@ molecules_extra =
   MDAnalysis
   ipywidgets
   google.colab
+  requests
```
IvanPombo, line 42: Maybe this should live in the default installation, since otherwise it creates a barrier from the simple installation with `pip install inductiva`.

I think it is okay to be there, since it should be a simple package, however verify if it does not cause any problems during installation.

IvanPombo (commented):
Left a few comments to iterate and test a bit further.

inductiva/molecules/protein_solvation/README.md
cristianacarpinteiro: It took about 5 minutes.

setup.cfg
luispcunha: why install requests? we have urllib3 already as a dependency, we can use it to download files.

cristianacarpinteiro: Ah good catch! I will do it with urllib.

sarmento:
Can you please add a section with "How did I test this?"

setup.cfg
cristianacarpinteiro: I've fixed it to use urllib.

inductiva/molecules/examples.py
```diff
@@ -0,0 +1,36 @@
+"""Functions to download and load example molecules for GROMACS simulations 
+from the RCSB PDB database."""
```
david451lima, line 2: This script is super cool! Great effort in developing it unprompted

inductiva/molecules/protein_solvation/README.md
```diff
@@ -38,25 +38,27 @@ To test this scenario we have available one example of a protein PDB file - [dow
...
+scenario = inductiva.molecules.ProteinSolvation(
+     protein_pdb = lysozyme,
      temperature = 300)
 
- # Run a simulation
- task = scenario.simulate(simulation_time_ns = 0.5)
+# Run a simulation
+task = scenario.simulate(simulation_time_ns = 0.05)
```
david451lima, line 48: If this still takes 5 mins I would decrease the time in another order of magnitude. Users should not have to wait so much

david451lima (approved):

inductiva/molecules/protein_solvation/README.md
cristianacarpinteiro: I can make it 0.01 and then it would take 2m, but then the output will have only 10 frames. But I think that for an example its okay.

sarmento:
THANK YOU! :)

inductiva/molecules/protein_solvation/README.md
cristianacarpinteiro: I changed it for 0.01! I don't want to decrease it too much because the output will have very few frames.

IvanPombo: But maybe we should keep alanine and show more frames, decide what you think is best here... I know that your goal is for the post-processing part, but there you can explain and introduce the lysozyme for having the backbone plotted :)

inductiva/molecules/examples.py
```diff
@@ -1,23 +1,22 @@
 """Functions to download and load example molecules for GROMACS simulations 
 from the RCSB PDB database."""
-import requests
-from inductiva.utils import optional_deps
+import urllib
+
 
-@optional_deps.needs_molecules_extra_deps
 def load_hemoglobin():
```
luispcunha: if we're downloading a file onto the user's computer, we should allow an optional argument for the user to decide where to store the file. We should probably also be using the `inductiva.working_dir` global var

inductiva/molecules/examples.py
```diff
@@ -27,15 +26,11 @@ def download_from_rcsb(pdb_id):
...
-    if response.status_code == 200:
-        local_file_path = f"{pdb_id}.pdb"
-        # Save the response content (PDB file) to the local file
-        with open(local_file_path, "wb") as pdb_file:
-            pdb_file.write(response.content)
+    local_file_path = f"{pdb_id}.pdb"
 
+    try:
+        urllib.request.urlretrieve(api_url, local_file_path)
         print(f"File downloaded to {local_file_path}")
```
luispcunha, line 33: use the absl.logging module for consistency with the rest of the library

inductiva/molecules/protein_solvation/README.md
```diff
@@ -38,25 +38,27 @@ To test this scenario we have available one example of a protein PDB file - [dow
...
 The last code line downloads the files necessary to render and post-process our simulation.
 
 ## Output and Post-processing 
 The simulation output folder contains trajectory data spanning the simulation duration and the protein's topology. These files are necessary to generate an interactive trajectory visualization using [NGLview](https://github.com/nglviewer/nglview). Users have the flexibility to select the protein's representation type, including options to display or hide the backbone, and they can also specify the criteria for visualization selection. You can refer to the [documentation](https://nglviewer.org/ngl/api/manual/usage/selection-language.html) for guidance on creating custom selections according to your preferences.
 
 ```python
- output = task.get_output()
- output.render_interactive(representation="ribbon", add_backbone=True, selection="protein")
+output = task.get_output()
+output.render_interactive(representation="ribbon", add_backbone=True, selection="protein")
```
luispcunha, line 61: this is also the example we have in the main readme right? we should probably change it, or add some information saying that the render works only on notebooks and requires the molecules_extra

luispcunha (commented):
left some feedback for improvement. nice work, thanks @cristianacarpinteiro

inductiva/molecules/protein_solvation/README.md
cristianacarpinteiro: Yes, I will update it in another PR :)

inductiva/molecules/protein_solvation/README.md
cristianacarpinteiro: I change the example to be insulin, since it is smaller and now takes 1 minute.

inductiva/molecules/examples.py
cristianacarpinteiro: image
It does :)

PR #594, opened 2023-09-14
Address the Simulators naming in the documentation

IvanPombo:


inductiva/coastal/README.md
```diff
@@ -1,45 +1,4 @@
-# Coastal dynamics
```
luispcunha, line 1: where is this going to?

IvanPombo: No where, for now! This part belongs on the simulators, which we are not casting the most focus now!

PR #600, opened 2023-09-14
Allow to download general example files

IvanPombo:
This PR builds on @cristianacarpinteiro for molecules in #591 and extends to allow download sample files from our Github resources page.

To test it one can do:
```python

import 

inductiva.examples.load_test_vehicle()

inductiva.examples.load_hemoglobin()
```

In turn you receive on your current working dir the "test_vehicle.obj" and "1A3N.pdb" protein.


inductiva/examples/examples.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    file_path = f"{pdb_id}.pdb"
+    api_url = f"https://files.rcsb.org/download/{file_path}"
+
+    file_path = download_from_external_url(api_url, file_path, save_dir)
+
+    return file_path
+
+
+def download_inductiva_resources(file_name: str,
```
cristianacarpinteiro, line 59: I think this file can me divided into three files. One with the fluids examples, one with the molecules examples and one with the general functions.

inductiva/fluids/wind_tunnel/README.md
```diff
@@ -22,22 +22,23 @@ Launching a simulation returns a task object, which can be used to verify the st
 
 ### Example:
 
-To run this example download a vehicle geometry in OBJ format from [here](/resources/geometry/test_vehicle.obj) and
-do not forget to insert your API Key (get one by filling this [form](https://docs.google.com/forms/d/e/1FAIpQLSflytIIwzaBE_ZzoRloVm3uTo1OQCH6Cqhw3bhFVnC61s7Wmw/viewform?usp=sf_link)).
+To run this example you will need a vehicle. We left you one sample [here](/resources/geometry/test_vehicle.obj), which you can load via `inductiva.examples.load_test_vehicle()`. Do not forget to insert your API Key (get one by filling this [form](https://docs.google.com/forms/d/e/1FAIpQLSflytIIwzaBE_ZzoRloVm3uTo1OQCH6Cqhw3bhFVnC61s7Wmw/viewform?usp=sf_link)).
```
cristianacarpinteiro, line 25: I would remove the part of where the file is located. Just say you can download it.

inductiva/examples/examples.py
```diff
@@ -0,0 +1,102 @@
+"""Functions to download and load examples from URLs.
```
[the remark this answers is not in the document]
sarmento: I agree! See issue: https://github.com/inductiva/inductiva/issues/604

inductiva/examples/examples.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+from typing import Optional
+from urllib import request, error
+
+from absl import logging
+
+import inductiva
+
+
+def load_insulin(save_dir: Optional[str] = None) -> str:
```
sarmento, line 13: There are many problems with this method. 
One is the naming, which obviously does not match the download_from_rcbp

Why would you even name a method *load* when what it does is *download* a file to a certain directory??
It also seems that we don't need to have these methods at all. The method   download_from_rcsb seems to be more than enough to illustrate our examples. It's actually perfect. Don't overcomplicate. Just use download_from_rcsb which is perfectly clear.

inductiva/examples/examples.py
```diff
@@ -0,0 +1,102 @@
...
+    return download_from_rcsb("1AKI", save_dir)
+
+
+def load_test_vehicle(save_dir: Optional[str] = None) -> str:
+    """Load a test vehicle from Inductiva Github.
+
+    Returns:
+        The path to the downloaded OBJ file."""
+
+    return download_inductiva_resources("test_vehicle.obj", save_dir)
```
sarmento, line 40: Same comment. Why not just use download_inductiva_resources 
It's perfect and self-explanatory!

Team: please don't make things more complicated than they need to be....!

sarmento (commented):
Please don't overcomplicate, ok?

Simplicity is one of our key tenets. Don't wrap methods in layers and layers that don't add any value, ok?

Be glad that I am cacthing this before @hpenedones did .... :)

inductiva/examples/examples.py
sarmento: OMG! We are already going for 3 files???? 
I can't take this anymore...!.... :)
What's wrong with you people?? :)

inductiva/examples/examples.py
cristianacarpinteiro: So you think we should remove the functions to download specific examples from RCSB ? For that the user needs to know the exact PDB ID for the protein. The wrapper would make the example super clean and a user could load a molecule without knowing anything about PDB IDs.

inductiva/examples/examples.py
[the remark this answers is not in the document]
IvanPombo: I am gonna address in the issue to centralize everything.

PR #603, opened 2023-09-14
[CoastalArea] Logs updated and README now has a low level example

david451lima:
This PR updates the logging of the scenario simulation. Now (tested with the test.py file, that contains all the code snippets present in the ReadME), the logging that is outputed is:
-INFO:absl:Generating random array with the Diamond-Square Algorithm...
INFO:absl:Task ID: 1694734554521773852
INFO:absl:Creating input archive...
INFO:absl:Input archive created.
INFO:absl:Input archive size: 103.49 KiB
INFO:absl:Uploading input...
INFO:absl:Input archive uploaded.
INFO:absl:Task request submitted.
INFO:absl:Waiting for resources...
INFO:absl:The task is being executed remotely.
INFO:absl:Task completed successfully.
INFO:absl:Task completed successfully.
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3.83M/3.83M [00:01<00:00, 3.32MiB/s]
INFO:absl:Saving movie to movie_path_besti.mp4

We no longer have the loggings about the uniform grids - the resolution is the same of the bathymetry.

Also the ReadME now has a much simpler example (randomly generated bathymetry)


inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
 
 ### Example
 
+Note: You can follow this example in a notebook or run the script `test.py`
```
sarmento, line 11: change the name test.py to something that is much more explicit. For example random_bathymetry_demo.py (if that is the case)

inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+available in this folder, that contains all the code snippets exhibited here.
+
```
sarmento, line 13: Don't mention that XYZ bathymetry until we need it.

inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
 Start by creating or loading a bathymetry. This can be done e.g. by reading an
 [ASCII XYZ file](https://emodnet.ec.europa.eu/sites/emodnet.ec.europa.eu/files/public/20171127_DTM_exchange_format_specification_v1.6.pdf),
 a standard format for bathymetry data (for examples, see e.g. the [European Marine Observation and Data Network](https://emodnet.ec.europa.eu/geoviewer/#!/)).
-Here we shall use the bathymetry data from Algarve available [here](https://sextant.ifremer.fr/record/SDN_CPRD_590_HR_Lidar_Algarve/)
-as an example.
+Here we shall use a artificial bathymetry, randomly generated by our code.
```
sarmento, line 17: To avoid having to download any data for now, we will start by run a simulation on synthetic a bathymetry , that is procedurally generated using the API.

inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+        max_depth=50)
 ```
 
-![Raw bathymetry.](resources/media/bathymetry.png)
-
-Raw bathymetry data will often span an area too large and sparsely sampled to
-be used in a simulation. In this case, the user can crop the bathymetry to a
-smaller area of interest:
+This bathymetry is defined in a 50x50 grid, that represent a domain 100 m
+wide and long. It can be inspected by plotting it.
```
sarmento, line 32: and 100m long

inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+        y_num=100,
+        max_depth=50)
 ```
 
-![Raw bathymetry.](resources/media/bathymetry.png)
-
-Raw bathymetry data will often span an area too large and sparsely sampled to
-be used in a simulation. In this case, the user can crop the bathymetry to a
-smaller area of interest:
+This bathymetry is defined in a 50x50 grid, that represent a domain 100 m
```
sarmento, line 31: 100m (no space)

inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+wide and long. It can be inspected by plotting it.
 
 ```python
-bathymetry = bathymetry.crop(x_range=(51000, 52000), y_range=(12150, 13000))
 bathymetry.plot()
```
sarmento, line 35: Does this plot actually work on the terminal.
According to Fabio, it is not supposed to work on the terminal (pls check his reply to my email).
What do we need to do to make it work? Is there any extra instruction we need to call?
(again: don't bias the development of the API toward jupyter or colab. Being able to run thing on the terminal is more important because the goal of the API is to provide functionalities to support longer pipelines)

inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
-case 200 m x 200 m:
-
-```python
-bathymetry.plot(x_resolution=200, y_resolution=200)
+bathymetry = inductiva.coastal.Bathymetry.from_random_depths(
+        x_range=(0, 100),
+        y_range=(0, 100),
+        x_num=100,
+        y_num=100,
+        max_depth=50)
```
sarmento, line 28: I just saw the png for this bathymetry. It's too flat :) Can you pls increase the roughness / noise that it become more epic?

inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
-In case the bathymetry is not defined on a regular grid when creating a
-scenario, interpolation to a regular grid is automatically attempted with a
-default resolution of 2 m x 2 m and no pre-configured strategy to fill depths at
-positions missing in the bathymetry.
-
 Once the scenario is created, run the simulation as follows:
 
 ```python
-task = scenario.simulate(simulation_time=100, output_time_step=1)
+task = scenario.simulate(simulation_time=30, output_time_step=1)
```
sarmento, line 54: I am nor sure if I understand what is the output_time_step. Are we dumping frames every second? Can you add a comment on that?

inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
 
 The user can specify the total simulation time and the time step between outputs
 (all in seconds).
 
-To visualize the results:
+To visualize the results, we can generate and save a movie of the simulation:
 
 ```python
-output.render()
+output.render(movie_path = "movie_path.mp4")
```
sarmento, line 65: It would be great if the render method logs something about where it is storing the file so that the user sees the message. Also, is this an operation that may take long? If so, make sure that it start by logging something like "Starting to render video from simulation data..." and end with "Done! Writing mp4 file to XXXX".

inductiva/coastal/test.py
```diff
@@ -0,0 +1,25 @@
...
+
+bathymetry = inductiva.coastal.Bathymetry.from_random_depths(x_range=(0, 100),
+                                                             y_range=(0, 100),
+                                                             x_num=100,
+                                                             y_num=100,
+                                                             max_depth=50,
+                                                             random_seed=12)
+
+bathymetry.plot()
+plt.show()
```
sarmento, line 14: Ah! Comment on this pls!
But we have a readability/pattern problem here.
The method  plot() gives *no* hint that we are using matplotlib, or that we then need to do

```
import matplotlib.pyplot as plt

plt.show()
```

It does not seem reasonable to assume that the user will know how to do that (I wasn't able to figure it out because I was really NOT expecting this pattern).

Either we have:
```
bathymetry.show()
```
that internally calls plot and plt.show without the user having to do imports or we allow plot to have an extra argument that allows us to optiobally show
```
bathymetry.plot(show=True)
```

@hpenedones : what do you think? We may need to go over the patterns we are using for visualizing objects, and there seems to be two different situations (basic / simple visualization such as this one, and more complex postprocessing / visualization methods). The simple ones, that require only matplotlib or something very standard can exist in the client side, but we need to make sure we follow a consistent way of calling them.

inductiva/coastal/test.py
```diff
@@ -0,0 +1,25 @@
...
+scenario = inductiva.coastal.CoastalArea(bathymetry=bathymetry,
+                                         wave_source_location="W",
+                                         wave_amplitude=5,
+                                         wave_period=5.5)
+
+task = scenario.simulate(simulation_time=30, output_time_step=1)
+
+output = task.get_output()
+
+output.render(movie_path="movie_path.mp4")
```
sarmento, line 25: I was looking at the movie, and it seems super fast. Why is that? What parameters are we NOT setting right? Is it fps? IT seems that if output_time_step is 1 then the video should be calibrated to 1 frame per second, right?
Please try to figure this out and also comment on what these parameters are :)

inductiva/coastal/README.md
```diff
@@ -8,83 +8,61 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
 
 ```python
-bathymetry = bathymetry.crop(x_range=(51000, 52000), y_range=(12150, 13000))
 bathymetry.plot()
 ```
 
-![Cropped bathymetry.](resources/media/bathymetry_cropped.png)
+![Raw bathymetry.](/resources/bathymetry/bathymetry_random.png)
 
-To be used in a simulation, the bathymetry data must be interpolated to a
```
sarmento, line 44: Don't delete this example because it's a good one. It's just not the ideal first example. Pls just move it below to a section called "Simulating with real bathymetries", and make sure you can run it too.

sarmento (commented):
Thank you very much David!
This is almost done! 
Please do not delete Fabio's example because it is a good one. It's just not ideal as a first example. But it's a great second example, once users already know that they can run a simpler case. Pls just move Fabio's example to a second section.
I am very excited about this CoastalArea scenario.

inductiva/coastal/test.py
```diff
@@ -0,0 +1,25 @@
+"""End to end CoastalArea scenario simulation"""
```
IvanPombo, line 1: This should not be uploaded as a test to the package, because it will fail due to not having a key.

IvanPombo (commented):
Besides Sarmento comments, and the removal of the test.py everything looks!

inductiva/coastal/README.md
david451lima: About this: I generated bathymetries with higher roughness factors, and the simulations "exploded" - the water levels went to unphysical values (>500 m).
While I dont understand the source of this error I am not confortable in increasing the roughness factor in the example

inductiva/coastal/README.md
```diff
@@ -8,10 +8,76 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
 
 ### Example
 
-Start by creating or loading a bathymetry. This can be done e.g. by reading an
+Note: You can follow this example in a notebook or run the script
+`random_bathymetry_demo.py` available in this folder, that contains all the code
```
IvanPombo, line 12: I still don't think this should live inside the coastal module. Move it to the demos.

inductiva/coastal/README.md
```diff
@@ -8,10 +8,76 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+snippets exhibited here. Don't forget to insert your API key for the script to work!
```
IvanPombo, line 13: Add here the APi key form. Check the windtunnel.

inductiva/coastal/README.md
```diff
@@ -8,10 +8,76 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+
+The bathymetry that we use can be loaded from a real scenario or randomly generated
+by us. To avoid having to download any data for now, we will start by run a simulation
+on synthetic bathymetry , that is procedurally generated using the API.
```
IvanPombo, line 17: on a synthetic bathymetry, ...

inductiva/coastal/README.md
```diff
@@ -8,10 +8,76 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+
+We start by generating a random bathymetry.
+```python
```
IvanPombo, line 20: Let's have a fully working python code snippet from end-to-end with bathymetry generation and simulation. Hence, people can copy paste it.

Explain the current sections between code in a comment above the line of code.

inductiva/coastal/README.md
```diff
@@ -8,10 +8,76 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+This bathymetry is defined in a 50x50 grid, that represent a domain 100m
+wide and 100m long. It can be inspected by plotting it.
+
+```python
+bathymetry.plot()
+```
+
+![Raw bathymetry.](/resources/bathymetry/bathymetry_random.png)
+
+The bathymetry is ready to be used in a simulation scenario. In this case, we set
```
IvanPombo, line 40: This could be explained before if you want. But notice that in most Readmes you are just providing the example.

inductiva/coastal/test.py
IvanPombo, line 1: Again, remove this, since it will break the pytest pipeline without an API Key.

IvanPombo (approved):
Left you with a set of comments to have a single code snippet.

david451lima:
I will keep the code snippets and the whole file. I add a recomendation for the user to specify the inductiva.api_key for the script to run.
With this, I can have both the code snipets intertwined with comments, which is more informative for the user; and have the full code in a script. This is further more confortable for terminal-only users.

inductiva/coastal/random_bathymetry_demo.py
IvanPombo, line 1: I understand your intentions, but this should also not be inside the package. It is a demo, it is named as a demo, hence should leave inside the demos folder. Please remove it, it is inconsistent with our workflow.

inductiva/coastal/README.md
```diff
@@ -8,10 +8,76 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+
+bathymetry = inductiva.coastal.Bathymetry.from_random_depths(
+        x_range=(0, 100),
+        y_range=(0, 100),
+        x_num=100,
+        y_num=100,
+        max_depth=50)
+```
+
+This bathymetry is defined in a 50x50 grid, that represent a domain 100m
```
IvanPombo, line 33: This can be a comment. Moreover, I do not see why this is a 50x50 grid. The code is readable and explained well. Just add a top comment to mention the units and that is it.

inductiva/coastal/README.md
```diff
@@ -8,10 +8,76 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+wide and 100m long. It can be inspected by plotting it.
+
+```python
+bathymetry.plot()
+```
+
+![Raw bathymetry.](/resources/bathymetry/bathymetry_random.png)
+
+The bathymetry is ready to be used in a simulation scenario. In this case, we set
+the wave source location to the west boundary of the domain (i.e. the lower x
```
IvanPombo, line 43: This is redundant with the names of the args below. I still vouch to remove this and unify this in a single script to copy+paste.

IvanPombo (commented):
@david451lima I still think there are some issues that weren't solved in this PR. Can you comment more specifically on the comments I made...

PR #605, opened 2023-09-15
Update resolve_path and use it correctly in download files

luispcunha:
This fixes some issues I encountered trying to use the protein solvation scenario with example proteins:
- `utils.files.resolve_path()` was being used incorrectly (I now documented it better). 
- Also, made the download return an absolute path to avoid errors in further processing.
- Explicitly imported `urllib.request` and `urllib.error` because I was having errors where `request` and `error` were not imported by just importing `urllib`.


inductiva/molecules/examples.py
```diff
@@ -42,14 +43,18 @@ def download_from_rcsb(pdb_id: str, save_dir: Optional[str] = None) -> str:
...
     if save_dir is not None:
-        save_dir = inductiva.utils.files.resolve_path(save_dir)
-        if not os.path.exists(save_dir):
-            os.makedirs(save_dir)
-        local_file_path = os.path.join(save_dir, local_file_path)
+        local_file_path = pathlib.Path(save_dir, local_file_path)
+
+    local_file_path = inductiva.utils.files.resolve_path(local_file_path)
+
+    if not local_file_path.parent.exists():
```
cristianacarpinteiro, line 50: Thank you for fixing it!

cristianacarpinteiro (approved):
Thank you Luis for fixing this and implementing all the tests

Issue #604, opened 2023-09-15
[Example data and Assets]: location in the package, naming, etc


sarmento:
It's good that we are adding some data artifacts to the code base to help running examples in an easier way.
However, we really need to think about where we place them in the structure.
For example, this:
```
inductiva.examples.load_insulin()
```
is not a good location since it suggest that we can use this on any domain.
The previous version:
```
inductiva.molecules.examples.load_insulin()
```
was better since, at the very least, it would define the context in which this example works.
But, I also don't think this is still the right way of doing it. 

a) Do we need to place it under examples package (weird)
b) Naming is very confusing too: what does this method return?

Let's discuss this.

cristianacarpinteiro:
I agree with you. But since there may be some functions that are necessary for all downloads, maybe we can do: 

`inductiva.examples.molecules.load_insulin()`  and inside the examples module we have the functions necessary to fetch all the examples. 
Regarding the name I was trying to follow  something like [scikit-learn](https://scikit-learn.org/stable/datasets/toy_dataset.html). But maybe I can change it to` download_insulin_protein()` to be more specific to what the function actually does.

sarmento:
No! No! No!  :)
Why are we making things overly complicated?
People: I can't always be fighting this war, ok?

PR #606, opened 2023-09-15
Download files from urls

IvanPombo:
This PR addresses the discussion that occured in #600 and in #604.

Now, to download a pdb file from the RCSB database one does
`inductiva.molecules.utils.download_pdb_from_rcsb(pdb_id="1ZNI").

To download general files from urls we can do:
```python
test_object_url = "https://raw.githubusercontent.com/inductiva/inductiva/main" \
                              "/resources/test_vehicle.obj"
object = inductiva.utils.files.download_from_url(url=test_object_url, local_file_path="test_object.obj")
```

The implementation was tested with the code snippets present in the Readmes of the ProteinSolvation and the WindTunnel scenarios.


inductiva/fluids/wind_tunnel/README.md
```diff
@@ -22,22 +22,29 @@ Launching a simulation returns a task object, which can be used to verify the st
 
 ### Example:
 
-To run this example download a vehicle geometry in OBJ format from [here](/resources/geometry/test_vehicle.obj) and
-do not forget to insert your API Key (get one by filling this [form](https://docs.google.com/forms/d/e/1FAIpQLSflytIIwzaBE_ZzoRloVm3uTo1OQCH6Cqhw3bhFVnC61s7Wmw/viewform?usp=sf_link)).
+To run this example you need a vehicle. You can fetch any vehicle you like directly with Inductiva package. On this example, we download a vehicle saved in our Github repository.
```
cristianacarpinteiro, line 25: you need a object, maybe? Since you can use other objects besides vehicles.

inductiva/molecules/protein_solvation/README.md
```diff
@@ -31,15 +31,20 @@ Moreover, the hardware and interaction are configured with the usual general par
 
 ### Example
 
-To test this scenario we have available one example of a protein PDB file - [download it here](inductiva/resources/alanine.pdb).
+To test this scenario you can download a protein PDB file from the RCSB database directly from Inductiva package. Below, we download the insulin protein with the PDB ID "1ZNI".
```
cristianacarpinteiro, line 34: download a protein PDB file from the RCSB database using the inductiva package.

inductiva/molecules/utils.py
```diff
@@ -42,3 +44,19 @@ def align_trajectory_to_average(universe, trajectory_output_path):
...
+def download_pdb_from_rcsb(pdb_id: str, save_dir: Optional[str] = None) -> str:
+    """Download a PDB file from the RCSB database according 
+    to its pdb ID.
+    Args:
+        pdb_id: The PDB identifier of the molecule to download.
+    Returns:
+        The path to the downloaded PDB file.
+    """
+    file_path = f"{pdb_id}.pdb"
+    api_url = f"https://files.rcsb.org/download/{file_path}"
```
cristianacarpinteiro, line 58: maybe call it rscb_url, because api maybe confused with our API.

cristianacarpinteiro (approved):
Thanks for this, @IvanPombo  :)

inductiva/molecules/utils.py
IvanPombo: Indeed, forget to change that. Thanks!

inductiva/fluids/wind_tunnel/README.md
```diff
@@ -31,11 +31,11 @@ import inductiva
 
 inductiva.api_key = "YOUR_API_KEY"
 
-# Url to a test vehicle in Inductiva Github repository
-vehicle_url = "https://raw.githubusercontent.com/inductiva/inductiva/main" \
+# Url to a test object in Inductiva Github repository
+object_url = "https://raw.githubusercontent.com/inductiva/inductiva/main" \
```
hpenedones, line 35: Ok, then we can see if to make the path shorter we host it on our website or not... 
It could become "https://inductiva.ai/resources/vehicle.obj" for example...
Another possibility would be to use an url shortener service, but it becomes a bit more obfuscated, so let's avoid that for now.

inductiva/fluids/wind_tunnel/README.md
```diff
@@ -22,22 +22,29 @@ Launching a simulation returns a task object, which can be used to verify the st
...
 ```python
 import inductiva
 
+# Url to a test object in Inductiva Github repository
+object_url = "https://raw.githubusercontent.com/inductiva/inductiva/main" \
+              "/resources/test_vehicle.obj"
+test_object_path = inductiva.utils.files.download_from_url(
+    url=vehicle_url, local_file_path="test_object.obj")
```
hpenedones, line 38: In this example, we can assume that we will preserve the filename, and that we don't need to explicitly use keyword arguments, just use positional, ie:

`vehicle_filepath = inductiva.utils.files.download_from_url(vehicle_url)`

inductiva/fluids/wind_tunnel/README.md
```diff
@@ -22,22 +22,29 @@ Launching a simulation returns a task object, which can be used to verify the st
...
+
 # Initialize the scenario
 scenario = inductiva.fluids.WindTunnel(
     flow_velocity=[30, 0, 0],
     domain={"x": [-5, 15], "y": [-5, 5], "z": [0, 8]})
 
 # Run a simulation
 task = scenario.simulate(
-    object_path="test_vehicle.obj",
+    object_path=test_object_path,
```
hpenedones, line 47: then use vehicle_filepath

inductiva/fluids/wind_tunnel/README.md
```diff
@@ -22,22 +22,29 @@ Launching a simulation returns a task object, which can be used to verify the st
...
 
 ```python
 import inductiva
 
+# Url to a test object in Inductiva Github repository
+object_url = "https://raw.githubusercontent.com/inductiva/inductiva/main" \
+              "/resources/test_vehicle.obj"
+test_object_path = inductiva.utils.files.download_from_url(
```
hpenedones, line 37: Still thinking if it makes sense to have the method under utils.files, or just utils... will have a look...

hpenedones (changes requested):

inductiva/fluids/wind_tunnel/README.md
```diff
@@ -22,22 +22,29 @@ Launching a simulation returns a task object, which can be used to verify the st
 
 ### Example:
 
-To run this example download a vehicle geometry in OBJ format from [here](/resources/geometry/test_vehicle.obj) and
-do not forget to insert your API Key (get one by filling this [form](https://docs.google.com/forms/d/e/1FAIpQLSflytIIwzaBE_ZzoRloVm3uTo1OQCH6Cqhw3bhFVnC61s7Wmw/viewform?usp=sf_link)).
+To run this example you need an object. You can fetch any object you like directly with Inductiva package. On this example, we download a test object saved in our Github repository.
```
hpenedones, line 25: Don't make the statement so generic (the user can't use "any object") we are talking about a specific file format (.obj) only -- and perhaps we should link to a reference source about this file format. In addition to that are there some constraints on the granularity of meshes, or the fact that perhaps obj files can have extra info that we won't be able to parse?

inductiva/fluids/wind_tunnel/README.md
IvanPombo: Alright! Will add that!

inductiva/utils/files.py
```diff
@@ -100,3 +104,34 @@ def get_alphanum_key(file):
...
+
+def download_from_url(url: str,
+                      local_file_path: str,
+                      save_dir: Optional[str] = None) -> str:
+    """Download a file from an URL.
+
+    Args:
+        url: The URL to download the file from.
+        local_file_path: Name attributed to the file to download.
+        save_dir: The directory to save the file to.
```
hpenedones, line 117: The docstring does not explain what happens if the default (None) is provided. I am assuming it saves to the current working directory?

hpenedones: Also, this is the only extra argument, compared to the urlretrieve underlying method, so it is the only reason why this method was created. One needs to say that if the save_dir does not exist, this method will try to create it. It's also a bit unclear what happens if it is not possible to create that save_dir, I think it will throw an exception, right?

It's a little strange that we are catching an exception if there is an urlib download problem, but we are throwing the one related to creating subfolders... No?

inductiva/molecules/utils.py
```diff
@@ -42,3 +44,19 @@ def align_trajectory_to_average(universe, trajectory_output_path):
                     select="protein and name CA",
                     filename=trajectory_output_path,
                     in_memory=False).run()
+
+
+def download_pdb_from_rcsb(pdb_id: str, save_dir: Optional[str] = None) -> str:
+    """Download a PDB file from the RCSB database according 
+    to its pdb ID.
+    Args:
+        pdb_id: The PDB identifier of the molecule to download.
```
hpenedones, line 53: What happens if I pass a non-existing pdb_id?

inductiva/fluids/wind_tunnel/README.md
IvanPombo: I mentioned that above - but will clarify it here as well. Thanks! There are no constraints on the granularity... In principle, we can parse all for the simulation, but the visuals may not work.

inductiva/molecules/utils.py
IvanPombo: You will receive a message saying this file does not exist.

inductiva/fluids/wind_tunnel/README.md
hpenedones: Yes, that's why I initially was against that solution that @sarmento suggested on hosting on the website, but we can reconsider giving the size of this url from github. But don't worry in this PR, leave it like this now. If we decide to change it will be a simple mechanical change to a url.

inductiva/fluids/wind_tunnel/README.md
hpenedones: thanks!

inductiva/fluids/wind_tunnel/README.md
hpenedones: It's a little verbose, but I think it's fine for now. I saw that there are several other util modules under the "utils" folder, and having this method on the files.py makes sense.

inductiva/molecules/utils.py
hpenedones: Yes, seems like it will get to:

`logging.error("Could not download file from %s", url)`

You know what happens next?

inductiva/utils/files.py
```diff
@@ -100,3 +104,34 @@ def get_alphanum_key(file):
...
+        local_file_path.parent.mkdir(parents=True)
+
+    try:
+        urllib.request.urlretrieve(url, local_file_path)
+        logging.info("File downloaded to %s", local_file_path)
+    except urllib.error.URLError as url_error:
+        logging.error("Could not download file from %s", url)
+        logging.error("%s", url_error)
+
+    return str(local_file_path.absolute())
```
hpenedones, line 137: Seems like this local path will be returned whether the file is download or not. Right? So it fails the download, it will return a local path to a non-existent file, no?

IvanPombo: Indeed! I will pass this to the try part and raise an error here.

inductiva/utils/files.py
IvanPombo: Yes, it will throw an exception if it cannot create - done with mkdir.

Given that it will throw an exception first if it does not create it, then it will be fine here.

inductiva/fluids/wind_tunnel/README.md
```diff
@@ -18,26 +18,30 @@ Moreover, the hardware and interaction are configured with the usual general par
...
 
 ```python
 import inductiva
 
+# Url to a test object in Inductiva Github repository
+object_url = "https://raw.githubusercontent.com/inductiva/inductiva/main" \
+              "/resources/test_object.obj"
+object_path = inductiva.utils.files.download_from_url(url=object_url)
```
hpenedones, line 35: Can't you avoid the "url=" ?

inductiva/molecules/protein_solvation/README.md
```diff
@@ -31,15 +31,20 @@ Moreover, the hardware and interaction are configured with the usual general par
...
+
 
 ```python
- import inductiva
+import inductiva
 
+# Download the insulin protein (ID - "1ZNI") from RCSB database
+insulin = inductiva.molecules.utils.download_pdb_from_rcsb(pdb_id="1ZNI")
```
hpenedones, line 45: insulin_pdb

(otherwise it looks like this creates some python object of a protein, but it's just a string with a local path)

hpenedones (approved):
Thanks Ivan, good round of improvements!

PR #607, opened 2023-09-15
[ProteinSolvation] Fix import

cristianacarpinteiro:
In this PR, I'm removing the google.colab from the dependencies. It was causing me problems installing it and when you are using the Google Colab the dependency is already there. 
## How did I test this? 
**In Colab:**
image
However the video is not playing in Colab, I need to go deeper on it. I will address it in another PR, since the pip install inductiva[molecules_extra] is broken and I want to fix it ASAP. 
**In Terminal:** 
image

### DIFF
--- a/inductiva/molecules/protein_solvation/post_processing.py
+++ b/inductiva/molecules/protein_solvation/post_processing.py
@@ -156,4 +156,5 @@ class ProteinSolvationOutput:
                               "only available in a python notebook.")
 
         if "google.cloud" in str(self.ipython_kernel):
+            print("here")
             gc.output.enable_custom_widget_manager()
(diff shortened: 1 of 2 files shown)


IvanPombo (approved):
The changes look good, however I don't really understand why you say that the `inductiva[molecules_extra]` is broken. Could you ellaborate?

cristianacarpinteiro:
> The changes look good, however I don't really understand why you say that the `inductiva[molecules_extra]` is broken. Could you ellaborate?

I was getting an error installing it, now I am not being able to reproduce it but it works without the dependency, so it is something we don't need to add.

IvanPombo:
> > The changes look good, however I don't really understand why you say that the `inductiva[molecules_extra]` is broken. Could you ellaborate?
> 
> I was getting an error installing it, now I am not being able to reproduce it but it works without the dependency, so it is something we don't need to add.

Alright! Just test a few more times, even though this package extra will just be for Colab.

PR #613, opened 2023-09-15
[fenicsx render] - Add render method

Joao97ribeiro:
In this PR, I've introduced the rendering feature to the FEniCSx outputs. To achieve this, I've utilized the `create_2d_field_from_xdmf` method, which I've added to the `visualization` file.

I plan to create another PR to conduct testing for rendering in the NeuroStruct repository.

@baribeiro, please take a look at how I've implemented the XDMF file reading process. It might offer valuable insights for converting the XDMF file output into a PyG dataset


Joao97ribeiro:
I'm adding a new method called `remove_time_values_from_xdmf` because the current version of PyVista lacks the ability to inspect time values in the reader (as mentioned in the [PyVista documentation](https://docs.pyvista.org/version/stable/api/readers/_autosummary/pyvista.XdmfReader.html)). In our specific use case, we only have one time instance, so by removing the lines with time values, we can ensure that it works correctly

inductiva/__init__.py
```diff
@@ -14,7 +14,7 @@
 from . import generative
 from . import molecules
 from . import resources
-# from . import structures
```
hpenedones, line 17: I think that this had been commented to make sure there are no problems with imports of external libraries. Can you please check with @luispcunha and @IvanPombo that this won't break our "import inductiva"?

inductiva/structures/post_processing/fenicsx.py
```diff
@@ -0,0 +1,149 @@
+"""Post-process FEniCSx simulation outputs."""
+
+import os
+import tempfile
+
+import pyvista as pv
+
+import inductiva
```
hpenedones, line 8: it is weird that you are inside the package inductiva but you are importing it, this seems like a circular import... No?

inductiva/structures/post_processing/fenicsx.py
```diff
@@ -0,0 +1,149 @@
+"""Post-process FEniCSx simulation outputs."""
+
+import os
+import tempfile
+
+import pyvista as pv
```
hpenedones, line 6: pyvista will have to go to an extra package dependency, right @IvanPombo @luispcunha ?

hpenedones (changes requested):

inductiva/__init__.py
IvanPombo: This is due to the gmsh not being anymore on the Client. I will check on this and you can notify me if it works on your side. We may need to follow the approach of this PR #538.

inductiva/structures/post_processing/fenicsx.py
IvanPombo: I think it is the usual practice - Google Style Guide quote at 2.2 "Do not use relative names in imports. Even if the module is in the same package, use the full package name. This helps prevent unintentionally importing a package twice."

inductiva/utils/visualization.py
```diff
@@ -512,3 +512,63 @@ def create_color_plot_movie(
         create_movie_from_frames(frames_dir=tmp_dir,
                                  movie_path=movie_path,
                                  fps=movie_fps)
+
+
+@optional_deps.needs_structures_extra_deps
```
IvanPombo, line 517: With this created just go to the setup.cfg to add the structures_extra package with the extra pyvista.

inductiva/utils/visualization.py
```diff
@@ -512,3 +512,63 @@ def create_color_plot_movie(
...
+
+    # Optionally, add a scalar bar to the visualization
+    if scalar_bar:
+        plotter.add_scalar_bar(file_name, vertical=True, interactive=True)
+
+    # Set the view to XY plane
+    plotter.view_xy()
+
+    # Set the background color to white
+    plotter.background_color = "white"
```
IvanPombo, line 564: This should be an input variable.

inductiva/utils/visualization.py
```diff
@@ -512,3 +512,63 @@ def create_color_plot_movie(
...
+
+    # Render off-screen
+    plotter.off_screen = True
```
IvanPombo, line 567: This as well.

inductiva/utils/visualization.py
```diff
@@ -512,3 +512,63 @@ def create_color_plot_movie(
...
+
+    # Initialize the PyVista plotter
+    plotter = pv.Plotter()
+
+    # Add the field data to the PyVista data object and set it as active scalars
+    data.point_data[file_name] = field_array
+    data.set_active_scalars(file_name)
+
+    # Add the data object to the plotter with customizable settings
+    plotter.add_mesh(data, show_edges=True, cmap="jet", show_scalar_bar=False)
```
IvanPombo, line 554: The cmap and show_edges should also be arguments. Provide this to the user, and you will be able to quickly change things as well :)

inductiva/structures/post_processing/fenicsx.py
luispcunha: doing `from inductiva import z` where z is a submodule would also adhere to the style guide, but we've been getting feedback to do it like this, so this is how it looks like almost everywhere now: https://github.com/inductiva/inductiva/pull/439#discussion_r1286626468

inductiva/structures/post_processing/fenicsx.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def __init__(self, sim_output_path: inductiva.types.Path):
+        """Initializes a 'FEniCSxSimulationOutput' object.
+
+        Args:
+            sim_output_path: Path to simulation output files.
+            """
+        self.sim_output_path = sim_output_path
+
+    def remove_time_values_from_xdmf(self, output_file_path: str) -> None:
```
IvanPombo, line 22: Isn't this function an utility function? It maybe useful for other future scenarios.

inductiva/structures/post_processing/fenicsx.py
```diff
@@ -0,0 +1,149 @@
...
+
+import os
+import tempfile
+
+import pyvista as pv
+
+import inductiva
+
+
+class FEniCSxSimulationOutput:
```
IvanPombo, line 11: The name of this class is too generic for FEniCSx, but actually your rendering is specific for the structures case. I think you won't be able to render all FEniCSx simulation scenarios with it, right?

inductiva/structures/post_processing/fenicsx.py
```diff
@@ -0,0 +1,149 @@
...
+        temp_dir = tempfile.mkdtemp()
+
+        # Specify the filename within the temporary directory
+        temp_file_path = os.path.join(temp_dir, "temp_results.xdmf")
+
+        # Remove time values from the XDMF file
+        self.remove_time_values_from_xdmf(self.sim_output_path)
+
+        # Read simulation output data
+        reader = pv.get_reader(temp_file_path)
```
IvanPombo, line 78: Imagine you have another rendering function or you call this function again, will you do this all the time? The information we are removing is it super important for the user to have, or can we just deliver the sim_output_path already without it?

inductiva/structures/post_processing/fenicsx.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # Clean up temporary files and directory
+        os.remove(temp_file_path)
+        os.rmdir(temp_dir)
+
+        if field_names is None:
+
+            # Default field names to render
+            field_names = [
+                "displacement_x", "displacement_y", "displacement_z",
```
IvanPombo, line 89: Can you highlight if the stress, strain and von_mises data is important for the user to manipulate and then render? Or they will just want to render and see the results?

inductiva/structures/post_processing/fenicsx.py
```diff
@@ -0,0 +1,149 @@
...
+                "stress_yz", "stress_zx", "stress_zy", "stress_zz", "strain_xx",
+                "strain_xy", "strain_xz", "strain_yx", "strain_yy", "strain_yz",
+                "strain_zx", "strain_zy", "strain_zz", "von_mises"
+            ]
+
+        # Get the list of available keys (field names) in the PyVista dataset
+        keys = list(data.keys())
+
+        # Iterate through the specified field names to render
+        for field_name in field_names:
```
IvanPombo, line 100: If I understand this, it will just create 21 graphs when field_names is None. Don't you think this is super cumbersome for the user (you)?

inductiva/__init__.py
Joao97ribeiro: @hpenedones, I uncommented it because I thought that this problem had been resolved
@IvanPombo, thanks. I will now verify if everything is functioning correctly

inductiva/structures/post_processing/fenicsx.py
Joao97ribeiro: @hpenedones, I'm following Luis's comment in the last PR (check out @luispcunha's comment with the link). Do you think it's better to approach it differently?

inductiva/structures/post_processing/fenicsx.py
Joao97ribeiro: Do I need to include this in the `utils/optional_deps.py` file, right? I've already added: 

```
needs_structures_extra_deps = functools.partial(
    _needs_optional_deps,
    ["pyvista"],
    _missing_deps_msg("structures"),
)
```

Since I already have 'pyvista' in optional_deps, can I use it here?

inductiva/utils/visualization.py
Joao97ribeiro: like this, right?

```
needs_structures_extra_deps = functools.partial(
    _needs_optional_deps,
    ["pyvista"],
    _missing_deps_msg("structures"),
)
```

inductiva/structures/post_processing/fenicsx.py
Joao97ribeiro: Yes, you're correct, I'll move it to the utils of the structure scenario, thanks!

inductiva/structures/post_processing/fenicsx.py
Joao97ribeiro: If it's correct, it depends on what you're saving, whether it's displacements, stresses, strains, I'll renaming the class, thanks Ivan!

inductiva/structures/post_processing/fenicsx.py
Joao97ribeiro: I understand your point, the idea here was to provide all the outputs as pictures in case people don't want to specify what they want to see. If you know what you want to see, you only need to specify the name. Do you think there's a better alternative?

inductiva/structures/post_processing/fenicsx.py
Joao97ribeiro: It could be important, depending on what people want to study. Right now, our focus is primarily on rendering the outputs

inductiva/structures/post_processing/fenicsx.py
Joao97ribeiro: Yes, it's not crucial. With this function, we're simply removing lines that contain the phrase 'time values,' so we're not losing any data. The only downside is having to repeat this process

inductiva/__init__.py
IvanPombo: I believe it may be working for you, but not for other users. So I think we should try to pass the gmsh dependencies to the backend - specially because this is a scenario.

inductiva/structures/post_processing/fenicsx.py
IvanPombo: But If I do another scenario and I provide the displacements, stresses and strains. Will I be able to use this?

inductiva/structures/post_processing/fenicsx.py
IvanPombo: Then just remove it once (even in the backend if you want) and do not repeat it - it can cause some unneeded crashes.

inductiva/structures/post_processing/fenicsx.py
IvanPombo: Yes, but with pyvista a little widget appears, so I need to just down 21 images before I move with my life. That is more painful, that selecting what I want to see. If you say to the user that he can see the stress, displacements, etc and the possible planes, then he can configure it himself. If you want to retrieve a list, that consider passing these args as a list.

inductiva/__init__.py
Joao97ribeiro: Should I uncomment it now and then pass the Gmsh to the backend since this will be my next task?

IvanPombo: Just uncomment, only if you are not using the gmsh anymore. Otherwise, and if you want to be able to merge keep it commented while you work. We will be deploying a new version of the package today and we do not want that to break it apart. Thanks :)

Joao97ribeiro:
Thank you all for the suggestions; I've implemented all of them. I'm currently experiencing some pytest errors, and I believe they may be related to the Gmsh issue we discussed earlier, correct? Could you please provide your suggestions on how to resolve this bug? Thanks

IvanPombo:
> Thank you all for the suggestions; I've implemented all of them. I'm currently experiencing some pytest errors, and I believe they may be related to the Gmsh issue we discussed earlier, correct? Could you please provide your suggestions on how to resolve this bug? Thanks

Remove the structures and the output_class = inductiva.structures....  this is crashing the pytest because the output_consts is being called from the package __init__, and then it checks every required package inside the structures module. Maybe adding the optional_deps can also insure that it passes the test when gmsh does not exist. But given that you will pass to the backend I would remove it from now.

Joao97ribeiro:
Thanks, @IvanPombo! With your suggestion, the pytest issue has disappeared. @luispcunha, could you please check if the YAPF error is related to the api_client?

IvanPombo (commented):
thanks for the updates @Joao97ribeiro! Everything is looking much more clean! I still have some notes that I would like for you to address in order to provide more readability. After them, it will be ready to merge!

inductiva/structures/post_processing/fenicsx_post_processing.py
```diff
@@ -0,0 +1,156 @@
...
+        # Define a XDMF temporary file path
+        temp_file_path = os.path.join(self.sim_output_path,
+                                      TEMP_SOLVER_OUTPUT_FILENAME)
+
+        # Remove time values from the XDMF file
+        structures.utils.remove_time_values_from_xdmf(file_path, temp_file_path)
+
+        # Read simulation output data
+        reader = pv.get_reader(temp_file_path)
+        py_multiblock = reader.read()
```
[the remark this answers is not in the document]
Joao97ribeiro: Sorry, I already changed it to pv_multiblock. It is by pyvista.MultiBlock: https://docs.pyvista.org/version/stable/api/core/_autosummary/pyvista.MultiBlock.html#pyvista-multiblock

inductiva/utils/visualization.py
```diff
@@ -512,3 +512,128 @@ def create_color_plot_movie(
         create_movie_from_frames(frames_dir=tmp_dir,
                                  movie_path=movie_path,
                                  fps=movie_fps)
+
+
+@optional_deps.needs_structures_extra_deps
+def create_2d_field_from_pv_dataset(
+        pv_dataset: "pv.UnstructuredGrid",
```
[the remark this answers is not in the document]
Joao97ribeiro: Good point! I've checked the `set_active_scalars function`, and it works with various types of [pyvista.DataSet](https://docs.pyvista.org/version/stable/api/core/_autosummary/pyvista.DataSet.html#pyvista.DataSet)

https://docs.pyvista.org/version/stable/api/core/_autosummary/pyvista.DataSet.set_active_scalars.html#pyvista-dataset-set-active-scalars

inductiva/structures/post_processing/fenicsx_post_processing.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Define the XDMF file path for the solver output file
+        file_path = os.path.join(self.sim_output_path, SOLVER_OUTPUT_FILENAME)
+
+        # Define a XDMF temporary file path
+        temp_file_path = os.path.join(self.sim_output_path,
+                                      TEMP_SOLVER_OUTPUT_FILENAME)
+
+        # Remove time values from the XDMF file
+        structures.utils.remove_time_values_from_xdmf(file_path, temp_file_path)
```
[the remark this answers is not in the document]
Joao97ribeiro: Why do you mention that it will be used and reused? People can provide the fields as a list, so this function is typically used only once, right?

inductiva/structures/post_processing/fenicsx_post_processing.py
```diff
@@ -0,0 +1,156 @@
...
+               colormap: str = "jet",
+               off_screen: bool = True,
+               scalar_bar: bool = True,
+               background_color: str = "white",
+               transparent_background: bool = True) -> None:
+        """Render the simulation fields as images.
+
+        Args:
+            field_names (list, optional): List of field names to render. Default
+              is None.
```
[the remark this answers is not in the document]
Joao97ribeiro: I know, but it's the best way to avoid the pylint error: W0102: Dangerous default value [] as argument (dangerous-default-value). If we use ["von_mises"], we will encounter this error

Joao97ribeiro:
Thanks again @IvanPombo for the suggestions! Please let me know if there are more things that I can improve before the merge

inductiva/structures/post_processing/fenicsx_post_processing.py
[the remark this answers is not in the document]
IvanPombo: Every time you call render you will re-do this again. But I can also want to render things in sequence.

inductiva/structures/post_processing/fenicsx_post_processing.py
```diff
@@ -0,0 +1,168 @@
...
+        pv_multiblock = reader.read()
+
+        # Get the list of available keys (field names) in the PyVista dataset
+        keys = list(pv_multiblock.keys())
+
+        # Select the PyVista dataset representing the mesh
+        pv_dataset = pv_multiblock["mesh"]
+
+        # Define a list of valid field names
+        valid_field_names = [
```
IvanPombo, line 84: Just set this as a global variable...

IvanPombo (approved):
LGTM! Thanks for the good work!

inductiva/structures/post_processing/fenicsx_post_processing.py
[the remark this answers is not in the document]
Joao97ribeiro: I understand your point. With this suggestion, it's impossible to delete this file after rendering, right? In this way, we can delete this file, and I think it's important delete it because it's the same as with the results, but it doesn't have the time value information (which in this problem doesn't mean anything because it's just one instance). What do you think?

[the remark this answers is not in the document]
IvanPombo: But you could change the existing file instead of creating a copy. If you just lose one line, that is not important from the previous discussion, then do it on the initialization of the object.

inductiva/structures/post_processing/fenicsx_post_processing.py
```diff
@@ -0,0 +1,156 @@
...
+                    "yx": 3,
+                    "yy": 4,
+                    "yz": 5,
+                    "zx": 6,
+                    "zy": 7,
+                    "zz": 8
+                }[subfield]
+
+                # Extract the specific subfield array from the dataset
+                field_array = py_multiblock[keys[field_id]].get_array(
```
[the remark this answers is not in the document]
Joao97ribeiro: The '0' represents the index of the array.  When we select a field, we have only one array -> index 0

Joao97ribeiro (commented):
I will now proceed with the merge and focus on implementing mesh generation on the backend side

Joao97ribeiro:
@hpenedones, I believe I require your approval to proceed with the merge. Whenever you have some time, please take a look

hpenedones (approved):
Thanks João!

PR #622, opened 2023-09-18
Manage Output Classes and Default files at Task Level

IvanPombo:
This PR allows the tasks to manage the output classes and default files, even when they are initialised just from the task ID. To do so, we need to configure for all scenarios their own `api_method_name`, which needs to match the backend. Then, we match this method name with the respective output class and default files when one does the `task.get_output()` - not in the `scenario.simulate()` anymore. In this way, by fetching a task like `task = inductiva.tasks.Task(task_id)` one will have the same access to the output class as when finishing a simulation.

Currently under testing!


inductiva/molecules/protein_solvation/protein_solvation.py
```diff
@@ -80,7 +67,7 @@ def simulate(
             n_steps_min: Number of steps for energy minimization.
             run_async: Whether to run the simulation asynchronously.
         """
-        simulator.override_api_method_prefix("proteinsolvation")
+        simulator.override_api_method_prefix("protein_solvation")
```
cristianacarpinteiro, line 70: This should be the same as in the output_consts, right?

inductiva/tasks/output_consts.py
```diff
@@ -0,0 +1,43 @@
+"""Constants for connecting tasks to their simulation output classes."""
```
cristianacarpinteiro, line 1: Suggestion: Definition of constants and configurations for connecting tasks to their simulation output classes.

cristianacarpinteiro (approved):
Thank you for this :) The code looks good! Be sure to test it for more than 1 scenario!

inductiva/tasks/output_consts.py
```diff
@@ -0,0 +1,43 @@
...
+import inductiva
+
+SCENARIO_METHODS_NAMES = [
```
hpenedones, line 4: Can you think of a way of avoiding these names repeated 3 times (this list and the keys for the two dictionaries)? DRY principle.

hpenedones (commented):
Thanks for this Ivan!!

inductiva/molecules/protein_solvation/protein_solvation.py
IvanPombo: It is, yes!

cristianacarpinteiro: got it 👍

inductiva/fluids/dam_break/dam_break.py
```diff
@@ -61,16 +61,20 @@ def simulate(
...
 
-        task = super().simulate(simulator=simulator,
-                                machine_group=machine_group,
-                                device=device,
-                                particle_radius=particle_radius,
-                                simulation_time=simulation_time,
-                                run_async=run_async)
-
-        task.set_output_class(fluids.SPHSimulationOutput)
+        task = super(fluids.FluidBlock, self).simulate(
```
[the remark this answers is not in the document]
IvanPombo: This a weird but cool feature, that let's you inherit the `simulate` of the grandparent instead of the parent. That is why `fluids.FluidBlock` is set first -> it says that you will do the super for the FluidBlock, which is `Scenario.simulate()`.

The problem this addresses is that the FluidBlock would override the api_method_name for the DamBreak, hence not allowing us to identify the DamBreak at all. Since I still want to use the file configuration functions, I thought about this method.

inductiva/tasks/task.py
```diff
@@ -216,19 +206,32 @@ def get_output(
         """
         self.wait()
 
-        # Check if default files are set by the output class
+        # Check if already have the info and the task is in a terminal status.
+        if self._info is not None and self._status in _TASK_TERMINAL_STATUSES:
```
hpenedones, line 210: Why do you need this if statement, get_info() is already caching results inside with a similar condition, no? Can't you just always call get_info() here?

IvanPombo: Indeed, sorry for this!

inductiva/tasks/task.py
```diff
@@ -216,19 +206,28 @@ def get_output(
...
 
         output_dir = self.download_outputs(
             filenames=filenames,
             output_dir=output_dir,
             uncompress=uncompress,
             rm_downloaded_zip_archive=rm_downloaded_zip_archive)
 
-        if self._output_class:
-            return self._output_class(output_dir)
+        if output_class is not None:
```
hpenedones, line 229: Does this mean that sometimes this method returns a string with a path and sometimes it returns a python object (some Scenario Output?). Is it consistent with the docstring?

PR #629, opened 2023-09-19
[CoastalArea] Fix bug for non-uniform grids plotting

david451lima:
Non uniform grids were not plotting since the x and y resolutions were set to None as default. This pr adds a resize option that resizes an already uniform grid. The default value is still to take the bathymetry resolution.

### DIFF
--- a/inductiva/coastal/bathymetry.py
+++ b/inductiva/coastal/bathymetry.py
@@ -328,8 +328,9 @@ class Bathymetry:
         cmap: Optional[str] = None,
         clim: Optional[Tuple[float]] = None,
         path: Optional[str] = None,
-        x_resolution: float = None,
-        y_resolution: float = None,
+        x_resolution: float = 10,
+        y_resolution: float = 10,
+        resize: bool = False,
         threshold_distance: float = 20,
     ) -> Union["matplotlib.axes.Axes", None]:
         """Plots the bathymetry.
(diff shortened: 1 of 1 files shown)


inductiva/coastal/bathymetry.py
```diff
@@ -328,8 +328,9 @@ def plot(
         cmap: Optional[str] = None,
         clim: Optional[Tuple[float]] = None,
         path: Optional[str] = None,
-        x_resolution: float = None,
-        y_resolution: float = None,
+        x_resolution: float = 10,
+        y_resolution: float = 10,
+        resize: bool = False,
```
IvanPombo, line 333: This seems very cumbersome for the user, because If I create a non-uniform grid I would be expecting to plot it as a non-uniform grid.

However, I know that this may take a few more iterations to reach there... WDYT?

david451lima: Agree with you, albeit the situation is a bit more tricky. There are non-uniform grids that are non-uniform due to the data being conveyed encoded in the latitude-longitude. This is non-uniform, but the user does not want to plot with the earth's curvature afaik. So (for now) I think we can leave it like this.

IvanPombo: Cool! That is different... it's a projection that needs to happen to a plane.

IvanPombo (approved):
Let's merge this, but open an issue to address the concerns exposed here.

PR #631, opened 2023-09-19
Remove (almost all) hardware configurations from simulators

IvanPombo:
This PR removes the hardware configurations from the arguments of the simulators and scenarios. Namely, it removes the device selection between CPU and GPU from the SPH scenarios/simulators and removes the `n_cores` from the swash and xbeach.

By default, now the simulations in the best hardware available through introspection in the machine they are running. If users want to increase their capacity they should update their `machine_group` configuration to use more cores or different devices. 

Currently, GPU is not yet available. So the backend executer will just default to use CPU.

Furthermore, OpenFOAM is not so clear, since we need to change one file with the n_cores which is currently done on the Client through jinja. The easy way would be to do jinja on the backend as well, but then things can be messy for low-level implementation. I am going to address this in an issue, but for the next steps I would prefer this are just left unchanged and keep the `n_cores` parameter.


hpenedones:
Thanks Ivan, for the WindTunnel can you set the default value of the num cores argument to be the best for the current default pool (not sure if it's 2 or 4...). Now it's set to 1...

hpenedones (approved):

PR #636, opened 2023-09-19
[CoastalArea] Add logging info about divergent simulations

david451lima:
Some simulations, in particular for low resolutions, tend to diverge (water level rises to unphysical values). This behaviour can de explained by numerical errors.
For a given resolution, the numerical instabilities should not cause issues, since the time step of the simulation is not fixed but instead controlled by a criterion that
aims to keep the Courant Number (a relation between the time step, velocity of the water and grid resolution) bellow an unstable treshold (Cr=1). The strictness of the time step controll is set in the
input file. At the moment, the timestep is changed if Cr>0.5, which should keep the instabilities in check. Nonetheless, they keep happening. The only thing that diminishes this
issue is increasing the resolution. Since there is not a clear way to controll this atm, I add a logging info that warns the user if the simulation exploded.

### DIFF
--- a/inductiva/coastal/output.py
+++ b/inductiva/coastal/output.py
@@ -205,7 +205,7 @@ class CoastalAreaOutput:
     @optional_deps.needs_coastal_extra_deps
     def check_stability(self):
 
-        time_list, data_list = read_swash_grid_quantity_file(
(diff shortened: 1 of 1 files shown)


IvanPombo (approved):
This is a nice add-on, however it may not be the complete solution!

PR #639, opened 2023-09-19
[ProteinSolvation] Add maxwarn flag to grompp command

cristianacarpinteiro:
The addition of the maxwarn flag to the grompp command prevents experiments to fail for this common errors: 
```
 WARNING 1 [file energy_minimization.mdp]:
  The largest distance between excluded atoms is 1.208 nm, which is larger
  than the cut-off distance. This will lead to missing long-range
  corrections in the forces and energies. If you expect that minimization
  will bring such distances within the cut-off, you can ignore this warning.
```
According to [this](https://gromacs.bioexcel.eu/t/the-largest-distance-between-excluded-atoms-is-1-725-nm-which-is-larger-than-the-cut-off-distance/4153/2) the user should take a look to check if the protein structure is correct, however, the problem can be overriden by setting the `maxwarn` flag to 1.
I also remove the ns_type variable, since it was already the default value.

### DIFF
--- a/inductiva/molecules/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/protein_solvation/protein_solvation.py
@@ -3,6 +3,7 @@
 from functools import singledispatchmethod
 from typing import Optional, Literal
 import os
+import io
 import shutil
 
 from inductiva import tasks, resources, simulators, scenarios, utils
@@ -10,7 +11,7 @@ from inductiva import tasks, resources, simulators, scenarios, utils
 SCENARIO_TEMPLATE_DIR = os.path.join(utils.templates.TEMPLATES_PATH,
                                      "protein_solvation")
 GROMACS_TEMPLATE_INPUT_DIR = "gromacs"
-COMMANDS_TEMPLATE_FILE_NAME = "commands.json"
+COMMANDS_TEMPLATE_FILE_NAME = "commands.json.jinja"
 
 
 class ProteinSolvation(scenarios.Scenario):
@@ -101,7 +102,14 @@ class ProteinSolvation(scenarios.Scenario):
                                               GROMACS_TEMPLATE_INPUT_DIR,
                                               COMMANDS_TEMPLATE_FILE_NAME)
 
-        commands = self.read_commands_from_file(commands_template_path)
+        inmemory_file = io.StringIO()
+        utils.templates.replace_params(
+            template_path=commands_template_path,
+            params={"max_warn": self.ignore_warnings},
+            output_file=inmemory_file,
+        )
+        commands = self.read_commands_from_file(inmemory_file)
+
         return commands
 
     @singledispatchmethod
@@ -124,12 +132,10 @@ def _(self, simulator: simulators.GROMACS, input_dir):  # pylint: disable=unused
     utils.templates.batch_replace_params(
         templates_dir=input_dir,
         template_filenames=[
-            "commands.json.jinja",
             "simulation.mdp.jinja",
             "energy_minimization.mdp.jinja",
         ],
         params={
-            "max_warn": self.ignore_warnings,
             "integrator": self.integrator,
             "nsteps": self.nsteps,
             "ref_temp": self.temperature,
--- a/inductiva/molecules/protein_solvation/protein_solvation.py
+++ b/inductiva/molecules/protein_solvation/protein_solvation.py
@@ -45,7 +45,8 @@ class ProteinSolvation(scenarios.Scenario):
             simulation_time_ns: float = 10,  # ns
             output_timestep_ps: float = 1,  # ps
             integrator: Literal["md", "sd", "bd"] = "md",
-            n_steps_min: int = 5000) -> tasks.Task:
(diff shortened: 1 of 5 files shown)


david451lima (approved):
LGTM

IvanPombo:
@cristianacarpinteiro according to this (and following the structures of other simulators) it means that if more than one warning appears, then the simulation crashes. Is this correct?

Moreover, this error won't be passed to the user until the end of the simulation, so we won't be able to stop it if he did not want this happening. So the simulation can work, but users may still want to kill it for these warnings. No? If so, we need to start thinking about how to send this to the user.

cristianacarpinteiro:
> @cristianacarpinteiro according to this (and following the structures of other simulators) it means that if more than one warning appears, then the simulation crashes. Is this correct?
> 
> Moreover, this error won't be passed to the user until the end of the simulation, so we won't be able to stop it if he did not want this happening. So the simulation can work, but users may still want to kill it for these warnings. No? If so, we need to start thinking about how to send this to the user.

Yes, you are correct! These warnings appear in a lot of RCSB files and are very specific to each protein to solve in a systematic way. Ignoring this warnings will produce a simulation with some minor inaccuracies according to the documentation, but I think it should be up to the user if he wants to ignore the warnings or not. What is your suggestion on how to pass this to the user?

IvanPombo:
> Yes, you are correct! These warnings appear in a lot of RCSB files and are very specific to each protein to solve in a systematic way. Ignoring this warnings will produce a simulation with some minor inaccuracies according to the documentation, but I think it should be up to the user if he wants to ignore the warnings or not. What is your suggestion on how to pass this to the user?

I do not have a suggestion as of yet, but at least for warnings there should be a certain stream from the backend so that users can immediately address them if they want. Otherwise, they consume computation and only at the end they see their simulation was useless.

cristianacarpinteiro:
I think the best way of approaching this as of now is to pass a ignore_warnings in the simulate method. If the user sets it to True we add the maxwarn flag and if it sets to False the simulation will fail and then the user can see the warnings on the stdout. Streaming the warnings to the user directly seems a too complicated solution to implement quickly.

david451lima:
> simulate method. If the user sets it to True we add the maxwarn flag and if it sets to False the simulation will fail and then the user can see the warnings on the stdout. Streaming the warnings to the user directly seems a too complicated solutio

Seems a reasonable solution. The user can check immediatly in the stdout if the simulation crashed due to warnings and pass the argument accordingly

cristianacarpinteiro:
> > I think the best way of approaching this as of now is to pass a ignore_warnings in the simulate method. If the user sets it to True we add the maxwarn flag and if it sets to False the simulation will fail and then the user can see the warnings on the stdout. Streaming the warnings to the user directly seems a too complicated solution to implement quickly.
> 
> I didn't mention to do it quickly :) Just to start thinking about it!
> 
> Okay, that seems like a reasonable solution! However, with `-maxwarn 1` this will just ignore the first warning and crash on the second. What do you think you should provide? Ignore all? A few? Let the user decide?

I will put it as a value the user can change instead of a bool.

cristianacarpinteiro:
@david451lima @IvanPombo I am re-requesting your review because I changed the way I was doing this to address our concerns.

inductiva/molecules/protein_solvation/protein_solvation.py
```diff
@@ -45,7 +46,8 @@ def simulate(
             simulation_time_ns: float = 10,  # ns
             output_timestep_ps: float = 1,  # ps
             integrator: Literal["md", "sd", "bd"] = "md",
-            n_steps_min: int = 5000) -> tasks.Task:
+            n_steps_min: int = 5000,
+            ingore_warnings: bool = False) -> tasks.Task:
```
IvanPombo, line 50: Typo

inductiva/molecules/protein_solvation/protein_solvation.py
```diff
@@ -78,6 +82,10 @@ def simulate(
         # of the simulation (2 fs)
         self.integrator = integrator
         self.n_steps_min = n_steps_min
+        self.ignore_warnings = 0
+        if ingore_warnings:
```
IvanPombo, line 86: typo

inductiva/molecules/protein_solvation/protein_solvation.py
```diff
@@ -78,6 +82,10 @@ def simulate(
...
+            self.ignore_warnings = -1
```
IvanPombo, line 87: Why -1?

IvanPombo: Is to catch all?

IvanPombo (approved):
Minor comments. LGTM!

inductiva/molecules/protein_solvation/protein_solvation.py
cristianacarpinteiro: yes, all of them are ignored.

IvanPombo: Cool, thanks!

PR #642, opened 2023-09-20
Allow downloading example from ZIP file

luispcunha:
This is a fix in order to allow downloading examples that are zipped in the original download website. This will check if the file is a zip after downloading and then unzip the file. This assumes that the ZIP will contain only a single file.

### DIFF
--- a/inductiva/coastal/output.py
+++ b/inductiva/coastal/output.py
@@ -223,8 +223,12 @@ class CoastalAreaOutput:
     @optional_deps.needs_coastal_extra_deps
     def render(
         self,
-        quantity: Literal["water_level", "velocity_x", "velocity_y",
-                          "velocity_magnitude",] = "water_level",
+        quantity: Literal[
+            "water_level",
+            "velocity_x",
+            "velocity_y",
+            "velocity_magnitude",
+        ] = "water_level",
         movie_path: Path = "movie.mp4",
         fps: int = 5,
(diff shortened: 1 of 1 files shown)


david451lima (approved):
LGTM! This will be very helpful for bathymetries that come from EU websites

IvanPombo (approved):

hpenedones:
This seems a super specific case, I don't see why we would rush to write code to cover this. @david451lima : why didn't you reply to my question on the other pull request explaining in more detail what are those limitations you identified (e.g. why can't we just store a decompressed file of a bathymetry somewhere?)

hpenedones (commented):
A bit puzzled on why we would rush to write code for this.

david451lima:
> This seems a super specific case, I don't see why we would rush to write code to cover this. @david451lima : why didn't you reply to my question on the other pull request explaining in more detail what are those limitations you identified (e.g. why can't we just store a decompressed file of a bathymetry somewhere?)

Both by Sarmento suggestion to not store the file since it belongs to an european agency (it is not our data) and because it's a large file (over 100Mb)

hpenedones:
Ok, thanks for the additional context. Remember that we have remote workers and we collaborate via GitHub, so if a reviewer asks clarifying questions, please don't ignore and do things based on office discussions, leaving reviewers in the dark.

Issue #633, opened 2023-09-19
[Coastal Area]: Task Failed is not handed gracefully by the client


sarmento:
Task failed -- I don't know the reason -- but the client does not handle that situation gracefully and keeps going.

```
Generating Random Bathymetry
INFO:absl:Generating random array with the Diamond-Square Algorithm...
Plotting
Creating Scenario
Calling Simulate
INFO:absl:Task ID: 1695137136368486711
INFO:absl:Creating input archive...
INFO:absl:Input archive created.
INFO:absl:Input archive size: 1.63 MiB
INFO:absl:Uploading input...
INFO:absl:Input archive uploaded.
INFO:absl:Task request submitted.
INFO:absl:Waiting for resources...
INFO:absl:Task failed.
INFO:absl:Download the task output and check the 'stdout.txt' and 'stderr.txt' files for more information.
INFO:absl:Task failed.
INFO:absl:Download the task output and check the 'stdout.txt' and 'stderr.txt' files for more information.
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 916/916 [00:00<00:00, 286kiB/s]
Traceback (most recent call last):
  File "test.py", line 30, in <module>
    output.render()
  File "/usr/local/anaconda3/lib/python3.8/site-packages/inductiva/utils/optional_deps.py", line 48, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/anaconda3/lib/python3.8/site-packages/inductiva/coastal/output.py", line 236, in render
    x_array, y_array = read_swash_grid_positions_file(
  File "/usr/local/anaconda3/lib/python3.8/site-packages/inductiva/utils/optional_deps.py", line 48, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/anaconda3/lib/python3.8/site-packages/inductiva/coastal/output.py", line 69, in read_swash_grid_positions_file
    grid_positions_dict = read_swash_output_file(file_path)
  File "/usr/local/anaconda3/lib/python3.8/site-packages/inductiva/utils/optional_deps.py", line 48, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/anaconda3/lib/python3.8/site-packages/inductiva/coastal/output.py", line 59, in read_swash_output_file
    raise ValueError(f"Error reading file '{file_path}'.") from None
ValueError: Error reading file '/Users/lsarmento/swash/inductiva_output/1695137136368486711/grid_positions.mat'.
[...]

PR #643, opened 2023-09-20
Fix failed tasks output configuration

IvanPombo:
This PR addresses the concerns in issue #633 to handle the tasks that fail more gracefully. The goal is that the stderr.txt and stdout.txt are downloaded immediately for the user when he does task.get_output() and that afterwards we are not initializing a class that won't work. Thus the output will be a path.

To address all of these a couple of loggings were added to clarify the user.

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -244,6 +244,7 @@ class Task:
             output_class, filenames = self._get_output_config(all_files)
         else:
             output_class = None
+            logging.info("Downloading the 'stdout.txt' and 'stderr.txt' files.")
             filenames = ["stdout.txt", "stderr.txt"]
 
         output_dir = self.download_outputs(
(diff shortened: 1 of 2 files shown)


inductiva/tasks/task.py
```diff
@@ -177,6 +178,36 @@ def kill(self) -> None:
         """
         self._api.kill_task(path_params=self._get_path_params())
 
+    def _get_output_config(self, all_files: bool = False):
+        """Get configuration of the output wrt the task method name.
```
begaiym-k, line 182: "wrt"?

inductiva/tasks/task.py
```diff
@@ -177,6 +178,36 @@ def kill(self) -> None:
...
+        
+        Args:
+            all_files: Whether to download all the files in the output.
+
+        Returns:
+            The output class and default files for the task related
```
begaiym-k, line 188: I don't quite understand what it returns

inductiva/tasks/task.py
```diff
@@ -156,7 +156,8 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
                     logging.info("Task failed.")
                     logging.info("Download the task output and check the "
```
begaiym-k, line 157: Does this log need to change as well, since those files are automatically downloaded?

inductiva/tasks/task.py
david451lima: with respect to

begaiym-k: Thanks David!

PR #648, opened 2023-09-20
Addressing Sarmento's Comments

cristianacarpinteiro:


README.md
```diff
@@ -159,18 +161,12 @@ Running simulations asynchronously allows users to launch multiple simulations i
 ```python
```
IvanPombo, line 161: Should this code snippet contain the vehicle download as well? We don't know if people will use it as is. Moreover, let's do velocity in velocity_list and just set velocity_list=[1, 10, 20, 30, 40]. Otherwise, people will clutter the machines with this snippet.

IvanPombo (commented):
I think over all it looks good, change the loop with run_async to not clutter the API. Otherwise, I am just not sure how to present the low-level simulation.

README.md
IvanPombo: This one works as well, and is just 5 simulations, not 50.

IvanPombo (approved):

PR #650, opened 2023-09-20
[CoastalArea] Fix README with better example

david451lima:
Implemented a better example in the readme, with a more realistic bathymetry.
Moved the demo script that can be run in the terminal to the demos folder.
Fixed some typos across the readme


IvanPombo (approved):
Thanks for this! Verify that the output video has the high of the bathymetry highlighted in yellow but the colorbar referring to the Water level. I doubt that in such a region there is water. So we need to split what is the bathymetry and what is the water level in that movie.

PR #654, opened 2023-09-20
[CoastalArea] Add warning if user inputs invalid wave source direction

david451lima:
The user can input to the scneario a bad set of instructions to set the wave. They can be:
-the user choses a direction that contains land (negative values of depths);
-the user choses a wave amplitude that is too big for the shallow waters of the bathymetry (there can't be waves of 5 meters in a pond with 0.1 m depth). Since what we are doing is forcing the 
boundary to have a sinusoidal behaviour, this breaks the simulation.

To check for this two cases, we can implement a _check_valid_wave() method that runs in the initialization of the scenario, raising an error if the wave direction is ill defined or if the amplitude 
exceeds the minimum depth in the boundary where the wave is being generated.

Test:

import inductiva
bathymetry = inductiva.coastal.Bathymetry.from_random_depths(
        x_range=(0, 100),
        y_range=(0, 100),
        x_num=100,
        y_num=100,
        max_depth=1,
        random_seed=12,
        roughness_factor=0.5)

The direction with the most depth is W. E represents land. So:

scenario = inductiva.coastal.CoastalArea(bathymetry=bathymetry,
                                         wave_source_location="E",
                                         wave_amplitude=0.1,
                                         wave_period=5.5)

Returns:

ValueError: Wave amplitude cannot be larger than the minimum depth of the bathymetry in the boundary where waves are generated.

### DIFF
--- a/inductiva/coastal/bathymetry.py
+++ b/inductiva/coastal/bathymetry.py
@@ -249,7 +249,7 @@ class Bathymetry:
     def y_num(self) -> int:
         """Returns the length of unique y values."""
 
-        return int(len(self.x_uniques()))
+        return int(len(self.y_uniques()))
 
     @optional_deps.needs_coastal_extra_deps
     def x_uniques(self, sort: bool = False) -> np.ndarray:
@@ -274,6 +274,12 @@ class Bathymetry:
         if sort:
             y_uniques = np.sort(y_uniques)
         return y_uniques
+
+    @optional_deps.needs_coastal_extra_deps
+    def depths_grid(self) -> np.ndarray:
+        """Returns the depths as a 2D grid."""
+
+        return self.depths.reshape((self.x_num(), self.y_num()))
 
     @optional_deps.needs_coastal_extra_deps
     def crop(self,
--- a/inductiva/coastal/coastal_area.py
+++ b/inductiva/coastal/coastal_area.py
@@ -160,9 +160,9 @@ class CoastalArea(scenarios.Scenario):
         if self.wave_amplitude > minimum_depth:
             raise ValueError("Wave amplitude is larger than the "
                              "minimum depth of the bathymetry in the "
-                             "boundary where waves are generated."
+                             "boundary where waves are generated. \n"
                              "This can be due to a very high wave "
-                             "amplitude the waves are being generated "
+                             "amplitude or the waves are being generated "
                              "on land")
 
 
--- a/inductiva/coastal/coastal_area.py
+++ b/inductiva/coastal/coastal_area.py
@@ -5,6 +5,7 @@ import math
 import os
 import shutil
 from typing import Literal, Optional
+import numpy as np
 
 from absl import logging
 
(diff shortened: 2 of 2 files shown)


inductiva/coastal/bathymetry.py
```diff
@@ -275,6 +275,12 @@ def y_uniques(self, sort: bool = False) -> np.ndarray:
             y_uniques = np.sort(y_uniques)
         return y_uniques
 
+    @optional_deps.needs_coastal_extra_deps
```
IvanPombo, line 278: Now that I am thinking, I think we do not need all of these optional_deps, since the numpy is actually in the default package. Only the plots should require this. Let's remove them in another PR, not urgent.

inductiva/coastal/coastal_area.py
```diff
@@ -126,6 +128,39 @@ def get_config_filename(self, simulator: simulators.Simulator):
...
+            wave_boundary = depths_grid[-1, :]
+
+        elif self.wave_source_location == "N":
+            wave_boundary = depths_grid[:, -1]
+
+        elif self.wave_source_location == "S":
+            wave_boundary = depths_grid[:, 0]
+
+        else:
+            raise ValueError("Invalid wave source location.")
```
IvanPombo, line 153: So this just validates if the options are not in "S, W, N, E", right? I thought the goal was to identify that the direction cannot be at the highest points of the bathymetry (the beach). Confirm me this.

inductiva/coastal/coastal_area.py
```diff
@@ -126,6 +128,39 @@ def get_config_filename(self, simulator: simulators.Simulator):
...
+
+        #The greater the value of depth, the deeper the water, so
+        #we want to check the minimum depth using the min value
+
+        minimum_depth = np.min(wave_boundary)
+        if self.wave_amplitude > minimum_depth:
```
IvanPombo, line 159: I think this kind of catches the error I mentioned above - Let's think that we have a bitch on the wave source location, then the minimum_depth will be negative. Which implies all wave_amplitudes will raise an error.

However, the error below is confusing because of the depths of the bathymetry (positive and negative signs), do you want to rephrase slightly so it is more intuitive what is going on.

inductiva/coastal/coastal_area.py
david451lima: It does both

IvanPombo: Ok great catch! Just clarify the error message below to mention the possibilities.

inductiva/coastal/coastal_area.py
IvanPombo: Great, but I don't think that is super explicit for some one using this for the first times. I actually like a lot more your first line: raises an error if the amplitude exceeds the minimum depth or if the wave source is in the beach - clean and simple.

david451lima: Chatting with @sarmento we found out that there is a bunch of pathological cases that we dont yet know if fall in one of the buckets. His suggestion is to mantain the error as is (an objective truth) and suggest to the user the possible causes for it. This leads the user to inspect the bathymetry and make the right decision. In the future we can implement this if we understand the pathological cases.

david451lima: So it will be something like: "Wave amplitude is larger than the "
                              "minimum depth of the bathymetry in the "
                              "boundary where waves are generated.
This can be either due to a very high wave amplitude or
ill defined wave source direction."

PR #655, opened 2023-09-20
Change dualsph example in main readme

luispcunha:
I added a zip file to a public cloud bucket with input files for the dualsphysics low level call, since it is the one in the main readme. The example now should run as is. I'm still fine tuning the input files to make sure the simulation doesn't take too long.
We can then follow a similar strategy for all our demos.


IvanPombo (approved):
LGTM! THANKS!

PR #658, opened 2023-09-20
Put table horizontal to fix empty space on the side

cristianacarpinteiro:


IvanPombo:
On the phone it does not look good... so I think we should keep the vertical one.

cristianacarpinteiro:
Horizontal looks like this : image
But if the formatting on the phone is worse, we can keep it vertical.

hpenedones:
I think that Sarmento's suggestion was somewhere in between those two options: if we have 4 scenarios, let's do a 2x2 table. I know that there is the detail of the headers with the names of the scenarios, so I am not sure how the rendering would look like.

cristianacarpinteiro:
@hpenedones like this? 
image

hpenedones:
yes!!  :) great!

hpenedones (approved):

PR #660, opened 2023-09-21
Use `replace` instead of `rename` in download utils for consistency Windows v Posix

luispcunha:


IvanPombo (approved):
Nicely thought! Can you check if the pytest related with the download_from_url is working correctly?

IvanPombo:
> > Nicely thought! Can you check if the pytest related with the download_from_url is working correctly?
> 
> The pipeline is green, so yes it is. Those tests are run as part of the github actions, so if they failed we would notice here in the PR

I know. Yesterday, it was also green on my side, but then it went red on main.

PR #662, opened 2023-09-21
[Protein Solvation] Update README 

cristianacarpinteiro:
In this PR, I've added the mention to the new ignore_warnings flag and the notebook only support for the visualization. I don't like the fact that we need to put the ignore_warnings flag in the example, so I will update it to an example from alpha fold to avoid this in another PR (I need to generate new visualizations to do it).


IvanPombo:
If you do not like it, why should we just not set the default to True and warn the user that is occuring. When he does simulate, he receives a message explaining that and telling him how to turn it off.

inductiva/molecules/protein_solvation/README.md
```diff
@@ -2,14 +2,14 @@
...
+
 To initialize the scenario, the user can define the following parameters:
 - `protein_pdb` : PDB file containing the initial conformation of the protein;
 - `temperature`: Temperature of the system in K.
 
-Before simulation, files undergo automatic pre-processing in our machines to strip away all non-protein components of the system, such 
-as ligands, non-standard residues, or water molecules, leaving only the protein.
-We tested the scenario with structures predicted by the [AlphaFold2 model](https://alphafold.ebi.ac.uk) and structures from the [RCSB Protein Data Bank](https://www.rcsb.org), but you can provide any other well-formatted PDB file as input, as long as its contains at least one protein chain with all its atoms coordinates. 
-
+Before simulation, files undergo automatic pre-processing in our machines to strip away all non-protein components of the system, such as ligands, non-standard residues, or water molecules, leaving only the protein.
```
IvanPombo, line 11: change in our machines to *in the backend.

inductiva/molecules/protein_solvation/README.md
```diff
@@ -2,14 +2,14 @@
...
+We tested the scenario with structures predicted by the [AlphaFold2 model](https://alphafold.ebi.ac.uk) and structures from the [RCSB Protein Data Bank](https://www.rcsb.org), but you can provide any other well-formatted PDB file as input, as long as its contains at least one protein chain with all its atoms coordinates. Please be aware that PDB files downloaded directly from RCSB may need extra preparation, like filling in missing atoms. Our API doesn't handle this kind of preparation, so you can use software like [Modeller](https://salilab.org/modeller/) or [Chimera](https://www.cgl.ucsf.edu/chimera/) for this purpose.
```
IvanPombo, line 12: I don't really like that we are saying this in the end... Since we actually do some pre-processing to strip things out and then for other extra work, like putting things in, we don't do it.

This bring any errors to the users? Have you found any?

inductiva/molecules/protein_solvation/README.md
```diff
@@ -25,13 +25,13 @@ The user can further define the following simulation parameters:
                 position Langevin dynamics.
     Default value if "md".
 - `n_steps_min`: Number of steps for energy minimization. Default value is 5000.
+- `ignore_warnings`: Whether to ignore warnings during grompp (gromacs preprocessor) step. If set to False, the simulation will fail if there are warnings. If True, the simulation will run and ignore the warnings, but the results may have inaccuracies. Use with caution.
```
IvanPombo, line 28: If this is just during pre-processing, should the name of the argument state that?

inductiva/molecules/protein_solvation/README.md
```diff
@@ -25,13 +25,13 @@ The user can further define the following simulation parameters:
                 position Langevin dynamics.
     Default value if "md".
 - `n_steps_min`: Number of steps for energy minimization. Default value is 5000.
+- `ignore_warnings`: Whether to ignore warnings during grompp (gromacs preprocessor) step. If set to False, the simulation will fail if there are warnings. If True, the simulation will run and ignore the warnings, but the results may have inaccuracies. Use with caution. 
 
-Moreover, the hardware and interaction are configured with the usual general parameters - `machine_group`, `run_async`.
- Launching a simulation returns a task object, which can be used to verify the status of the simulation, get the simulation outputs and access post-processing tools. See more in [Tasks](inductiva/README.md).
+Moreover, the hardware and interaction are configured with the usual general parameters - `machine_group`, `run_async`. Launching a simulation returns a task object, which can be used to verify the status of the simulation, get the simulation outputs and access post-processing tools. See more in [Tasks](inductiva/README.md).
 
 ### Example
```
IvanPombo, line 32: Following a Sarmento suggestion for the WindTunnel, I would set this example right at the top after the first introduction.

inductiva/molecules/protein_solvation/README.md
```diff
@@ -25,13 +25,13 @@ The user can further define the following simulation parameters:
...
 To test this scenario, you can download a protein from the RCSB database using its PDB ID. For this example, let's consider insulin with a PDB ID of 1ZNI.
-Do not forget to insert your API Key (get one by filling this [form](https://docs.google.com/forms/d/e/1FAIpQLSflytIIwzaBE_ZzoRloVm3uTo1OQCH6Cqhw3bhFVnC61s7Wmw/viewform?usp=sf_link)).
+Do not forget to insert your API Key, if you don't have one, checkout [API access tokens](inductiva/README.md).
```
IvanPombo, line 34: Set the full link with a pointer to the section of the API - see the WindTunnel to see how.

cristianacarpinteiro:
> If you do not like it, why should we just not set the default to True and warn the user that is occuring. When he does simulate, he receives a message explaining that and telling him how to turn it off.

Because the default to True will hide the problem and I want the user to inspect the simulation warnings before running a big simulation. I can set the default to True but then the simulation will run and the user will have to kill it... I don't know, ignore warnings by default seems  a bad practice. I prefer to have the extra work and do it with a clean alpha fold prediction.

IvanPombo (commented):
I left a first round comments to clean-up! Thanks for this!

IvanPombo:
> > If you do not like it, why should we just not set the default to True and warn the user that is occuring. When he does simulate, he receives a message explaining that and telling him how to turn it off.
> 
> Because the default to True will hide the problem and I want the user to inspect the simulation warnings before running a big simulation. I can set the default to True but then the simulation will run and the user will have to kill it... I don't know, ignore warnings by default seems a bad practice. I prefer to have the extra work and do it with a clean alpha fold prediction.

Alright!

inductiva/molecules/protein_solvation/README.md
cristianacarpinteiro: Yes, it is a big problem with the RCSB PDB files. I opened an issue, about it actually. Look here to understand better https://github.com/inductiva/inductiva-web-api/issues/311. But I changed the text to: "Please be aware that PDB files downloaded directly from RCSB may need extra preparation, other than removing non-protein elements, like filling in missing atoms. Our API doesn't handle this of pre-processing step, so you can use software like [Modeller](https://salilab.org/modeller/) or [Chimera](https://www.cgl.ucsf.edu/chimera/) for this purpose."

inductiva/molecules/protein_solvation/README.md
cristianacarpinteiro: You call grompp several times in our commands, each time you make a energy minimization. So I wouldn't call it ignore_warnings_pre_processing, maybe ignore_grompp_warnings?

inductiva/molecules/protein_solvation/README.md
cristianacarpinteiro: thank you :) fixed it

inductiva/molecules/protein_solvation/README.md
IvanPombo: I understand, but as an argument name it is rather unclear... ignore_grompp is to specific since that is not passed to the user either. Do what you feel it is best.

inductiva/molecules/protein_solvation/README.md
```diff
@@ -54,6 +34,30 @@ output = task.get_output()
  ```
 This may take up to 2 minute, depending on which machine you use. 
 The last code line downloads the files necessary to render and post-process our simulation.
+To initialize the scenario, the user can define the following parameters:
```
IvanPombo, line 37: Break with title ## Scenario configuration

IvanPombo (approved):
LGTM!

PR #663, opened 2023-09-21
[CoastalArea] README edit

david451lima:
Edited REadME to conform with various suggestions.

CoastalArea demo was removed -> Now all the code previously there is in the beggining of the readme


inductiva/coastal/README.md
```diff
@@ -9,29 +9,55 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
 `pip install --upgrade "inductiva[coastal_extra]"`.
 
-### Example
+### Example 1: Simulating waves on a synthetic bathymetry
 
-Note: You can follow this example in a notebook or run [this script](/demos/py_scripts/coastal_area.py), that contains all the code
-snippets exhibited here. Don't forget to insert your API key
-(`inductiva.api_key = "your_api_key"`) for the script to work! (You can get one by
-filling this [form](https://docs.google.com/forms/d/e/1FAIpQLSflytIIwzaBE_ZzoRloVm3uTo1OQCH6Cqhw3bhFVnC61s7Wmw/viewform?usp=sf_link))
+You can follow this example in your machine to simulate and visualize the wave propagation in a random bathymetry
```
IvanPombo, line 14: take out in your machine, since it may point to them using their machine to run things.

inductiva/coastal/README.md
```diff
@@ -9,29 +9,55 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+that represents a Coastal Area.
 
-The bathymetry that we use can be loaded from a real scenario or randomly generated
-by us. To avoid having to download any data for now, we will start by run a simulation
+Do not forget to insert your API Key (check the [main page](https://github.com/inductiva/inductiva/tree/main#api-access-tokens) to see how get one).
```
IvanPombo, line 17: Awesome!

inductiva/coastal/README.md
```diff
@@ -9,29 +9,55 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+
+```python
```
IvanPombo, line 19: I like this!

inductiva/coastal/README.md
```diff
@@ -9,29 +9,55 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+                                             wave_period=5.5)
+
+task = scenario.simulate(simulation_time=80, output_time_step=1, fps=5)
+
+output = task.get_output()
+
+output.render(movie_path="movie_path.mp4")
+```
+
+To avoid having to download any data for now, we will start by run a simulation
```
IvanPombo, line 45: But then we will repeat things all over again...

IvanPombo: I would suggest that this first lines are moved to before the full example.

inductiva/coastal/README.md
```diff
@@ -9,29 +9,55 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
 on synthetic bathymetry, that is procedurally generated using the API.
 
 We start by generating a random bathymetry.
```
IvanPombo, line 48: This is a comment on the previous example.

inductiva/coastal/README.md
```diff
@@ -9,29 +9,55 @@ is performed using the [SWASH](https://swash.sourceforge.io/) simulator.
...
+          x_num=100,
+          y_num=100,
+          max_depth=1,
+          random_seed=12,
+          initial_roughness=1,
+          roughness_factor=0.5,
+          percentile_above_water=20)
 ```
 
 This bathymetry is defined in a 100x100 grid, that represent a domain 100m
```
IvanPombo, line 63: With this as well.

IvanPombo: After the example, you can comment a bit more on the bathymetry object if you want and say the methods it contains, like for example the plot... Then you show the pic.

inductiva/coastal/README.md
```diff
@@ -56,30 +82,31 @@ scenario = inductiva.coastal.CoastalArea(bathymetry=bathymetry,
                                          wave_period=5.5)
 ```
 
-Once the scenario is created, run the simulation as follows:
+Once the scenario is created, run the simulation by specifying the simulation
```
IvanPombo, line 85: I feel that this part is very explicit on the code. So you can just remove it. and at the end of script you pass to the next paragraph you have below with the visual.

IvanPombo (changes requested):
See my comments below to remove redundancy.

inductiva/coastal/README.md
david451lima: Better to be redundant. There were some complaints abt output_time_step definition, so I prefer to be thorough here

IvanPombo: But you say that below. For the simulate you can mention it afterwards. For the initialisation i I think it is very explicit.

IvanPombo (approved):
This looks much better! Thanks for doing this @david451lima!

PR #668, opened 2023-09-21
Fix Table links to be absolute

IvanPombo:


cristianacarpinteiro (approved):
Thank you for fixing it, @IvanPombo

PR #678, opened 2023-09-21
Improvements to inductiva.tasks.list()

### LINKED ISSUE #638: Task list(): Duration = n/a is not helpful while the job is still running
This is a little detail that is not a priority but it's a bit annoying.
Listing the tasks tells you the duration only if the task has finished. For example:

```
                 ID       Simulator               Status            Submitted              Started        Duration            VM Type
1695139103940006242           swash              started     19 Sep, 15:58:25     19 Sep, 15:58:25             n/a      c2-standard-4
```

I happen to know that this specific task should take around 10 minutes, so having the duration info there, even if the taks is still running, would be great.  It if ok even if we add a little asterisk just to "suggest" that it's a "different" duration such as

```
                 ID       Simulator               Status            Submitted              Started        Duration            VM Type
1695139103940006242           swash              started     19 Sep, 15:58:25     19 Sep, 15:58:25     *0h 5m 32s      c2-standard-4
```

But it would be very helpful for the user to know how long it took since it was submitted (or started if this is what this duration means).

luispcunha:
This PR pushes a few improvements to inductiva.tasks.list(), namely:
- the times displayed are now in the correct timezone (at least in the timezone that python assumes correct);
- for currently running tasks, the duration is shown with an asterisk to indicate for how long the task has been running;
- a column with the scenario name was added.
Example:
```
>>> inductiva.tasks.list()
                 ID               Scenario       Simulator               Status            Submitted              Started        Duration            VM Type
1695309310050950437                    n/a    dualsphysics              started     21 Sep, 16:15:10     21 Sep, 16:15:10       *0h 2m 5s      c2-standard-4
1695307352995292367      protein solvation         gromacs              started     21 Sep, 15:42:33     21 Sep, 15:42:35     *0h 34m 41s      c2-standard-4
1695306097851999678      protein solvation         gromacs  executer-terminated     21 Sep, 15:21:38     21 Sep, 15:21:40      0h 15m 43s      c2-standard-4
1695294928922012701            wind tunnel        openfoam              success     21 Sep, 12:15:31     21 Sep, 12:15:33       0h 0m 50s      c2-standard-4
1695294363618736570            wind tunnel        openfoam              success     21 Sep, 12:06:06     21 Sep, 12:06:06       0h 0m 50s      c2-standard-4
```
One unaddressed suggestion was the possible confusing status names. Those I think could benefit from a more thoughtful change, possibly including the backend, in order to keep status consistent with the DB, the python client, and the web console.

Closes #638

### DIFF
--- a/inductiva/tasks/methods.py
+++ b/inductiva/tasks/methods.py
@@ -122,12 +122,12 @@ def list(last_n: int = 5,
     Example usage:
         # list the last 5 tasks that were successful
         inductiva.tasks.list(5, status="success")
-                         ID       Simulator               Status            Submitted              Started        Duration            VM Type
-        1691150776862178362        openfoam              success     04 Aug, 12:06:17     04 Aug, 12:06:18       0h 1m 53s      c2-standard-8
-        1691149904961476240        openfoam              success     04 Aug, 11:51:46     04 Aug, 11:51:46       0h 1m 28s      c2-standard-8
-        1691081881158823776        openfoam              success     03 Aug, 16:58:02     03 Aug, 16:58:02       0h 1m 20s     n2-standard-32
-        1691081409916619414        openfoam              success     03 Aug, 16:50:11     03 Aug, 16:50:11       0h 1m 20s     n2-standard-32
-        1691080520213617518        openfoam              success     03 Aug, 16:35:21     03 Aug, 16:35:21       0h 1m 23s     n2-standard-32
+                         ID               Scenario       Simulator               Status            Submitted              Started        Duration            VM Type
+        1695309310050950437                    n/a    dualsphysics              started     21 Sep, 16:15:10     21 Sep, 16:15:10       *0h 2m 5s      c2-standard-4
+        1695307352995292367      protein solvation         gromacs              started     21 Sep, 15:42:33     21 Sep, 15:42:35     *0h 34m 41s      c2-standard-4
+        1695306097851999678      protein solvation         gromacs              success     21 Sep, 15:21:38     21 Sep, 15:21:40      0h 15m 43s      c2-standard-4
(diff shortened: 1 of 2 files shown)


IvanPombo (approved):
LGTM!

inductiva/tasks/task.py
```diff
@@ -353,22 +364,27 @@ def _get_path_params(self) -> _PathParams:
...
             return None
         # start_time may be None if the task was killed before it started
         if info["start_time"] is None:
             return None
 
         # Format the time to datetime type
         start_time = datetime.datetime.fromisoformat(info["start_time"])
-        end_time = datetime.datetime.fromisoformat(info["end_time"])
+        end_time = info.get("end_time")
+        if end_time is None:
```
begaiym-k, line 384: What is the case when the end time is none?

inductiva/tasks/task.py
```diff
@@ -353,22 +364,27 @@ def _get_path_params(self) -> _PathParams:
...
+            end_time = datetime.datetime.now(datetime.timezone.utc)
```
begaiym-k, line 385: And why are you setting it to now?

inductiva/tasks/methods.py
```diff
@@ -51,33 +51,45 @@ def _fetch_tasks_from_api(
...
+            execution_time = format_utils.seconds_formatter(execution_time)
+            if end_time is None:
+                execution_time = f"*{execution_time}"
+
         row = [
             task.id,
+            scenario,
             simulator,
             task.get_status(),
             info.get("input_submit_time", None),
```
hpenedones, line 80: is there a way to get this info (e.g. scenario name, and the execution time, being computed outside this method that is responsible to also render the table as a ASCII string? The goal is that I could also use it from the web console.

hpenedones: Thanks Luis! Yes, that is ok, it is not super urgent.

PR #683, opened 2023-09-21
Implement cli to call "list" functions

luispcunha:
This PR adds a "quick and dirty" CLI that can be used to call the `list` methods we currently have in the API. Implementing it was quite straightforward, so it is a nice tool that we can explore for some of the features of the API. Listing tasks is a natural candidate since I found myself several times opening a Python interpreter just to call the `tasks.list()` function.

Another change is that now the version is not directly specified in the `setup.cfg`, but a `__version__` attribute in the `__init__.py`. This is so that the version is easily accessible from the code, and we can display it for `inductiva --version`.

Snippets:
Screenshot 2023-09-21 at 18 09 50

### DIFF
--- a/inductiva/_cli/main.py
+++ b/inductiva/_cli/main.py
@@ -8,7 +8,7 @@ from inductiva import _cli
 def main():
     parser = argparse.ArgumentParser(
         prog="inductiva",
-        description="CLI tool for Inductiva API.",
+        description="CLI tool for Inductiva API",
     )
 
     parser.add_argument(
@@ -22,21 +22,21 @@ def main():
         "--api-key",
         type=str,
         help=("API key to use. If not provided, it "
-              "will be read from the INDUCTIVA_API_KEY environment variable."),
+              "will be read from the INDUCTIVA_API_KEY environment variable"),
     )
 
     # If no subcommand is provided, print help
-    parser.set_defaults(func=parser.print_help)
+    _cli.utils.show_help_msg(parser)
 
     # Create subcommands
     subparsers = parser.add_subparsers()
     tasks_subparser = subparsers.add_parser(
         "tasks",
-        help="View tasks information.",
+        help="View tasks information",
     )
     resources_subparser = subparsers.add_parser(
         "resources",
-        help="View resources information.",
+        help="View resources information",
     )
 
     # Register subcommands (e.g. list) for each subcommand (e.g. tasks)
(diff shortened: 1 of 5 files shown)


hpenedones:
This is cool!!! :)

inductiva/resources/storage.py
```diff
@@ -9,8 +8,8 @@ def get_space_used():
     try:
         api = instance_api.InstanceApi(inductiva.api.get_client())
         instance_price = api.get_storage_size()
-        logging.info("Total storage used: %s GB",
-                     float(round(instance_price.body, 3)))
-        return float(round(instance_price.body, 3))
+        instance_price = float(round(instance_price.body, 3))
+        print(f"Total storage used: {instance_price} GB")
```
IvanPombo, line 12: Can you clarify why do we need this change? 

Just worried since this also modifies the calling of the function.

IvanPombo (approved):
This is great indeed! Thanks for this @luispcunha!!

inductiva/resources/storage.py
IvanPombo: When we remove the async, that will make completely sense, since users will lose the information otherwise.

inductiva/_cli/tasks.py
```diff
@@ -0,0 +1,20 @@
...
+
+def register_tasks_cli(parser):
+    subparsers = parser.add_subparsers()
+    list_subparser = subparsers.add_parser("list", help="List tasks.")
+    list_subparser.add_argument(
+        "-n",
+        "--last-n",
+        type=int,
+        default=5,
+        help="List last N tasks. Default: %(default)s",
```
begaiym-k, line 17: Number? to be more explicit.

inductiva/_cli/main.py
```diff
@@ -0,0 +1,54 @@
+"""Main CLI entrypoint."""
```
begaiym-k, line 1: You can also add an example commands you have in the PR description here.

inductiva/_cli/tasks.py
begaiym-k: Number of recent tasks to list?

inductiva/_cli/main.py
begaiym-k: That's true, but then I think as we move forward we will still need to add a small documentation in the readme and keep it up to date.

PR #690, opened 2023-09-22
point links to readme titles

cristianacarpinteiro:
I tried to solve the links issue directing them to the titles of the readme, however the title gets highlighted, like this: 
image
The links with the readme pointer appear to work if you open them in another tab. What solution is best?

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,10 +92,10 @@ These are the currently available scenarios:
 
 <div align="center">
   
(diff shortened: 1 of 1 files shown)


sarmento:
That's ok! At least the users sees text instead of a list of unrelated filenames.
That's a better experience. If it works that way then LGTM

sarmento (approved):

PR #696, opened 2023-09-22
Attempt to fix image links

IvanPombo:


IvanPombo:
Based on this: https://stackoverflow.com/questions/41983209/how-do-i-add-images-to-a-pypi-readme-that-works-on-github it should contain raw in the end, or change blob by raw in the link.

Issue #697, opened 2023-09-24
[Info to the user] Add more detail (and a few more little things)


sarmento:
Here is a dump of running on example from the terminal (swash)

```
INFO:absl:Task ID: 1695390640821818395
INFO:absl:Creating input archive...
INFO:absl:Input archive created.
INFO:absl:Input archive size: 1.56 MiB
INFO:absl:Uploading input...
INFO:absl:Input archive uploaded.
INFO:absl:Task request submitted.
INFO:absl:Waiting for resources...
INFO:absl:The task is being executed remotely.
INFO:absl:Task completed successfully.
INFO:absl:Task completed successfully.
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 76.5M/76.5M [03:20<00:00, 400kiB/s]
INFO:absl:Post-processing tools are available through the output object.
INFO:absl:Starting to render video from simulation data...
INFO:absl:Writing mp4 file to movie.mp4.
```

Here are my comments/wishes:

```
INFO:absl:Task ID: 1695390640821818395
```

1) I wish I had a timestamp/date in all these messages.
2) I wish I had a link so I could be redirected to the console

```
INFO:absl:Creating input archive...
```
I wish I had more context: "Creating input archive to be upload to inductiva.ai"
```
INFO:absl:Input archive created.
INFO:absl:Input archive size: 1.56 MiB
```
These to messages could be merged.
Also, what is "MiB"? :)

```
INFO:absl:Uploading input...
```
"input archive."
Since we have a progress bar for downloading, can we have one for uploading.
```
INFO:absl:Input archive uploaded.
INFO:absl:Task request submitted.
```
to which pool?
```
INFO:absl:Waiting for resources...
```
If you could tell me a bit more, that would be great.
Metric suggestion (for our metrics): how long is it taking from a task to be picked up?
If we have live access to that metric, then let's show it here ("Current average waiting time = Xs").

```
INFO:absl:The task is being executed remotely.
```
Can you tell me which hardware picked this up?
```
INFO:absl:Task completed successfully.
INFO:absl:Task completed successfully.
```

Why are repeating this message?
```
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 76.5M/76.5M [03:20<00:00, 400kiB/s]
```
What does this progress bar refer to? (I know it's download but it's not explicit)
Also, print the path to where this data is being downloaded

```
INFO:absl:Post-processing tools are available through the output object.
```
[...]

PR #702, opened 2023-09-26
Improve logging overall

IvanPombo:
This PR attempts to address some of the concerns presented in the issue #697. It addresses some of the inconsistencies between loggings, it breaks between subsequent steps of the pipeline, adds time logs and improves the overall feedback!

Notice that moving to just using async, some of the logs will be unused, so after that merge I will pick this again and do another round of improvements.

Here is the log obtain with the following changes:

Screenshot 2023-09-26 at 13 06 54

### DIFF
--- a/inductiva/coastal/output.py
+++ b/inductiva/coastal/output.py
@@ -14,7 +14,7 @@ except ImportError:
     scipy = None
 
 from inductiva.types import Path
-from inductiva.utils import visualization, optional_deps
+from inductiva.utils import visualization, optional_deps, files
 
 # Labels used in SWASH simulation outputs.
 QUANTITY_SWASH_LABELS = {
@@ -260,7 +260,10 @@ class CoastalAreaOutput:
             quantity=quantity,
         )
 
-        logging.info("Starting to render video from simulation data...")
+        movie_path = files.resolve_path(movie_path)
+        logging.info("Rendering video to %s of the %s with %d fps.",
+                     movie_path,
+                     QUANTITY_PLOT_LABELS[quantity], fps)
 
         _render_quantity_grid_data(
             x_array=x_array,
@@ -273,5 +276,3 @@ class CoastalAreaOutput:
             cmap=cmap,
             clim=clim,
         )
-
-        logging.info("Writing mp4 file to %s.", movie_path)
--- a/inductiva/fluids/_post_processing.py
+++ b/inductiva/fluids/_post_processing.py
@@ -1,7 +1,7 @@
 """Post process SPH simulation outputs."""
 import os
 
-from inductiva.utils import visualization, optional_deps
+from inductiva.utils import visualization, optional_deps, files
 from inductiva.types import Path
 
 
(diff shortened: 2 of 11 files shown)


inductiva/coastal/output.py
```diff
@@ -260,7 +260,9 @@ def render(
             quantity=quantity,
         )
 
-        logging.info("Starting to render video from simulation data...")
+        movie_path = files.resolve_path(movie_path)
+        logging.info("Rendering video to %s of the %s with %d fps.", movie_path,
```
cristianacarpinteiro, line 264: Maybe " Saving rendered video to ..."

inductiva/fluids/_post_processing.py
```diff
@@ -33,7 +35,9 @@ def render(self,
         """
 
         vtk_dir = os.path.join(self.sim_output_dir, "vtk")
+        movie_path = files.resolve_path(movie_path)
 
+        logging.info("Rendering movie to %s with %d fps.", movie_path, fps)
```
cristianacarpinteiro, line 40: Maybe say video here instead of movie for consistency.

inductiva/fluids/fluid_tank/output.py
```diff
@@ -57,6 +59,7 @@ def render(
 
         movie_path = files.resolve_path(movie_path)
 
+        logging.info("Rendering movie to %s with %d fps.", movie_path, fps)
```
cristianacarpinteiro, line 62: Same here, either movie or video, but use the same term everywhere.

inductiva/fluids/post_processing/openfoam.py
```diff
@@ -279,7 +281,7 @@ def render(self,
         plotter.show()
         if save_path is not None:
             plotter.screenshot(save_path, return_img=False)
-            logging.info("Flow slice rendering saved to %s", save_path)
+            logging.info("Flow slice rendering saved to %s.", save_path)
```
cristianacarpinteiro, line 284: Shouldn't it be flow slice render instead of rendering? not sure!

inductiva/tasks/task.py
```diff
@@ -347,13 +349,14 @@ def download_outputs(
 
         zip_path = output_dir.joinpath("output.zip")
 
+        logging.info("Downloading simulation outputs to %s.", zip_path)
         data.download_file(response, zip_path)
 
         if uncompress:
+            logging.info("Uncompress the outputs to %s.", output_dir)
```
cristianacarpinteiro, line 356: Maybe say Uncompressing instead to be consistent with the Downloading.

cristianacarpinteiro (approved):
Thank you @IvanPombo :) This looks more consistent!

cristianacarpinteiro:
Looking at the print maybe we can say "Starting machine group. This may take a few minutes." it's more concise and passes the same message.

begaiym-k (approved):
Thanks @IvanPombo! LGTM!

inductiva/fluids/_post_processing.py
IvanPombo: I will actually revert to movie, since everything contains movie.

PR #710, opened 2023-09-27
Fix var name

begaiym-k:


inductiva/resources/storage.py
```diff
@@ -7,9 +7,9 @@ def get_space_used():
     """Returns the occupied storage size in GB."""
     try:
         api = instance_api.InstanceApi(inductiva.api.get_client())
-        instance_price = api.get_storage_size()
-        instance_price = float(round(instance_price.body, 3))
-        print(f"Total storage used: {instance_price} GB")
-        return instance_price
+        response = api.get_storage_size()
+        sstorage_used = float(round(response.body, 3))
```
IvanPombo, line 11: typo *sstorage. Isn't round already a float?

begaiym-k: not really, it comes in the FastAPI schema type, so when I print it, it looks like this: `<DynamicSchema 10>`, that's why I am converting it to float.

begaiym-k: Also, fixed the typo

begaiym-k: Apparently, yes. I am not sure how exactly it works.

IvanPombo: It feels like it should be reversed, for any case we change the response.body and round won't work anymore.

Otherwise, it seems good to go!

IvanPombo (approved):

hpenedones (approved):

PR #722, opened 2023-09-28
[OpenFOAM] Remove the n_cores parameter from Inductiva

IvanPombo:
This PR matches the one in the backend and finally removes the `n_cores` from its final dependency on the OpenFOAM.

The commands were shifted to contain a framework similar to the backend.

Tested locally with the matching version of the backend and by running the WindTunnel scenario.


hpenedones (approved):
Thanks Ivan, our interface is getting cleaner, step by step! :)

PR #724, opened 2023-09-29
[FEniCSx] Move Gmsh to the backend

Joao97ribeiro:
This PR includes modifications to the client-side code associated with the relocation of Gmsh to the backend (https://github.com/inductiva/inductiva-web-api/pull/328)


inductiva/__init__.py
```diff
@@ -13,7 +13,7 @@
 from . import generative
 from . import molecules
 from . import resources
-# from . import structures
```
IvanPombo, line 16: Please make sure this completely works :)

inductiva/structures/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -55,28 +67,33 @@ def __init__(
...
-            material_filename: Material filename.
         """
         simulator.override_api_method_prefix("deformable_plate")
-        task = super().simulate(simulator,
-                                machine_group=machine_group,
-                                run_async=run_async,
-                                mesh_filename=MESH_FILENAME,
-                                bcs_filename=BCS_FILENAME,
-                                material_filename=MATERIAL_FILENAME)
+        task = super().simulate(
```
IvanPombo, line 87: See this is exactly as the Linear one, I think...

inductiva/simulators/fenicsx.py
```diff
@@ -15,25 +15,44 @@ def __init__(self):
...
+
         return super().run(input_dir,
                            machine_group=machine_group,
                            run_async=run_async,
-                           mesh_filename=mesh_filename,
+                           geometry_filename=geometry_filename,
                            bcs_filename=bcs_filename,
-                           material_filename=material_filename)
+                           material_filename=material_filename,
+                           global_refinement_factor=global_refinement_factor,
```
IvanPombo, line 57: These parameters should just be passed to the Deformable Plate at this point... At least, for now they aren't used for the Linear.

You should be able to just pass them below with *kwargs or something.

IvanPombo (changes requested):
Left minor suggestion to uniformize with the backend. Address it first and check that everything is running by installing in different ways :)

I will unlock you when you give me a thumbs up with a test running :)

inductiva/simulators/fenicsx.py
```diff
@@ -15,25 +15,44 @@ def __init__(self):
...
         run_async: bool = False,
     ) -> tasks.Task:
         """Run the simulation.
 
         Args:
-            mesh_filename: Mesh filename.
+            geometry_filename: Geometry filename.
             bcs_filename: Boundary conditions filename.
             material_filename: Material filename.
+            global_refinement_factor (float): The refinement factor for global
```
fabiocruz, line 32: This should be called `meshing_global_refinement_factor` or something similar, to make it explicit that this is connected to the meshing process.

inductiva/simulators/fenicsx.py
fabiocruz: Not sure I understand or agree with this, @IvanPombo. These parameters control the resolution of the simulation, so they should be arguments here and not attributes of the plate class.

inductiva/structures/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -28,13 +26,13 @@ class DeformablePlate(scenarios.Scenario):
...
-        bcs_list: List[structures.bcs.BoundaryCondition],
-        material: structures.materials.IsotropicLinearElasticMaterial,
-    ):
+    def __init__(self,
+                 plate: structures.plates.RectangularPlate,
+                 holes_list: List[structures.holes.Hole],
+                 bcs_list: List[structures.bcs.BoundaryCondition],
+                 material: structures.materials.IsotropicLinearElasticMaterial,
+                 global_refinement_factor: float = 1.0,
+                 local_refinement_factor: float = 0.0):
```
fabiocruz, line 35: These should be parameters of the `simulate` method and not attributes of this class. See my comment above.

fabiocruz (changes requested):
Thanks @Joao97ribeiro this is looking good. I have a few requests for changes - see them below.

Also, this should completely remove our dependency from gmsh, so we can remove it from the package requirements?

inductiva/simulators/fenicsx.py
IvanPombo: On the Backend they are not used by the LinearElastic...Executer, so I am not sure they should be going here so explicitly. It assumes people using the low-level (if though it may not be perfect) can use this to modify the mesh.

inductiva/structures/plate_linear_elastic/plate_linear_elastic.py
Joao97ribeiro: If these attributes do not belong to this class, how will the client provide them?

fabiocruz: As arguments to the simulate method of the scenario or the run method of the simulator.

Joao97ribeiro: Make sense, I will change it, thanks

Joao97ribeiro:
@IvanPombo and @fabiocruz, thank you once again for your valuable comments! I have already implemented all of your suggestions.

Regarding the mesh refinement parameters, I believe they are now in the correct positions. If the client is using the high level, they can call them within the simulation method, and in the low level, they can use them in the run method. Do you both agree?

inductiva/structures/plate_linear_elastic/plate_linear_elastic.py
```diff
@@ -59,24 +54,41 @@ def __init__(
     def simulate(self,
                  simulator: simulators.Simulator = simulators.FEniCSx(),
                  machine_group: Optional[resources.MachineGroup] = None,
-                 run_async: bool = False) -> tasks.Task:
+                 run_async: bool = False,
```
IvanPombo, line 57: Remove this, right now simulations are all async.

IvanPombo (approved):
It is good to go!

fabiocruz (approved):

PR #734, opened 2023-10-03
Add is elastic param

begaiym-k:
Since the FastAPI client was pushed with the changes made in the backend and those are not reflected yet in the code, if you try to develop API locally it will give errors. Here I am adding a missing param to fix it. However, later today I will open a PR with all the changes in the machines.

### DIFF
--- a/inductiva/resources/machines.py
+++ b/inductiva/resources/machines.py
@@ -46,6 +46,7 @@ class MachineGroup():
         # to the backend.
         self._api = instance_api.InstanceApi(api.get_client())
         self._estimated_cost = None
+        self._is_elastic = False
 
         if register:
             self._register_machine_group()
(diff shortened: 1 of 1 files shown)


cristianacarpinteiro (approved):

IvanPombo (changes requested):
I don't really understand why we need this to develop. Is this due to the gen_client? Won't the error still persist?

I am using the main and I am able to develop locally without problems...

begaiym-k:
> I don't really understand why we need this to develop. Is this due to the gen_client? Won't the error still persist?
> 
> I am using the main and I am able to develop locally without problems...

Are you sure you're using the latest version of main? because there it doesn't work.

IvanPombo:
> > I don't really understand why we need this to develop. Is this due to the gen_client? Won't the error still persist?
> > I am using the main and I am able to develop locally without problems...
> 
> Are you sure you're using the latest version of main? because there it doesn't work.

Maybe not the latest... However, I don't like to merge this to main, since it's a quick fix that will be replaced... I think if anyone wants to use this fix, they can merge it to their branch.

PR #743, opened 2023-10-04
[Storage] Delete directory from storage

cristianacarpinteiro:
In this PR, I introduce a new feature to allow the user to delete the storage of a given task.

```python
inductiva.storage.delete_task_directory(task_id)
```

**Example Usage:** 
image
When the folder doesn't exist  the following exception is raised: 
image

I will keep this PR as a draft while #737 is still open.


inductiva/storage/storage.py
```diff
@@ -0,0 +1,103 @@
...
+    return format_utils.get_tabular_str(
+        rows,
+        columns,
+        default_col_space=15,
+        override_col_space=override_col_space,
+        formatters=formatters,
+    )
+
+
+def delete_task_directory(task_id: str):
```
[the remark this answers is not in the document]
cristianacarpinteiro: Yes, I will rename it and renaming it in the backend too.

inductiva/storage/storage.py
```diff
@@ -90,3 +90,18 @@ def _print_contents_table(contents):
...
+    """Deletes a directory inside the user's storage bucket. Currently,
+    directories are named by the task id of the task that created them, 
+    so this function can be used to delete the directory of a task. 
+    Args:
+        dir_name (str): The name of the directory to be deleted.
+    """
+    try:
+        api = instance_api.InstanceApi(inductiva.api.get_client())
+        api.delete_directory({"dir_name": dir_name})
+        print("Task directory deleted successfully.")
```
begaiym-k, line 105: Can you also provide a dir name here?

inductiva/storage/storage.py
```diff
@@ -90,3 +90,18 @@ def _print_contents_table(contents):
         override_col_space=override_col_space,
         formatters=formatters,
     )
+
+
+def rmdir(dir_name: str):
+    """Deletes a directory inside the user's storage bucket. Currently,
```
begaiym-k, line 96: Maybe replace the bucket with `Deletes a directory inside the user's storage in Inductiva API*`? Because I am fond of exposing the GC or backend specific names in the client since it can always change.

begaiym-k (approved):
Thanks @cristianacarpinteiro! This looks great! I just left a few minor comments regarding the logs and readability.

inductiva/client/apis/paths/gcp_instances_storage_dir_name.py
```diff
@@ -0,0 +1,6 @@
+from inductiva.client.paths.gcp_instances_storage_dir_name.delete import ApiFordelete
```
sarmento, line 1: I don't understand this name convention?
Should this be APIForDelete?
Also, are we still importing classes?
Maybe this is auto-generated because it is the client? Is that it?

@hpenedones @begaiym-k @luispcunha : can you please clarify?

cristianacarpinteiro:
image
Updated example

inductiva/client/apis/paths/gcp_instances_storage_dir_name.py
cristianacarpinteiro: Yes, it is the generated client.

sarmento: OK! But, just out of curiosity: would we be able to have the generator generate:

1) better names?
2) shorter structures. For example, do we need all this:

```
inductiva.client.paths.gcp_instances_storage_dir_name.delete
```

I am not saying this is a priority, but I want to make sure we are not "leaving crumbs"...

@hpenedones @begaiym-k @luispcunha : can you please clarify?

sarmento: ok! thank you for the explanation.
I am assuming the InstanceApi is also that way because of the autogeneration (i.e. no InstaceAPI)?

PR #759, opened 2023-10-16
[Storage] Create parent folder to save simulation outputs

cristianacarpinteiro:
In this PR, I introduce a new variable in the simulate method called `storage_parent_dir`. This variable defines the parent directory in the user's storage where the simulation outputs will be saved. 

**Code snippet:** 
```
# Initialize the scenario
scenario = inductiva.molecules.ProteinSolvation(
     protein_pdb = insulin_pdb_file,
     temperature = 300)

# Run a simulation
task = scenario.simulate(simulation_time_ns = 0.001, 
                         storage_dir="proteins/alpha",
                         ignore_warnings = True)
print(task.get_storage_path()) 
"proteins/alpha/1697473263459828764"
```


begaiym-k:
Are you trying to merge it to main afterwards? You should compare it to development because there a lot of unrelated files that are coming from the dev branch.

IvanPombo (approved):
LGTM! I think my only comment is regarding the naming, it could be simpler like `storage_dir`, wdyt?

PR #771, opened 2023-10-18
Change type of caught exception when checking optional deps

### LINKED ISSUE #770: Wrong error message if dependency is not properly installed
Currently, if one of the optional dependencies (*e.g.*, `vtk`) raises an exception when importing, we assume it is due to the package not being installed, thus reporting an error similar to:
```
'SteadyStateOutput.get_object_pressure_field' missing: 'vtk'. You can install the missing dependencies for 'fluids' with `pip install 'inductiva[fluids_extra]'`."
```
However, the package may fail to import for other reasons (*e.g.*, fail to open some shared lib), so we should reflect that in the error message.

luispcunha:
Closes #770

### DIFF
--- a/inductiva/utils/optional_deps.py
+++ b/inductiva/utils/optional_deps.py
@@ -30,7 +30,7 @@ def _needs_optional_deps(module_names, extra_msg, func):
         for module_name in module_names:
             try:
                 importlib.import_module(module_name)
-            except ImportError:
+            except ModuleNotFoundError:
                 missing_imports.append(module_name)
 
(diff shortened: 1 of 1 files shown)


david451lima (approved):
LGTM! SHould we give any warning if we catch an ImportError?

IvanPombo:
> > While you are at it, add the `--upgrade` down below on the `_missing_deps_msg` message.
> 
> why's that one needed here?

To inform the user, or when one needs to install the extra the packages will work flawlessly?

PR #778, opened 2023-10-19
[OpenFOAM] Add multiple versions

IvanPombo:
This PR matches the changes done in the backend, to adequate the deployment of multiple versions of openFOAM.

Right, now users only select between the two main versions: `version="esi"` and `version="foundation"` that constitutes the main differences between forks of OpenFOAM. In the future, we may want to have multiple versions inside the same fork.

But for now, this simple change satisfies both users simultaneously.

This was tested by deploying the backend with both images and using the low-level to test both versions. Notice further that, the "esi" version only is used in the low-level simulator. The scenarios will keep using the foundation, since the template files obey their rules.


begaiym-k (approved):
LGTM!

PR #786, opened 2023-10-23
[Storage] List specific directory with inductiva.storage.listdir

cristianacarpinteiro:
In this PR, I updated the parameters for the` inducitva.storage.listdir() `function.  
Here are some examples on how this works: 
**List root folder:**
image
**List a specific subdirectory:** 
image
**List files:** 
image

Look into the [PR on the backend](https://github.com/inductiva/inductiva-web-api/pull/371) to understand the implementation.


inductiva/storage/storage.py
```diff
@@ -17,11 +17,13 @@ def get_space_used():
...
 
 
-def listdir(max_results: int = 3,
+def listdir(path="/",
+            max_results: int = 10,
             order_by: Literal["size", "creation_time"] = "size",
             sort_order: Literal["asc", "desc"] = "desc"):
     """List and display the contents of the user's storage bucket.
     Args:
+        path (str) : directory inside storage to list from. default is root.
```
begaiym-k, line 26: Default*

begaiym-k: Directory*

begaiym-k: (str):* (no space in between)

begaiym-k: Maybe "Storage directory to list"

inductiva/storage/storage.py
```diff
@@ -17,11 +17,13 @@ def get_space_used():
         raise api_exception
 
 
-def listdir(max_results: int = 3,
+def listdir(path="/",
+            max_results: int = 10,
             order_by: Literal["size", "creation_time"] = "size",
             sort_order: Literal["asc", "desc"] = "desc"):
     """List and display the contents of the user's storage bucket.
```
begaiym-k, line 24: I thought we agreed to not to expose the name bucket in the client?

cristianacarpinteiro:
If we are following the defaults of `os.listdir()` the default should be ".", the current directory. Which for our API, should mean the root of the bucket, however it is not a striagthforward assumption for the user. I set the default path to be "/", which stands for root. What do you think the default should be?

begaiym-k (approved):
LGTM! Thanks @cristianacarpinteiro for these changes, I just left a few comments to update the docstring.

inductiva/storage/storage.py
IvanPombo: You won't need this redudancy, since you already mention above that you will list from this storage path

inductiva/storage/README.md
```diff
@@ -19,9 +19,9 @@ After determining the total storage space used, you may want to identify which d
 
 ```python
 import inductiva
-inductiva.storage.listdir(max_results=10, order_by="size", sort_order="desc")
+inductiva.storage.listdir(path = "/", max_results=10, order_by="size", sort_order="desc")
```
IvanPombo, line 22: If this is the default, than don't state here the path..

inductiva/storage/README.md
```diff
@@ -31,6 +31,12 @@ This function generates a table displaying the storage directories:
 | 1234567| 97.59 KiB  | 29 Sep, 14:13:24|
 | 123    | 0 B        | 29 Sep, 14:13:29|
 ```
+To examine the contents of a specific folder, execute: 
+
+```python
+import inductiva
+inductiva.storage.listdir(path = "1234/", max_results=10, order_by="size", sort_order="desc")
```
IvanPombo, line 38: Do they need to pass the "/" at the end?

IvanPombo (approved):
Left some minor comments, to quickly iterate on. But it LGTM!

inductiva/storage/README.md
cristianacarpinteiro: No, if the user doesn't put it We will add it in the backend.

sarmento: Have we tested that the user can't go up using ".."?

cristianacarpinteiro: image

Issue #573, opened 2023-09-13
Can you please test this? Can you run the code Joao wrote?


IvanPombo:
Can you please test this? Can you run the code Joao wrote?

_Originally posted by @sarmento in https://github.com/inductiva/inductiva/issues/571#issuecomment-1717425493_

IvanPombo:
This issue was now moved into the backend where it is being address by @Franreno. Closing it on the client.

PR #790, opened 2023-10-23
Add FDS fire simulator to API

IvanPombo:
This PR adds the fire simulator FDS to the API.


sarmento:
Let's take the chance to think about how to test if this simulator is running correctly.

@IvanPombo : can we add a test somewhere? 
What would be that test like?
Can we use a simple simulation script provided by Celine?

IvanPombo:
> Let's take the chance to think about how to test if this simulator is running correctly.
> 
> @IvanPombo : can we add a test somewhere? What would be that test like? Can we use a simple simulation script provided by Celine?

That test requires an api-key at the moment. That is the main reason we don't have yet tests for simulations. This was indicated in issue #573... Locally, I have tested with an example script and with the first Céline script.

sarmento:
> > Let's take the chance to think about how to test if this simulator is running correctly.
> > @IvanPombo : can we add a test somewhere? What would be that test like? Can we use a simple simulation script provided by Celine?
> 
> That test requires an api-key at the moment. That is the main reason we don't have yet tests for simulations. This was indicated in issue #573... Locally, I have tested with an example script and with the first Céline script.

Please bring this issue to the attention of @lneves-ai : this is very important to make sure we can properly test things when we submit code to the repo. 

Still it would be ok to have a way of running scripts with no API key. (cc @hpenedones).

hpenedones:
Small notes:

- Even if the test requires an API key at the moment, this could be done using the GitHub Secrets, without exposing the API key in the unit test code. This is a mechanism that we already use in GitHub actions:
https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions

- Regarding not requiring any API key to run things, let's chat a bit more about that @sarmento . The implications would be that under that regime: 1) we couldn't have a storage bucket dedicated to the user 2) the user could not monitor its simulations or outputs via the web console 3) the user could not use a lot of functions of the API (see my tasks, etc).  So the approach I'm suggesting is: let's just make it trivial for users to get personal API keys via the web console (this is simple, and we are working on it).

sarmento:
> Small notes:
> 
> * Even if the test requires an API key at the moment, this could be done using the GitHub Secrets, without exposing the API key in the unit test code. This is a mechanism that we already use in GitHub actions:
>   https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions
> * Regarding not requiring any API key to run things, let's chat a bit more about that @sarmento . The implications would be that under that regime: 1) we couldn't have a storage bucket dedicated to the user 2) the user could not monitor its simulations or outputs via the web console 3) the user could not use a lot of functions of the API (see my tasks, etc).  So the approach I'm suggesting is: let's just make it trivial for users to get personal API keys via the web console (this is simple, and we are working on it).

Then let's go via option 1. 
Let's create an API Key for a test user.
I would really like if we could build some tests asap. Testing if the simulator can run, is a basic one.

david451lima (approved):
LGTM, pretty standard approach.
The scenario only requires a sim_config_file, right? If we require a structure file of the environment that we are simulating that can be passed in the input_dir, right?

IvanPombo:
> LGTM, pretty standard approach. The scenario only requires a sim_config_file, right? If we require a structure file of the environment that we are simulating that can be passed in the input_dir, right?

Yes, indeed! Everything else will be added through the input_dir.

IvanPombo:
> > Small notes:
> > 
> > * Even if the test requires an API key at the moment, this could be done using the GitHub Secrets, without exposing the API key in the unit test code. This is a mechanism that we already use in GitHub actions:
> >   https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions
> > * Regarding not requiring any API key to run things, let's chat a bit more about that @sarmento . The implications would be that under that regime: 1) we couldn't have a storage bucket dedicated to the user 2) the user could not monitor its simulations or outputs via the web console 3) the user could not use a lot of functions of the API (see my tasks, etc).  So the approach I'm suggesting is: let's just make it trivial for users to get personal API keys via the web console (this is simple, and we are working on it).
> 
> Then let's go via option 1. Let's create an API Key for a test user. I would really like if we could build some tests asap. Testing if the simulator can run, is a basic one.

After talking with @hpenedones, I will leave the tests on hold, since they shouldn't be done on github actions due to a recurring call to the cloud. In the next weeks, integration tests will be developed and these ones you desire here can fit nicely there.

hpenedones:
Yes, thanks @IvanPombo : @Franreno and @lneves-ai are looking into a good setup to implement integration tests.

PR #794, opened 2023-10-24
Simplify scenario creation

IvanPombo:
This PR is the first on the line of simplifying the scenario creation. The main goal is to provide this ability to users - heavily requested. 

For this purpose, we move the mechanics of changing the template files all to the Scenario class, based on the function arguments. 

Users will have to follow a certain structure for their template files, but I think the benefit is huge with the pipeline provide to them.

Currently, there are a few extra functions that are required for the scenarios with more than one simulator - the sole example is the FluidBlock (DamBreak inherits this). This type of scenarios are still complex to accomodate the different simulation config files. 

However, most users will just stay with one simulator at a time. The simplification is seen in the WindTunnel scenario. Here, most of the code is to check the args we are setting in, which users may want to do. Then, we provide a method that manages the file arguments - protein file or vehicle file. In these cases, users are responsible to set it in the right place on the folder.

These PR just changes the FluidBlock and WindTunnel scenario, since they represent the two types of structures we currently have. After I receive all the feedback and iterate on it, I clean the rest of the scenarios and create a ReadME to teach users how to create their own scenario.

The code here presented was tested with the default code blocks on the ReadMe's of each scenario based on the development branch of both the Client and backend.


hpenedones:
Thanks Ivan, the code does seem to get simpler!
Could you test that the simulator config files generated from the jinja templates (before and after your change), looked the same?

IvanPombo:
> Thanks Ivan, the code does seem to get simpler! Could you test that the simulator config files generated from the jinja templates (before and after your change), looked the same?

Yes, they looking exactly the same!

inductiva/fluids/wind_tunnel/wind_tunnel.py
```diff
@@ -74,25 +73,25 @@ def __init__(self,
...
         elif len(flow_velocity) != 3:
             raise ValueError("`flow_velocity` must have 3 values.")
         elif np.linalg.norm(np.array(flow_velocity)) > 100:
             raise ValueError("The `flow_velocity` magnitude is too high,"
                              " it must be less than 100 m/s.")
         else:
-            self.flow_velocity = flow_velocity
+            self.params["flow_velocity"] = flow_velocity
 
         if domain is None:
```
begaiym-k, line 85: I just noticed it, but the code will look cleaner if you provide this value as default in the template itself, you can look it up [here](https://support.sendwithus.com/jinja/default/#:~:text=Jinja%20has%20a%20built%20in,when%20one%20isn%27t%20provided.&text=If%20the%20value%20isn%27t,value%20is%20defined%20but%20blank.)

inductiva/fluids/wind_tunnel/wind_tunnel.py
```diff
@@ -122,73 +121,20 @@ def simulate(
...
-        """Returns the commands for the simulation."""
-
-        commands_file_path = os.path.join(SCENARIO_TEMPLATE_DIR,
-                                          OPENFOAM_TEMPLATE_INPUT_DIR,
-                                          COMMANDS_FILE_NAME)
-
-        commands = self.read_commands_from_file(commands_file_path)
-
-        return commands
+    def add_extra_input_files(self, simulator: simulators.Simulator, input_dir):  # pylint: disable=unused-argument
```
begaiym-k, line 135: The name of the method is not self-explanatory, can you be more specific like what extra files?

inductiva/fluids/wind_tunnel/wind_tunnel.py
```diff
@@ -122,73 +121,20 @@ def simulate(
...
-        remove_templates=True,
-    )
-
-    # Add object path to its respective place
-    object_dir = os.path.join(input_dir, "constant", "triSurface")
-    os.mkdir(object_dir)
-    shutil.copy(self.object_path, os.path.join(object_dir, "object.obj"))
+        object_dir = os.path.join(input_dir, "constant", "triSurface")
+        os.mkdir(object_dir)
+        shutil.copy(self.object_path, os.path.join(object_dir, "object.obj"))
```
begaiym-k, line 140: Wow, looks so clean! It is awesome!

inductiva/scenarios.py
```diff
@@ -1,11 +1,13 @@
 """Base class for scenarios."""
 
-from abc import ABC, abstractmethod
+from abc import ABC
```
begaiym-k, line 3: If it's possible, can you also, pick up and fix some of the imports, please? We still have a lot of class imports in these files.

begaiym-k (approved):
This is a great work @IvanPombo! Thanks for this! I left a few comments that I want you to address, once they are done, it's ready to go!

inductiva/fluids/wind_tunnel/wind_tunnel.py
IvanPombo: I agree with you about being more clean, however it will be a bit more obscure for devs looking in, since they will have to check multiple files. However, it is a nice thing to explain for users! Thanks!!

inductiva/fluids/wind_tunnel/wind_tunnel.py
IvanPombo: It depends of the simulator and what users want to "parameterize"... That is why I didn't make more expliciti..

PR #810, opened 2023-11-02
Refactor the rest of scenarios

IvanPombo:
This PR implements the changes started in #794 and that passes most of the mechanics to the base scenario class.

Currently, to develop a scenario the user or dev will need to implement the `__init__` and the `simulate` methods. These are the core methods. Then, the `template_files_dir` should be set to the path of template files directory. These folder needs to follow the structure:
```
template_files_dir
|
|- sim_config_files
|       |- input_file1.jinja
|       |- input_file2
|
|- commands.json(.jinja)
```

All of the attributes of the scenario class that require changes in the templates should be set inside the dictionary `params`. E.g.: `self.params["attribute_1"] = attribute_1`.

This is enough to run one simple scenario. However, some scenarios require extra functionality, like the addition of extra input files that users should insert for the simulation or further configuration of the parameters.

These can be done with the following functions:
- `config_params` - Allows to configure further the params, in order to not pollute the `__init__` or `simulate` methods.
- `add_extra_input_files` - Allows to insert files in the input directory of the simulation. The structure is fully defined by the user.

In this PR, the following scenarios were adapted:
- DamBreak;
- WindTerrain;
- HeatSink;
- WindTerrain;
- MDWaterBox;
- ProteinSolvation;
- Structures;
- Stellerators;
- CoastalArea.

To confirm the changes, tests were performed for all scenarios with the scripts available on the ReadMe's. The Structures and Stellerators tests were provided by their owners @Joao97ribeiro and @joaobernardo23. Notice that in these nothing is changed, just cleaned, due to the nature of the scenarios.

There are things that could still be simplified, but I am happy with this first version! There are lots of changes, but most of them are just to add the `self.params` in the attributes and clean up the extra templating mechanics.


hpenedones (approved):
Love it when there are more deleted lines than new lines added!

PR #814, opened 2023-11-03
Reef3d is almost alive

IvanPombo:
This PR adds the client low-level call for Reef3D. 

Reef3D is called with two files "control.txt" and "ctrl.txt" which call two executables `DiveMESH` and `Reef3D`, respectively. On the backend, we are assuming that this will be always called one after the other. The first performs the meshing, the second the simulation. 

To test the simulator, we have used several tutorials examples present in https://github.com/REEF3D/REEF3D/tree/master/Tutorials. Notice that Reef3D as the ability to apply 3 types of models to simulate the hydrodynamics. This is all configured by the input files.

Further testing will occur, with green lights being presented below for a scenario of each type of model.

### DIFF
--- a/inductiva/simulators/README.md
+++ b/inductiva/simulators/README.md
@@ -165,7 +165,7 @@ import inductiva
 inductiva.api_key = "YOUR_API_KEY"
 
 # Set simulation input directory
-input_dir = "swash_example"
+input_dir = "xbeach_example"
 
 # Initialize the Simulator
 simulator = inductiva.simulators.XBeach()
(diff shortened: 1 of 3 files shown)


hpenedones (approved):
Cool, thanks Ivan!

sarmento:
I am very happy that we are moving in this direction.
I wouldn't like to approve just now for two reasons:

1) I think it is important to add a readme for this simulator, including the explanation you just gave about meshing, and some extra information about the input and output files expected.

2) I would like to see the results of running one of the tests. Ideally, we would call some visualisation tool -- even if it is not included in the API -- to make sure results are as expected.

Also: where did we submit the backend part of this? Where is the docker?

inductiva/simulators/README.md
```diff
@@ -174,6 +174,31 @@ simulator = inductiva.simulators.XBeach()
...
+A simulator call will execute the following two steps sequentially: the meshing with **DiveMESH** and the simulation with **Reef3D**. Each step is configured with an input files, `control.txt` and `ctrl.txt`, respectively. Other files may be used to inform the simulator about the grid, geographical data or wave information. Reef3D has strict naming policies for each file and we reccommend users to follow [their guidelines](https://reef3d.wordpress.com/user-guide/). Due to this, the only required argument to run the simulator is the name of the input directory with the respective input files. 
+
+Furthermore, the parallelization strategy is controlled by selecting the machine type even if stated explicitly on the input files otherwise.
+
+### Example
+
+```python
+import inductiva
+
+input_dir = "reef3d_example"
```
sarmento, line 194: Note for the future.
For next year, let's have a global dir inside inductiva bucket that all users can refer to:

```
inductiva://examples/reef3d/break_water_1
```

sarmento (approved):
LGTM!

inductiva/simulators/README.md
IvanPombo: I thought about that, and will do that for all simulators. But not next year...

PR #826, opened 2023-11-07
Add n_cores and modify assets folder

IvanPombo:
This PR adds the num of cores to the FDS simulator, due to a current constraint in the input files of generating the mesh manually.

Moreover, it corrects the path to the images in the ReadMes.

### DIFF
--- a/inductiva/coastal/README.md
+++ b/inductiva/coastal/README.md
@@ -71,7 +71,7 @@ bathymetry.plot(show=True)
 ```
 
 <p align="center">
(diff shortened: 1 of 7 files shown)


begaiym-k (commented):
Thanks @IvanPombo for fixing the READMEs!

Why can't you define the cores in the config file?

IvanPombo:
> Thanks @IvanPombo for fixing the READMEs!
> 
> Why can't you define the cores in the config file?

The mesh depends on that, and the only way to do them partition of the mesh to feed to the different cores is currently a manual job.

begaiym-k (approved):
LGTM!

PR #837, opened 2023-11-13
[FDS] Add visual parameter

IvanPombo:
This PR just serves to add a parameter to send the config file for Smokeview, the default visualizer built-in with FDS.

The bulk of the work is done in the backend. But now with this call, users can run their visuals after the simulation finishes.

### DIFF
--- a/inductiva/simulators/fds.py
+++ b/inductiva/simulators/fds.py
@@ -17,6 +17,7 @@ class FDS(Simulator):
         self,
         input_dir: types.Path,
         sim_config_filename: str,
+        post_processing_config: str = None,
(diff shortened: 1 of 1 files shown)


inductiva/simulators/fds.py
```diff
@@ -17,6 +17,7 @@ def run(
         self,
         input_dir: types.Path,
         sim_config_filename: str,
+        post_processing_config: str = None,
```
begaiym-k, line 20: Should it be named as `post_processing_config_filename` to be consistent and more explicit?

begaiym-k (approved):

PR #845, opened 2023-11-21
[Dualsph] Add commands to run method

IvanPombo:
This PR shifts the concept of passing a single simulation file for a DualSPHysics, since in fact the commands are essential to the method run in the backend. In this way, users are now able to pass the commands they want and execute the respective executables to run the desired simulations.

TLDR: Before, the backend was stuck to a single line of commands. Now, users can pass their own.

Next Step:
- Shift FluidBlock for DualSPH.
- Change the ReadMEs accordingly.

### DIFF
--- a/inductiva/simulators/dualsphysics.py
+++ b/inductiva/simulators/dualsphysics.py
@@ -1,6 +1,6 @@
 """DualSPHysics simulator module of the API."""
 
-from typing import Optional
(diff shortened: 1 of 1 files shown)


IvanPombo:
> Shouldn't you also update the scenarios here?

It will just be a question of adding a commands file, but I want to do it in a separate PR.

PR #852, opened 2023-11-23
Add unzip argument to download from url method

IvanPombo:
Based on the least astonishment principle and on giving control to the user, this PR adds an argument to unzip the file in `download_from_url` function. The previous behaviour extracted the argument automatically if the the file in the url had zip extension.

The existing test was modified to include both behaviour.

### DIFF
--- a/inductiva/molecules/utils.py
+++ b/inductiva/molecules/utils.py
@@ -1,5 +1,4 @@
 """Tools to analyze molecular dynamics simulations."""
-from typing import Optional
 
 try:
     import MDAnalysis as mda
--- a/inductiva/molecules/utils.py
+++ b/inductiva/molecules/utils.py
@@ -47,7 +47,7 @@ def align_trajectory_to_average(universe, trajectory_output_path):
                     in_memory=False).run()
 
 
-def download_pdb_from_rcsb(pdb_id: str, save_dir: Optional[str] = None) -> str:
+def download_pdb_from_rcsb(pdb_id: str) -> str:
     """Download a PDB file from the RCSB database according
     to its pdb ID.
     Args:
@@ -61,6 +61,6 @@ def download_pdb_from_rcsb(pdb_id: str, save_dir: Optional[str] = None) -> str:
     """
     pdb_url = f"https://files.rcsb.org/download/{pdb_id}.pdb"
 
-    file_path = files.download_from_url(pdb_url, save_dir)
+    file_path = files.download_from_url(pdb_url)
 
     return file_path
--- a/inductiva/utils/files.py
+++ b/inductiva/utils/files.py
@@ -117,8 +117,7 @@ def _unzip(zip_path: pathlib.Path):
     zip_path.unlink()
 
 
-def download_from_url(url: str,
-                      unzip: bool = False) -> str:
+def download_from_url(url: str, unzip: bool = False) -> str:
     """Download a file from an URL.
 
     This function downloads from a set of files from an url.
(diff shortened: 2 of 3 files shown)


inductiva/utils/files.py
```diff
@@ -130,7 +130,9 @@ def _unzip(zip_path: pathlib.Path, dest_path: pathlib.Path):
     zip_path.unlink()
 
 
-def download_from_url(url: str, save_dir: Optional[str] = None) -> str:
+def download_from_url(url: str,
+                      save_dir: Optional[str] = None,
+                      unzip: bool = False) -> str:
```
hpenedones, line 135: this is still a bit surprising for the user, I think... because the method can download any file, but there is an option that is called "unzip", which actually only applies to downloading zip files... What if I pass here "True" and it is not a zip file? I guess it will be ignored...

hpenedones: When we pass unzip = True for a zip file, what I would expect from the description is that file would be unzipped to a local folder, with all its contents. But in fact this only applies to zip files with one single file... and there is more: then that file will be renamed (not super clear to what -- the docstring mentions 'local_file_path' but that's unclear...)

hpenedones: My take: isn't this method trying to do too many things?
Feels like it should only download a file, that's it. Or if we pass an unzip option, it should not be surprising: just decompress it all, independently of number of files, and do not do any renaming.

Seems like this method is doing all this, because historically we wanted to use it in some snippets of documentation and we want the code there to be simple and compact, but that's not a strong reason enough, to create a method that is super "entangled" with many behaviours inside it... 
Probably we just need a simpler way to store example files of the API, but not implement a method with surprising behaviours. 
What do you think?

IvanPombo: I see your points! I will look into this and make it simpler and not so entangled.

inductiva/molecules/utils.py
```diff
@@ -47,7 +46,7 @@ def align_trajectory_to_average(universe, trajectory_output_path):
                     in_memory=False).run()
 
 
-def download_pdb_from_rcsb(pdb_id: str, save_dir: Optional[str] = None) -> str:
+def download_pdb_from_rcsb(pdb_id: str) -> str:
```
begaiym-k, line 49: Why did you remove save_dir?

If so you need to remove it in the docstring as well.

inductiva/utils/files.py
```diff
@@ -108,80 +107,41 @@ def get_alphanum_key(file):
...
-            zip_ref.extractall(dest_path.parent)
-
-    # Rename the file to dest_path
+        zip_ref.extractall()
+
     zip_path.unlink()
 
 
-def download_from_url(url: str, save_dir: Optional[str] = None) -> str:
+def download_from_url(url: str, unzip: bool = False) -> str:
```
begaiym-k, line 119: Just curious, Is there any reason why the users would not want to unzip it right away?

begaiym-k (approved):
Thanks @IvanPombo! I left couple of comments, other than that LGTM!

inductiva/molecules/utils.py
IvanPombo: It will be by default, to not add extra complexity to the download.

inductiva/utils/files.py
IvanPombo: This can be used for several things, and the unzip tells they users they are actually uncompressing.

PR #866, opened 2023-11-28
Add list of machines and termination to CLI 

IvanPombo:
This PR attempts to address one hole of the script centric API: managing machines via the CLI.

We add two new functions to the CLI:
- List available machines to launch;
- Terminate machines.

In this way, before launching a python script users can check which machines are available to insert in their Python script, by doing on their terminal:
```inductiva resources available```
Outputs:
Screenshot 2023-11-28 at 14 39 34

Further, when submitting a bunch of tasks and coming back later to check if everything is okay and end the machines, users would need a separate python script to fetch the machine and then kill it. To make this easier, users can just kill their machines from terminal as well.

## How to use it:
Find your machine name (already implemented):
```inductiva resources list```

Then, with the name they want to kill:
```inductiva resources terminate -n {api-machine-name}```

## Concerns
- The code to get the available resources should come from the backend, due to possible updating of new images (not urgent).
- Overlapping abilities between the API and the console, but I guess it is better to have them both available. 
- The code available here should be in the general package and just called from here.

Closes issue #859.

### DIFF
--- a/inductiva/_cli/resources.py
+++ b/inductiva/_cli/resources.py
@@ -57,13 +57,13 @@ def register_resources_cli(parser):
         "list", help="List currently active resources")
     available_subparser = subparsers.add_parser(
         "available", help="List available machine types")
-    terminate_subparser = subparsers.add_parser("terminate",
-                                                help="Terminate a machine")
+    terminate_subparser = subparsers.add_parser(
+        "terminate", help="Terminate a machine-group")
 
-    terminate_subparser.add_argument("-n",
-                                     "--machine-name",
-                                     type=str,
-                                     help="Name of the machine to terminate")
+    terminate_subparser.add_argument(
+        "machine-group-name",
+        type=str,
+        help="Name of the machine group to terminate")
 
     # Register function to call when this subcommand is used
     list_subparser.set_defaults(func=list_resources)
--- a/inductiva/_cli/resources.py
+++ b/inductiva/_cli/resources.py
@@ -48,6 +48,7 @@ def terminate_machine(args):
 
     print(f"Machine {machine_name} not found.")
 
+
 def register_resources_cli(parser):
     _cli.utils.show_help_msg(parser)
     subparsers = parser.add_subparsers()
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: inductiva/_cli/resources.py [hunk 1]
```diff
@@ -9,12 +9,63 @@ def list_resources(args):
...
+        "n1-standard-": [1, 2, 4, 8, 16, 32, 64, 96]
+    }
+
+    print("Available machine types\n")
+    print("machine-type: [cores-available]")
+    for machine_type, cores in machines_dict.items():
+        cores_str = ", ".join(str(core) for core in cores)
+        print(f"{machine_type}: [{cores_str}]")
+
+    print("\n E.g. of machine: c2-standard-8\n")
```
begaiym-k, line 33: a machine type*?

inductiva/_cli/resources.py
```diff
@@ -9,12 +9,63 @@ def list_resources(args):
...
+
+
+def terminate_machine(args):
+    """Terminate a machine with the given name."""
+    machine_name = args.machine_name
+
+    print("Terminating machine... If exists.")
+    machines_list = inductiva.resources.machine_groups.get()
+
+    for machine in machines_list:
```
begaiym-k, line 43: This is one way of doing it, or if we want to avoid iterating over the machines and retrieving all of them, you could use [this](https://github.com/inductiva/inductiva/blob/fd8a213786328fa8c939e45ec1526aadb160fa9e/inductiva/resources/resources.py#L33) that I initially wrote to retrieve a single machine by it's name. The endpoint is still there so you could just implement it here in the client side.

inductiva/_cli/resources.py
```diff
@@ -9,12 +9,63 @@ def list_resources(args):
...
     subparsers = parser.add_subparsers()
 
     list_subparser = subparsers.add_parser(
         "list", help="List currently active resources")
+    available_subparser = subparsers.add_parser(
+        "available", help="List available machine types")
+    terminate_subparser = subparsers.add_parser(
+        "terminate", help="Terminate a machine-group")
+
+    terminate_subparser.add_argument(
```
begaiym-k, line 63: maybe in the command you could do machines instead of resources, and here instead of machine-group-name, which seems quite long, it could be just name.

inductiva/_cli/resources.py
IvanPombo: The problem is that I think there is no arg/flag in the backend to differentiate between elastic or non-elastic.
Given that the purpose is to just kill the machine, I will keep it this way and in the future we can re-think the whole machine_groups with the idea of coupling things to it.

begaiym-k (approved):
LGTM! Excited to try it out later!

PR #868, opened 2023-11-28
Move estimate cloud cost to an upper layer

IvanPombo:
This PR moves the computation of the cloud cost to an upper layer, i.e., users don't need to register a machine group or create the object, just to get the cost of that said machine.

In this way, we pass the core of the cost computation - an api request - to the `machine_groups.py` and call it as well inside the MachineGroup class for the estimation of the MachineGroup cost.

The advantage is that now users can do: 
```python
inductiva.resources.estimate_machine_cost(machine_type="c2-standard-60", zone="europe-west1-b", spot=False)
```

To get the cost of a single machine of that type. Accordingly, and to facilitate the usage, we add a method to the CLI.

### DIFF
--- a/inductiva/resources/machine_groups.py
+++ b/inductiva/resources/machine_groups.py
@@ -6,6 +6,34 @@ from inductiva.utils import format_utils
 from inductiva import resources
 
 
+def estimate_machine_cost(machine_type: str,
+                          zone: str = "europe-west1-b",
+                          spot: bool = False):
+    """Estimate the cloud cost of one machine per hour in US dollars.
+    
+    Args:
+        machine_type: The type of GC machine to launch. Ex: "e2-standard-4".
+            Check https://cloud.google.com/compute/docs/machine-resource for
+            more information about machine types.
+        zone: The zone where the machines will be launched.
+        spot: Whether to use spot machines.
+    """
+
+    api = compute_api.ComputeApi(inductiva.api.get_client())
+
+    instance_price = api.get_instance_price({
+        "machine_type": machine_type,
+        "zone": zone,
+    })
+
+    if spot:
+        estimated_cost = instance_price.body["preemptible_price"]
+    else:
+        estimated_cost = instance_price.body["on_demand_price"]
+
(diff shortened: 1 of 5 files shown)


IvanPombo:
Tagging @AugustoPeres for awareness since this was a bit of his request with an issue open in the backend. Let me know what you think!

AugustoPeres:
> Tagging @AugustoPeres for awareness since this was a bit of his request with an issue open in the backend. Let me know what you think!

Thanks Ivan, This will be very useful

inductiva/resources/machine_groups.py
```diff
@@ -6,6 +6,34 @@
 from inductiva import resources
 
 
+def estimate_machine_cost(machine_type: str,
```
begaiym-k, line 9: what if you add a parameter "num_machines" here, because then it will make the code in other scripts shorter, there was one moment where you first looked up if the cost was already estimated, I am not sure if that's the right approach, because the user could have changed the machine type along the way, so it might be better to always retrieve the cost according to the latest info.

begaiym-k:
Are you intentionally pushing this to main branch? Because there is a high chances of conflicts when we merge dev to main.

IvanPombo:
> Are you intentionally pushing this to main branch? Because there is a high chances of conflicts when we merge dev to main.

Upsy!

inductiva/resources/machine_groups.py
IvanPombo: However, note that the Elastic Machine Group uses the single machine price to compute the lower and upper bound. So you would need to call the API twice to compute that. I guess that is the main reason to keep this structure.

begaiym-k: > For your second part, I think that is not true, since that is on register moment, so if people change stuff I think it will not be registered again, or will it?

it will still register it.

IvanPombo: Actually, if you change any variable after initialising the class and then try to start the machines this will crash.
Which can make sense in some way, but then, the start() method seems to be relatively pointless.

PR #897, opened 2023-12-15
Add mixins and respective Tests

IvanPombo:
In this PR, we add the mechanics to template files via `mixins.FileManager` that can be plugged in into your class to add the functionalities of rendering files and directories.

A suit of tests were added to verify that the behaviour is according to our expectations.


inductiva/mixins/file_manager.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if root_dir is None:
+            logging.info("Root directory not set. "
+                         "Setting default root_dir to be `input_dir`.")
+            root_dir = f"input_dir-{misc.create_random_tag()}"
+        elif os.path.isdir(root_dir):
+            logging.info(
+                "Directory %s already exists."
+                " Adding random tag to directory name.", root_dir)
+            root_dir = f"{root_dir}-{misc.create_random_tag(size=5)}"
```
sergiomsantos, line 29: random tags are hard to read. People like to have a sense of sequence and this is why I suggest adding a sequential number.

This is the kind of logic we need:

```python
def gen_suffix(name, filenames):
    regex = f'^{name}#([0-9]+$)'

    max_suffix = -1
    for filename in filenames:
        if match := re.match(regex, filename):
            max_suffix = max(max_suffix, int(match.group(1)))

    exists_solo = name in filenames
    exists_derived = max_suffix >= 0

    if exists_derived:
        suffix = f'#{max_suffix+1}'
    elif exists_solo:
        suffix = '#2'
    else:
        suffix = ''
    return suffix

def gen_name(name):
    name = name.strip()
    filenames = os.glob.glob(f'{name}*')
    return name + gen_suffix(name, filenames)
```

```suggestion
        elif os.path.isdir(root_dir):
            new_root_dir = gen(root_dir)
            logging.info(
                "Directory %s already exists."
                " Setting root dir to %s", root_dir, new_root_dir)
            root_dir = new_root_dir"
```

(naming might no be the most adequate).

inductiva/utils/misc.py
```diff
@@ -1,8 +1,18 @@
...
     return re.sub("([A-Z][a-z]+)", r" \1", re.sub("([A-Z]+)", r" \1",
                                                   s)).split()
+
+
+def create_random_tag(size: int = 4):
+
+    samples = [random.choice(string.ascii_lowercase) for _ in range(size)]
+    tag = "".join(samples)
+
+    return tag
```
sergiomsantos, line 18: no necessary. random tags are hard to follow

inductiva/utils/templates.py
```diff
@@ -1,83 +1,74 @@
...
-    template_dir = template_path.parent
-    template_filename = template_path.name
+    return string
 
-    environment = Environment(loader=FileSystemLoader(template_dir))
-    template = environment.get_template(template_filename)
-    stream = template.stream(**params)
-    stream.dump(output_file)
+
+def render_file(source_file, target_file, remove_template=False, **render_args):
```
sergiomsantos, line 25: why should we have the `remove_template` argument? I thin it doesn't make sense to remove templates

inductiva/utils/templates.py
```diff
@@ -1,83 +1,74 @@
...
+
+    if remove_template_dir:
+        shutil.rmtree(source_dir)
+
+
+def get_jinja_files(src_dir):
+    """Get all jinja files in a directory."""
+
+    jinja_paths = glob.glob(os.path.join(src_dir, "**"
+                                         "*.jinja"),
```
sergiomsantos, line 67:
```suggestion
    template_paths = glob.glob(os.path.join(src_dir, "**", TEMPLATE_EXTENSION),
```

inductiva/utils/templates.py
```diff
@@ -1,83 +1,74 @@
...
+        jinja_path = os.path.join(target_dir, jinja_file)
+        target_path = jinja_path.split(".jinja")[0]
+
+        render_file(source_file=jinja_path,
+                    target_file=target_path,
+                    remove_template=True,
+                    **render_args)
+
+    if remove_template_dir:
+        shutil.rmtree(source_dir)
```
sergiomsantos, line 60: let's not couple variable names to rendering tech.
let's have them named `template_*` rather than `jinja_*`. The only place jinja should be enunciated in where we actually perform the rendering.

Also, the template extension should be defined as a module constant so that we can change it whenever we want with just a single line change.

inductiva/utils/templates.py
```diff
@@ -1,83 +1,74 @@
...
+
+
+def get_jinja_files(src_dir):
```
sergiomsantos, line 63:
```suggestion
def get_template_files(src_dir):
```

inductiva/utils/templates.py
```diff
@@ -1,83 +1,74 @@
...
+
+    source_path = pathlib.Path(source_file)
+
+    source_dir = source_path.parent
+    source_file = source_path.name
+
+    environment = jinja2.Environment(loader=jinja2.FileSystemLoader(source_dir))
+    template = environment.get_template(source_file)
+    stream = template.stream(**render_args)
+    stream.dump(target_file)
```
sergiomsantos, line 35: is the stream properly disposed after this point?

inductiva/utils/templates.py
```diff
@@ -1,83 +1,74 @@
...
-from typing import Dict, List, Union
-from pathlib import Path
+import shutil
 
-from jinja2 import Environment, FileSystemLoader
+import jinja2
 
 from inductiva.utils.files import find_path_to_package
 
 TEMPLATES_PATH = find_path_to_package("templates")
```
sergiomsantos, line 12: is this needed?

inductiva/tests/assets/template_dir/template.json.jinja
```diff
@@ -0,0 +1,4 @@
+{
+    "simulator": "{{ simulator }}",
+    "simulator_version": "{{ simulator_version }}"
+}
```
sergiomsantos, line 4: missing line break

inductiva/tests/mixins/template_file.txt.jinja
```diff
@@ -0,0 +1 @@
+This is a test file, which {{ text }} is a jinja file!
```
sergiomsantos, line 1: missing line break

inductiva/tests/mixins/test_command_manager.py
```diff
@@ -0,0 +1,45 @@
+"""Tests the command manager mixin."""
```
sergiomsantos, line 1: remove file

inductiva/utils/templates.py
IvanPombo: The remove template is because in the `add_dir` we copy the files first and then iterate over them to template (to also pass all the other files that aren't jinja). Hence, at the end we remove the .jinja template files.

inductiva/mixins/file_manager.py
```diff
@@ -0,0 +1,94 @@
...
+    def __check_root_dir(self):
+        if self.__root_dir is None:
+            raise ValueError("Root directory not set.")
+
+    def set_root_dir(self, root_dir=None):
+        """Set the root directory for the manager."""
+
+        if root_dir is None:
+            logging.info("Root directory not set. "
+                         "Setting default root_dir to be `input_dir`.")
```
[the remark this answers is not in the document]
sergiomsantos: `None` is because it has not yet been initialised. We can do it because we can. Users should not be able to uninitialise something that they are trying to initialise. They should know what they want to do when they call the method and not hand out the decision to the devs. If that is the case, they shouldn't be trying to set the root dir.

inductiva/utils/templates.py
IvanPombo: I don't find anything about the stream needing to be closed. But I am going to go over it.

inductiva/utils/templates.py
IvanPombo: Not anymore.

inductiva/mixins/file_manager.py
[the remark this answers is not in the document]
IvanPombo: Okay, that seems like a good pattern to keep in mind! Thanks for the explanation!

inductiva/mixins/file_manager.py
sergiomsantos: BTW, the reason I extract the `gen_suffix` logic to a separate method is because we can easily test `gen_suffix` without having to work with the file system. It is easily testable, whereas `gen_name` requires real folders in the fs. This is a good pattern to keep in mind when testing is important.

inductiva/utils/templates.py
IvanPombo: If users want to a stream dump again, it doesn't provide anything.

By the documentation: http://mitsuhiko.pocoo.org/jinja2docs/Jinja2.pdf it doesn't mention anyway to close the stream - presumably it is automatic.

inductiva/tests/mixins/test_file_manager.py
```diff
@@ -0,0 +1,201 @@
...
+    if render_args are passed. Further, if render_args doesn't contain all
+    arguments that are on the template files, then those extra arguments are 
+    rendered with None. We leave this to the responsability of the user.
+    """
+    file_manager = mixins.FileManager()
+    target_dir = file_manager.add_dir(source_dir, target_dir, **render_args)
+
+    root_dir = file_manager.get_root_dir()
+    target_path = os.path.join(root_dir, target_dir)
+    assert os.path.isdir(target_path)
```
sergiomsantos, line 195: this is testing `get_root_dir` but the test is testing `add_dir`.
This assertion should be moved to it's own test.

inductiva/tests/mixins/test_file_manager.py
```diff
@@ -0,0 +1,201 @@
...
+
+    expected_render_files = [
+        file.split(mixins.file_manager.TEMPLATE_EXTENSION)[0]
+        for file in os.listdir(TEST_DIR)
+    ]
+    assert sorted(expected_render_files) == sorted(os.listdir(target_path))
```
sergiomsantos, line 201: works but is not enough.
Say that the content of `TEST_DIR` got deleted while copying and the `add_dir` method messed and deleted the copied content at the end (or some variation of the sort). In the end, both directories are empty and the assertion if `True` but the method failed to do the job!
We need some additional assertion to test this. Something along:

```suggestion
    assert len(expected_render_files) > 0
    assert sorted(expected_render_files) == sorted(os.listdir(target_path))
```

inductiva/tests/mixins/test_file_manager.py
```diff
@@ -0,0 +1,201 @@
...
+    assert os.listdir(TEST_DIR) == os.listdir(target_path)
+
+
+@pytest.mark.parametrize("source_dir, render_args, target_dir", [(TEST_DIR, {
+    "text": "yes",
+    "simulator": "openfoam"
+}, None), (TEST_DIR, {
+    "text": "yes",
+    "simulator": "openfoam"
+}, "rendered_dir")])
```
sergiomsantos, line 176: hard to read:

```suggestion
@pytest.mark.parametrize("source_dir, render_args, target_dir", [
    (TEST_DIR, {"text": "yes", "simulator": "openfoam"}, None),
    (TEST_DIR, {"text": "yes", "simulator": "openfoam"}, "rendered_dir")])
```

inductiva/tests/mixins/test_file_manager.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+
+    file_manager = mixins.FileManager()
+    target_file = file_manager.add_file(source_file, target_file)
+
+    root_dir = file_manager.get_root_dir()
+    assert os.path.isfile(os.path.join(root_dir, target_file))
+
+
+@pytest.mark.parametrize("source_file, render_args, target_file",
```
sergiomsantos, line 124: to prevent very large indentantions, why not import `parametrize` independently?
We have very small chances for name clashing and the function name is clearly understandable in the test context.

inductiva/tests/mixins/test_file_manager.py
```diff
@@ -0,0 +1,201 @@
...
+from inductiva import mixins
+
+TEST_DIR = os.path.join(os.path.dirname(__file__), "assets")
+
+
+@pytest.mark.parametrize("name, listnames, expected_output",
+                         [("test", ["inductiva", "."], "test"),
+                          ("test", ["inductiva", "test"], "test#2"),
+                          ("test", ["inductiva", "test#2"], "test#3"),
+                          ("test", ["inductiva", "test-2"], "test")])
```
sergiomsantos, line 16: imagine `gen_suffix` is hard coded to deal with the `"test"` string. It might be reasonable to have different test fixtures.

inductiva/mixins/file_manager.py
```diff
@@ -0,0 +1,208 @@
...
+        Args:
+            root_dir: Path to the root directory.
+                If None, an error is raised.
+                If the directory already exists. A new directory
+                name is created following a sequence of #1, #2, #3, etc."""
+
+        if root_dir is None:
+            raise ValueError("Given root directory cannot be None")
+        elif os.path.isdir(root_dir):
+            generated_root_dir = gen_name(root_dir)
```
sergiomsantos, line 44:
```suggestion
        elif os.path.isdir(root_dir):
            if os.getenv('INDUCTIVA_DISABLE_FILEMANAGER_AUTOSUFFIX', False):
                raise FileExistsError(f'Directory {root_dir} already exists.')
            generated_root_dir = gen_name(root_dir)
```

sergiomsantos (commented):
Looking good!
Some comments to make code more resilient.

inductiva/__init__.py
```diff
@@ -5,6 +5,7 @@
 import absl
 
 from . import api
+from . import mixins
```
sergiomsantos, line 8: no sure the mixins should be imported here 🤔

inductiva/utils/templates.py
sergiomsantos: Thanks.
Yes, the stream is closed at the end (see [here](https://github.com/pallets/jinja/blob/d594969d722ceb4e8f3da8861befc9c0ac87ae1b/src/jinja2/environment.py#L1626)).

inductiva/mixins/file_manager.py
IvanPombo: Shall this be set with Inductiva package, like, `inductiva.filemanager_autosuffix = True`, as the other env variables we have?

inductiva/__init__.py
IvanPombo: TBH, I usually like to just do `import inductiva` and go from it to all the main functionalities. But ain't sure about design, since mixins are not used for everything. So I see your point.

inductiva/mixins/file_manager.py
IvanPombo: Wondering about this further... Shouldn't we allow override as well? I don't see the point of disable just to raise an error. Could you elaborate on this?

inductiva/tests/mixins/test_file_manager.py
IvanPombo: The bare minimum is from pytest import mark (exactly the same workflow if we don't set `from . import mixins` in the init.

IvanPombo:
Thanks @sergiomsantos for all the input gonna push the remaining changes now, my only open question is the env variable.

Moreover, I have tested passing a directory "folder_1/folder_2" with files inside into a directory "target_1/target_2/target_3" and everything worked wondefully, with the files and folders inside `folder_2` being passed to `target_3`.

inductiva/__init__.py
sergiomsantos: let's be explicit here and remove the mixins from the main namespace.
users will have to explicitly import the mixin.

inductiva/mixins/file_manager.py
```diff
@@ -0,0 +1,230 @@
...
+    def set_root_dir(self, root_dir=None):
+        """Set a root directory for the file manager.
+        
+        All files managed through the manager will be
+        placed inside a newly created folder with the given
+        name. If a folder with the same name already exists,
+        the newly created one will be appended with the
+        "#1" suffix. If folders having the "#N" suffix already
+        exist, the given name will be appended with "#{N+1}"
+        where N is determined by the largest existing N.
```
sergiomsantos, line 44:
```suggestion
        "__2" suffix. If folders having the "__N" suffix already
        exist, the given name will be appended with "__{N+1}"
        where N is determined by the directory with the largest N.
```

inductiva/mixins/file_manager.py
```diff
@@ -0,0 +1,230 @@
...
+
+        Args:
+            root_dir: Path to the root directory.
+                If None, an error is raised.
+                If the directory already exists. A new directory
+                name is created following a sequence of #1, #2, #3, etc."""
```
sergiomsantos, line 50:
```suggestion
        Args:
            root_dir: Path to the root directory.
                If None, an error is raised."""
```

inductiva/mixins/file_manager.py
```diff
@@ -0,0 +1,230 @@
...
+
+        if root_dir is None:
+            raise ValueError("Given root directory cannot be None")
+        elif os.path.isdir(root_dir):
+            if os.getenv("INDUCTIVA_DISABLE_FILEMANAGER_AUTOSUFFIX", False):  #pylint: disable=invalid-envvar-default
```
sergiomsantos, line 55: There's a bug here (my bad).
env vars are always str. we need to convert correctly to a boolean.

in `inductiva.utils`, add a helper method to get boolean env vars:

```python
from distutils.util import strtobool

def getenv_bool(varname, default):
    return bool(strtobool(os.getenv(varname, default)))
```


```suggestion
# no need to pylint disable
            if utils.getenv_bool("INDUCTIVA_DISABLE_FILEMANAGER_AUTOSUFFIX", False):
```

inductiva/mixins/file_manager.py
```diff
@@ -0,0 +1,230 @@
...
+    stream = template.stream(**render_args)
+    stream.dump(target_file)
+
+
+def get_template_files(src_dir):
+    """Get all template files in a directory."""
+
+    template_paths = glob.glob(os.path.join(src_dir, "**",
+                                            "*" + TEMPLATE_EXTENSION),
+                               recursive=True)
```
sergiomsantos, line 180:
```suggestion
    template_paths = glob.glob(
        os.path.join(src_dir, "**", "*" + TEMPLATE_EXTENSION),
        recursive=True)
```

inductiva/__init__.py
```diff
@@ -14,6 +15,9 @@
 api_url = os.environ.get("INDUCTIVA_API_URL", "https://api.inductiva.ai")
 output_dir = os.environ.get("INDUCTIVA_OUTPUT_DIR", "inductiva_output")
 api_key = os.environ.get("INDUCTIVA_API_KEY")
+disable_filemanager_autosuffix = os.environ.get(
+    "INDUCTIVA_DISABLE_FILEMANAGER_AUTOSUFFIX", False)
+
```
sergiomsantos, line 20: no need to have this here because it can be dynamically set (via `os.setenv`). In that case, `disable_filemanager_autosuffix` would be out of sync.

inductiva/mixins/file_manager.py
```diff
@@ -0,0 +1,230 @@
...
+
+        if target_file is None:
+            new_target_file = pathlib.Path(source_file).name
+        else:
+            new_target_file = target_file
+
+        new_target_file = os.path.join(self.__root_dir, new_target_file)
+        if source_file.endswith(TEMPLATE_EXTENSION):
+            if target_file is None:
+                new_target_file = new_target_file.split(TEMPLATE_EXTENSION)[0]
```
sergiomsantos, line 101: will break. example:

```python
>>> TEMPLATE_EXTENSION='txt'
>>> new_target_file = 'a.b.c.txt.txt.txt'
>>> new_target_file.split(TEMPLATE_EXTENSION)[0]
'a.b.c'
```

use:
```suggestion
                new_target_file, _ = os.path.splitext(new_target_file)
```

inductiva/tests/mixins/test_file_manager.py
```diff
@@ -0,0 +1,224 @@
...
+    assert suffix_tag == expected_output
+
+
+@mark.parametrize("source_file, render_args, expected_output",
+                  [(path("template_file.txt.jinja"), {
+                      "text": "yes"
+                  }, "This is a test file, which yes is a jinja file!"),
+                   (path("template.yaml.jinja"), {
+                       "simulator": "openfoam"
+                   }, "simulator: openfoam")])
```
sergiomsantos, line 47: can we format this with less line breaking?

inductiva/tests/mixins/test_file_manager.py
```diff
@@ -0,0 +1,224 @@
...
+    assert os.listdir(ASSETS_DIR) == os.listdir(target_path)
+
+
+@mark.parametrize("source_dir, render_args, target_dir", [(ASSETS_DIR, {
+    "text": "yes",
+    "simulator": "openfoam"
+}, None), (ASSETS_DIR, {
+    "text": "yes",
+    "simulator": "openfoam"
+}, "rendered_dir")])
```
sergiomsantos, line 198: less line breaking?

sergiomsantos (changes requested):
we have 2 bugs and one inconsistency.
Other than that, all looks good.

inductiva/mixins/file_manager.py
IvanPombo: No luck! We have to change yapf, to avoid this problems. But that would move away from google style, so a more "general" discussion of code formatting is required so that every repo is uniform.

inductiva/tests/mixins/test_file_manager.py
IvanPombo: Same as above..

IvanPombo:
Address everything, except to the line breaking questions, since yapf complains!

sergiomsantos (approved):

PR #923, opened 2024-01-09
Adds automatic simulation metadata logging

AugustoPeres:
inductiva/tasks#18

## Contents

This pull request add the option to automatically save metadata whenever a simulation is ran.
This behaviour can be deactivated by setting `DISABLE_TASK_METADATA_LOGGING=True`.
It also saves that metadata to the `input_dir` so that the metadata is uploaded when the task is submited.

Expected behaviour:

```python
import inductiva

    "splishsplash-input-example.zip",
    unzip=True)

splishsplash = inductiva.simulators.SplishSplash()

task = splishsplash.run(input_dir=input_dir,
                        extra_metadata={"extra_metadata_1": 1234})
```

Should yield:

```
2024-01-09 16:20:02 INFO Simulation metadata logged to: /path/to/cwd/1704817201870341731/task_metadata.json
```

Where `/path/to/cwd/1704817201870341731/sim_metadata.json` contains:

```
{"task_id": "1704817201870341731", "machine_group_id": null, "input_dir": "/Users/augustoperes/Documents/inductiva/inductiva_test_scripts/splishsplash-input-example", "storage_dir": "", "input_filename": "config.json", "particle_radius": 0.025, "extra_metadata_1": 1234}
```

## Testing

I tested locally for all simulators, with different `save_sim_metadata_dir` and `extra_sim_metadata_to_save` arguments. And everything seems fine.
I also experimented with `DISABLE_TASK_METADATA_LOGGING=True` and it behaved as expected.
Unit tests where added in this pull-request.

To test it yourself please install from the branch `ap-add-metadata-logs`

### DIFF
--- a/inductiva/tests/tasks/test_run_task.py
+++ b/inductiva/tests/tasks/test_run_task.py
@@ -29,10 +29,10 @@ def _id_in_metadata_file(task_id):
     return task_id in map(lambda x: x["task_id"], json_list)
 
 
-@pytest.mark.parametrize("task_id,disable_logging", [("id_1", True),
-                                                     ("id_2", False),
-                                                     ("id_3", True),
-                                                     ("id_4", False)])
+@pytest.mark.parametrize("task_id,disable_logging", [(".id_1", True),
+                                                     (".id_2", False),
+                                                     (".id_3", True),
+                                                     (".id_4", False)])
 def test_run_simulation_logging(task_id, disable_logging):
     """Tests if the id of the task was added to the file."""
     os.chdir(TEMP_DIR.name)


inductiva/tasks/run_simulation.py
```diff
@@ -39,4 +45,43 @@ def run_simulation(
...
+    if format_utils.getenv_bool("SAVE_SIM_METADATA", True):
+        metadata = {
+            **{
+                "machine_group_id": resource_pool_id,
+                "task_id": task_id,
+                "input_dir": str(input_dir),
+                "storage_dir": str(storage_dir),
+            },
+            **kwargs,
+        }
```
sergiomsantos, line 57:
```suggestion
        metadata = {
            **{
                "task_id": task_id,
                "input_dir": str(input_dir),
                "storage_dir": str(storage_dir),
                "api_method_name": api_method_name,
                "machine_group_id": resource_pool_id,
            },
            **kwargs,
        }
```

inductiva/tasks/run_simulation.py
```diff
@@ -1,16 +1,22 @@
...
+from inductiva.utils import format_utils
 
 
 def run_simulation(
     api_method_name: str,
     input_dir: pathlib.Path,
     machine_group: Optional[resources.MachineGroup] = None,
     storage_dir: Optional[types.Path] = "",
+    save_sim_metadata_dir: Optional[types.Path] = None,
+    extra_sim_metadata_to_save: Optional[dict] = None,
```
sergiomsantos, line 19: for now let's not have any additional argument to these methods.
Let's assume that they get called with everything that is strictly required. If it becomes imperative to have additional, custom, metadata fed into these calls, then we'll look into it.

inductiva/tasks/run_simulation.py
```diff
@@ -39,4 +45,43 @@ def run_simulation(
...
+            **kwargs,
+        }
+        if extra_sim_metadata_to_save is not None:
+            metadata = {**metadata, **extra_sim_metadata_to_save}
+        _save_metadata(save_sim_metadata_dir, metadata)
+
     return task
+
+
+def _save_metadata(save_sim_metadata_dir, metadata):
```
sergiomsantos, line 65: let's always save this metadada to a file in the current working directory that only gets appended.
The ame of the file can be defined using a module specific global variable.

inductiva/tasks/run_simulation.py
```diff
@@ -39,4 +45,43 @@ def run_simulation(
...
+
+    task_id = metadata["task_id"]
+    save_sim_metadata_dir /= f"{task_id}"
+    save_sim_metadata_dir.mkdir(parents=True, exist_ok=True)
+    file_path = save_sim_metadata_dir / "sim_metadata.json"
+
+    with open(file_path, "w", encoding="utf-8") as f:
+        json.dump(metadata, f)
+
+    logging.info("Simulation metadata logged to: %s", file_path)
```
sergiomsantos, line 87: This can be simplified because we're appending to the same file and the filename is hard coded.

```suggestion
    TASK_METADATA_FILENAME = "task_metadata.json"
    file_path = pathlib.Path().cwd() / TASK_METADATA_FILENAME

    with open(file_path, "a", encoding="utf-8") as f:
        json.dump(metadata, f)
```

this will produce a file like:

```json
# tmp.json
{"a": "1", "b": "2"}
{"a": "3", "b": "4"}
```

that can easily be manipulated with `jq`:

```bash
$ jq '.a' tmp.json
"1"
"3"
```

inductiva/simulators/dualsphysics.py
```diff
@@ -19,6 +19,8 @@ def run(
         commands: List[dict],
         machine_group: Optional[resources.MachineGroup] = None,
         storage_dir: Optional[types.Path] = "",
+        save_sim_metadata_dir: Optional[types.Path] = None,
+        extra_sim_metadata_to_save: Optional[dict] = None,
```
sergiomsantos, line 23: for now let's not have any additional argument to these methods.
Let's assume that they get called with everything that is strictly required. If it becomes imperative to have additional, custom, metadata fed into these calls, then we'll look into it.

sergiomsantos (changes requested):
I have some suggestion to make the code a bit leaner.
Also, we don't have any unit test to make sure the metadata is actually written. Please add them (you will need to mock calls to `methods.invoke_async_api`). I can help you with that.

Also, there is no need (at least for now) to have additional metadata tags passed around. This will significantly simplify the code and give you a bit of leeway to add unit tests.

IvanPombo (commented):
Just wanted to highlight to @sergiomsantos the pain that I was talking with you on Monday whenever we need to change anything in the simulators - it is painful and these arguments are not specific of the simulator configuration.

Anyway, I agree with the comments suggested and don't have anything else to add :D

inductiva/tasks/__init__.py
```diff
@@ -1,4 +1,4 @@
 #pylint: disable=missing-module-docstring
 from inductiva.tasks.task import Task
-from inductiva.tasks.run_simulation import run_simulation
+from inductiva.tasks.run_simulation import run_simulation, TASK_METADATA_FILENAME
```
IvanPombo, line 3: Given this change, let's match this `__init__.py` file as the other ones, and just do `from .task.run_simulation`

inductiva/tasks/run_simulation.py
```diff
@@ -39,4 +45,26 @@ def run_simulation(
...
+            **kwargs,
+        }
+        _save_metadata(metadata)
+
     return task
+
+
+def _save_metadata(metadata):
+    """Appends metadata to the TASK_METADATA_FILENAME in the cwd."""
+    file_path = pathlib.Path().cwd() / TASK_METADATA_FILENAME
```
IvanPombo, line 66: Here you should use the `inductiva.utils.files.resolve_path` which does this and/or sets in the desired `inductiva.working_dir` established by the user.

inductiva/tasks/run_simulation.py
```diff
@@ -39,4 +45,26 @@ def run_simulation(
...
+    with open(file_path, "a", encoding="utf-8") as f:
+        json.dump(metadata, f)
+        f.write("\n")
+    logging.info("Simulation metadata logged to: %s", file_path)
```
IvanPombo, line 70: This is unnecessary now, since all metadata is always saved to the same file in the current working directory. As long as it will be explained in the documentation it is enough for users to grasp things.

inductiva/tests/tasks/test_run_task.py
```diff
@@ -0,0 +1,51 @@
...
+
+def test_run_simulation_logging(mocker):
+    """Tests if the id of the task was added to the file."""
+
+    # Old logs may cause problems so we remove the file.
+    if os.path.exists(tasks.TASK_METADATA_FILENAME):
+        os.remove(tasks.TASK_METADATA_FILENAME)
+
+    invoke_api_mock = mocker.patch("inductiva.api.methods.invoke_async_api")
+    mocker.patch("inductiva.tasks.Task")
```
IvanPombo, line 28: Why is this patch necessary here?

inductiva/tests/tasks/test_run_task.py
```diff
@@ -0,0 +1,51 @@
...
+    with open(tasks.TASK_METADATA_FILENAME, "r", encoding="utf-8") as f:
+        json_list = [json.loads(line) for line in f]
+    return task_id in map(lambda x: x["task_id"], json_list)
+
+
+def test_run_simulation_logging(mocker):
+    """Tests if the id of the task was added to the file."""
+
+    # Old logs may cause problems so we remove the file.
+    if os.path.exists(tasks.TASK_METADATA_FILENAME):
```
IvanPombo, line 24: Maybe we should mock the file as well to avoid this. In this case, we. An start from a barebones file and not have problems with what was previously appended.

inductiva/tests/tasks/test_run_task.py
```diff
@@ -0,0 +1,51 @@
...
+                               id_exists_in_file=True)
+    _run_simulation_and_assert("some_api_method",
+                               input_dir,
+                               invoke_api_mock,
+                               NOT_LOGGED_TASK_ID,
+                               disable_logging=True,
+                               id_exists_in_file=False)
+
+
+def _run_simulation_and_assert(api_method_name, input_dir, invoke_api_mock,
```
IvanPombo, line 46: Instead of creating an helper function for this purpose, we should parameterize the test for the different possibilities. In this way, some of the variables are set in during the test method and not passed as arguments, while the ones that are, are really the variations we want to test. Check the `test_file_manager.py` for inspiration on that.

inductiva/tests/tasks/test_run_task.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def _id_in_metadata_file(task_id):
+    """Checks if the ID exists in the file."""
+    with open(tasks.TASK_METADATA_FILENAME, "r", encoding="utf-8") as f:
+        json_list = [json.loads(line) for line in f]
+    return task_id in map(lambda x: x["task_id"], json_list)
+
+
+def test_run_simulation_logging(mocker):
```
IvanPombo, line 20: Currently, we are trying to follow a more descriptive and focused pattern in the tests. 

For the descriptive part, the suggestion is to give a first name to the test that represents the function we are testing, then what we are inputing and then the expected output or setting you are using. 

For example, for this case it could be something like: `test_run_simulation__valid_task__write_in_metadata_file`. 

With this, your tests can be more focused on just the setting where the environment variable is not set. Then, you can create another one for the reverse. In this way, when a test fails, we will immediately know the case which is failing. 

To finish, don't be afraid to add a parameterization that tests more than one case (tests are the first QA barrier). For all of it, check the `mixins/test_filemanager.py`

IvanPombo (commented):
Thanks for the first round of tests Augusto. Leaving a bit of feedback to iterate and improve them.

inductiva/tests/tasks/test_run_task.py
AugustoPeres: We mock the `Task` to prevent this line of code from happening:

```
task = tasks.Task(task_id)
```

Which would make several call for the API. I believe that we would even have to validate an API which should not happen during this test.

inductiva/tests/tasks/test_run_task.py
AugustoPeres: Could you elaborate a bit on mocking the file?

inductiva/tests/tasks/test_run_task.py
sergiomsantos: yes, one test with mutiple parameters.

inductiva/tests/tasks/test_run_task.py
sergiomsantos: yea, just patch what is strictly required.
sometimes, rather than patching we can use "inversion of control". In this case, it means changing the `tasks.run_simulation` method to receive, as an argument, the `methods.invoke_async_api` method:

```python
def tasks.run_simulation(..., api_invoker=None):
    if api_invoker is None:
        api_invoker = methods.invoke_async_api
   ...
    task_id = api_invoker(...)
```

In the test, you'd then call the `tasks.run_simulation` passing it you api mock. No need to patch and nothing else can break due to patching.

sergiomsantos (commented):
Looking good!
After some discussion with Fábio, I think we're now in a position where we can have custom metadata fed into the logging system. Something along the lines:

```python
class Mysimulator():
    ...
    def run(self, ..., extra_metadata=None):
        super().run(..., extra_metadata)
```
with this we can then finalize this task.
Thanks

inductiva/tasks/run_simulation.py
AugustoPeres: I would prefer to leave this here because the behaviour can be deactivated. Therefore, it would be nice to have some type of logs of stating whether or not metadata is being saved

sergiomsantos (approved):

inductiva/tasks/run_simulation.py
```diff
@@ -27,16 +39,58 @@ def run_simulation(
...
+    if api_invoker is None:
+        api_invoker = methods.invoke_async_api
+
+    if not format_utils.getenv_bool("DISABLE_TASK_METADATA_LOGGING", False):
+        metadata = {
+            "storage_dir": str(storage_dir),
+            "api_method_name": api_method_name,
+            "machine_group_id": resource_pool_id,
+            **kwargs,
+        }
```
IvanPombo, line 51: Change the `storage_dir` to last, since it will be the least used.

```suggestion
        metadata = {
            "api_method_name": api_method_name,
            "machine_group_id": resource_pool_id,
            "storage_dir": str(storage_dir),
            **kwargs,
         }
```

PR #924, opened 2024-01-09
Add MPI cluster class

IvanPombo:
This PR adds the class to launch an MPI cluster as a child of the `BaseMachineGroup` class. The only change is the addition of the attribute `type` which differs in the backend between `mpi` and `standard` to launch the group.

Moreover, the logs are already pointing to an MPI cluster and with this reasoning I followed the decision of inheriting from the base class instead of the `MachineGroup` for now.


inductiva/resources/machine_cluster.py
```diff
@@ -0,0 +1,78 @@
...
+                 disk_size_gb: int = 70,
+                 register: bool = True) -> None:
+        """Create a MPICluster object.
+
+        Args:
+            machine_type: The type of GC machine to launch. Ex: "e2-standard-4".
+              Check https://cloud.google.com/compute/docs/machine-resource for
+              information about machine types.
+            num_machines: The number of virtual machines to launch.
+            disk_size_gb: The size of the disk in GB, recommended min. is 60 GB.
```
AugustoPeres, line 27: Missing documentation for the `register` argument.

inductiva/simulators/dualsphysics.py
```diff
@@ -17,21 +17,22 @@ def run(
...
         storage_dir: Optional[types.Path] = "",
     ) -> tasks.Task:
         """Executes a DualSPHysics simulation.
 
         Args:
             input_dir: Directory with simulation input files.
             sim_config_filename: Simulation config file.
-            machine_group: Machine group for simulation.
+            on: The computational resource to launch the simulation in. If None
+                the simulation is launched in a machine of the default pool.
```
AugustoPeres, line 29: Make `in` `on` the same either on the method argument or on the documentation

```suggestion
            on: The computational resource to launch the simulation on. If None
                the simulation is launched in a machine of the default pool.
```

AugustoPeres (approved):
LGTM:)

inductiva/resources/machine_cluster.py
IvanPombo: Will attempt to making it clear, but on the description of the method, since this is not supposed to be changed by the users. Eventually, this will need to be changed. But  that will be done on an extra step.

inductiva/resources/machine_cluster.py
```diff
@@ -0,0 +1,85 @@
...
+        can not be started in the cloud. This serves has an helper argument for
+        retrieving already registered machine groups that can be started, for
+        example, when retrieving with the `machines_groups.get` method.
+        Users should not set this argument.
+
+        Args:
+            machine_type: The type of GC machine to launch. Ex: "e2-standard-4".
+              Check https://cloud.google.com/compute/docs/machine-resource for
+              information about machine types.
+            num_machines: The number of virtual machines to launch.
```
sergiomsantos, line 34: I'd rather have `num_vcpus` instead of `num_machines`. This option might require other classes to have the same argument renamed.

inductiva/resources/machine_groups.py
```diff
@@ -127,8 +129,10 @@ def get():
     for mg in machine_groups:
         if mg["is_elastic"]:
             mg_class = resources.ElasticMachineGroup
-        else:
+        elif mg["type"] == "standard":
             mg_class = resources.MachineGroup
+        else:
+            mg_class = resources.MPICluster
```
sergiomsantos, line 135: this should not be the final else clause

inductiva/resources/machines_base.py
```diff
@@ -60,19 +73,20 @@ def _register_machine_group(self, **kwargs):
...
 
         machine_group = cls(
             machine_type=resp["machine_type"],
-            spot=bool(resp["spot"]),
             disk_size_gb=resp["disk_size_gb"],
             register=False,
         )
         machine_group._id = resp["id"]
         machine_group._name = resp["name"]
+        machine_group.spot = resp["spot"]
```
sergiomsantos, line 89: I think this attribute only makes sense to exist in a Derived class that actually supports spot.
spot is non-sensical in the context of the MPICluster and does not make sense to have this attr available there

inductiva/simulators/dualsphysics.py
```diff
@@ -2,36 +2,37 @@
...
 
-class DualSPHysics(Simulator):
+class DualSPHysics(simulators.Simulator):
     """Class to invoke a generic DualSPHysics simulation on the API."""
 
     def __init__(self):
         super().__init__()
         self.api_method_name = "sph.dualsphysics.run_simulation"
 
+    @simulators.simulator.mpi_disabled
```
sergiomsantos, line 15: Afeter some thought, I came to the conclusion that I don't like have a decorator marking something that is not available. It should be marking something that is available. Let's talk about this

sergiomsantos (commented):
somme comments; let's talk about them

inductiva/resources/machine_cluster.py
```diff
@@ -0,0 +1,86 @@
...
+            disk_size_gb: The size of the disk in GB, recommended min. is 60 GB.
+        """
+        super().__init__(machine_type=machine_type,
+                         disk_size_gb=disk_size_gb,
+                         register=register)
+        self.num_machines = num_machines
+        self.is_elastic = False
```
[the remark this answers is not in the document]
IvanPombo: The current reasoning follows from the backend schema, where the is_elastic is not set by default. Later we can change that. For now I am gonna hide it for now.

inductiva/resources/machine_cluster.py
IvanPombo: In this case it is the actual number of machines. Notice, that a single machine can have multiple `vcpus`. Moreover, currently, we cannot assemble as we wish a group of machines to achieve a certain `num_vcpus`.

inductiva/resources/machine_cluster.py
```diff
@@ -0,0 +1,86 @@
...
+    def from_api_response(cls, resp: dict):
+        machine_group = super().from_api_response(resp)
+        machine_group.num_machines = int(resp["num_vms"])
+        machine_group.register = False
+        return machine_group
+
+    def start(self):
+        """Start the MPI Cluster."""
+        return super().start(num_vms=self.num_machines,
+                             is_elastic=self.is_elastic,
```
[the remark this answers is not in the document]
IvanPombo: It breaks because it won't have all the necessary parameters in the backend. Let's make it hidden for now and later we change all of this from the ground up (`BaseMachineGroup` I am looking at you).

inductiva/resources/machines_base.py
```diff
@@ -9,12 +10,18 @@
 from inductiva.client.apis.tags import compute_api
 
 
+class ResourceType(enum.Enum):
+    """Enum to represent the type of machine to be launched."""
+
+    STANDARD = "standard"
+    MPI = "mpi"
+
+
```
sergiomsantos, line 19: Use a `StrEnum`; its members can be used as string wherever strings are needed:

```suggestion
class ResourceType(enum.StrEnum):
    """Enum to represent the type of machine to be launched."""

    STANDARD = "standard"
    MPI = "mpi"


```

inductiva/resources/machine_cluster.py
```diff
@@ -0,0 +1,89 @@
...
+              Check https://cloud.google.com/compute/docs/machine-resource for
+              information about machine types.
+            num_machines: The number of virtual machines to launch.
+            disk_size_gb: The size of the disk in GB, recommended min. is 60 GB.
+        """
+        super().__init__(machine_type=machine_type,
+                         disk_size_gb=disk_size_gb,
+                         register=register)
+        self.num_machines = num_machines
+        self.__type = machines_base.ResourceType.MPI.value
```
sergiomsantos, line 41:
```suggestion
        self.__type = machines_base.ResourceType.MPI
```

inductiva/resources/machine_groups.py
```diff
@@ -47,25 +49,31 @@ def _machine_group_list_to_str(machine_group_list) -> str:
...
+        if isinstance(machine_group, resources.ElasticMachineGroup):
             num_active_machines = machine_group.num_active_machines
             max_machines = machine_group.max_machines
             num_active_machines = f"{num_active_machines}/{max_machines}"
+            is_elastic = True
         else:
-            # TODO: retrieve the number of max. requested machines from the API
             num_active_machines = machine_group.num_machines
+            if isinstance(machine_group, resources.MPICluster):
+                resource_type = machines_base.ResourceType.MPI.value
```
sergiomsantos, line 63:
```suggestion
                resource_type = machines_base.ResourceType.MPI
```

inductiva/resources/machines.py
```diff
@@ -71,11 +83,12 @@ def estimate_cloud_cost(self):
...
 
         Returns:
-            The estimated cost per hour of the machine group in US
+            The estimated cost per hour of the machine group, in US
               dollars ($/h)."""
         #TODO: Contemplate disk size in the price.
-        estimated_cost = super()._get_estimated_cost() * self.num_machines
-        logging.info("Estimated cloud cost per hour for all machines : %s $/h",
+        estimated_cost = super()._get_estimated_cost(
+            self.spot) * self.num_machines
```
sergiomsantos, line 90: the breaking of lines make it difficult to follow.

```suggestion
       cost_per_machine = super()._get_estimated_cost(self.spot)
       estimated_cost =  cost_per_machine * self.num_machines
```

inductiva/simulators/simulator.py
```diff
@@ -6,12 +6,30 @@
 from inductiva.utils import files
 
 
+def mpi_enabled(cls):
+    """Class decorator that adds MPICluster to the supported resources.
+    """
+    supports = getattr(cls, "_supported_resources", None)
+    cls._supported_resources = supports | {resources.MPICluster}  # pylint: disable=protected-access
```
sergiomsantos, line 13: The default in getattr shouldn't be None, because None isn't a supported resource.
At most, it could be `type(None)`, but `NoneType` isn't a supported resource.

```suggestion
    supports = getattr(cls, "_supported_resources", set())
    cls._supported_resources = supports | {resources.MPICluster}  # pylint: disable=protected-access
```

inductiva/simulators/simulator.py
```diff
@@ -6,12 +6,30 @@
...
+        resources.MachineGroup, resources.ElasticMachineGroup
+    }
+
     def __init__(self):
         self.api_method_name = ""
 
+    @classmethod
+    def get_supported_resources(cls):
+        """Get the supported computational resources for this simulator."""
+        return tuple(cls._supported_resources.copy())
```
sergiomsantos, line 31: why return a tuple? We're already returning a copy of the set which is guaranteed to be unique.

Either
```suggestion
        return tuple(cls._supported_resources)
or
        return cls._supported_resources.copy()
```

inductiva/tests/resources/test_machines.py
```diff
@@ -0,0 +1,52 @@
...
+    Goal: Verify that the MPICluster is initializating and the registration is
+    processed, by mocking it instead of calling the API.
+    """
+    inductiva.api_key = "dummy"
+    cluster = inductiva.resources.MPICluster(machine_type="c2-standard-16",
+                                             num_machines=2)
+
+    # Check that the cluster has been initialized correctly
+    assert cluster.name == "name-resource"  # pylint: disable = protected-access
+    assert cluster.id == "id-resource"  # pylint: disable = protected-access
```
sergiomsantos, line 27: are these accesses protected? It doesn't look like they are

inductiva/tests/resources/test_machines.py
```diff
@@ -0,0 +1,52 @@
...
+@mock.patch.object(inductiva.resources.MPICluster,
+                   attribute="_register_machine_group",
+                   new=fake_register)
+def test_machines__mpicluster__register():
+    """Check the registering of a MPICluster.
+    
+    Goal: Verify that the MPICluster is initializating and the registration is
+    processed, by mocking it instead of calling the API.
+    """
+    inductiva.api_key = "dummy"
```
sergiomsantos, line 21: why not set this at the top of the the module?

```python
"""Unit tests for the Computational Resources classes."""
from unittest import mock
import inductiva

inductiva.api_key = "dummy"

...
```

BTW, why not have a key for these tests? It can be a secreted inject by github when running the tests.

inductiva/tests/simulators/test_simulator.py
```diff
@@ -1,11 +1,119 @@
...
+                                   register=False)
+
+    simulator = TesterSimulator()
+
+    with pytest.raises(ValueError) as excinfo:
+        simulator.validate_computational_resources(cluster)
+
+    expected_error = "The computational resource is invalid."
+
+    assert expected_error in str(excinfo.value)
```
sergiomsantos, line 41:
```suggestion
    assert "The computational resource is invalid" in str(excinfo.value)
```

inductiva/tests/simulators/test_simulator.py
```diff
@@ -1,11 +1,119 @@
...
+
+
+def test_simulator_run__mpi_enabled__with_mpi_cluster():
+    """Check mpi simulator validates MPICluster.
+
+    Goal: Verify that mpi_enabled simulators correctly validate the MPICluster
+    on a simulation.run call."""
+
+    inductiva.api_key = "dummy"
```
sergiomsantos, line 50: set as Github secret or remove to top of module.

inductiva/tests/simulators/test_simulator.py
```diff
@@ -1,11 +1,119 @@
...
+
+    with pytest.raises(ValueError) as excinfo:
+        simulator.validate_computational_resources(cluster)
+
+    expected_error = "The computational resource is invalid."
+
+    assert expected_error in str(excinfo.value)
+
+
+def test_simulator_run__mpi_enabled__with_mpi_cluster():
```
sergiomsantos, line 44: no call to `run` method as test name suggests.

inductiva/tests/simulators/test_simulator.py
```diff
@@ -1,11 +1,119 @@
...
+    MPIEnabledSim = simulators.simulator.mpi_enabled(TesterSimulator)
+    simulator = MPIEnabledSim()
+
+    try:
+        simulator.validate_computational_resources(cluster)
+    except ValueError:
+        assert False, "'validate_computational_resources' raised an exception."
+
+
+def test_simulator_run__non_mpi_enabled__with_machine_group():
```
sergiomsantos, line 64: same here; no call to `run` method as test name suggests.

inductiva/simulators/simulator.py
```diff
@@ -50,18 +68,37 @@ def run(
...
+            valid_resources: The valid computational resources for the simulator
+        """
+
+        supported_resources = self._supported_resources
+        if resource is not None and not isinstance(resource,
+                                                   tuple(supported_resources)):
+            raise ValueError(
+                "The computational resource is invalid. for "
+                "this simulator. Valid computational resources for this "
+                f"simulator are: {supported_resources}.")
```
sergiomsantos, line 104:
```suggestion
        if resource is not None \
            and not isinstance(resource, tuple(self._supported_resources)):
            raise ValueError(
                "The computational resource is invalid for this simulator. "
                f"Expected one of {self._supported_resources} but got {resource}.")
```

sergiomsantos (commented):
The logic in `validate_computational_resources` needs testing.

inductiva/resources/machines_base.py
IvanPombo: The reason I didn't follow this suggestion is because the StrEnum is only available from Python 3.11.

inductiva/tests/resources/test_machines.py
IvanPombo: This are the properties used to fetch the private attributes.

inductiva/tests/resources/test_machines.py
IvanPombo: This key isn't used for API calls so no need to have it being a secret.

sergiomsantos (approved):

PR #928, opened 2024-01-10
Warns the user when there are active machine groups but the tasks are being submitted to the default

AugustoPeres:
inductiva/tasks#22

### DIFF
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -7,7 +7,6 @@ import os
 import pathlib
 import signal
 import time
-import warnings
 from urllib3.exceptions import MaxRetryError, NewConnectionError
 from contextlib import contextmanager
 from typing import Any, Dict, List, Optional, Tuple, Type
@@ -332,10 +331,6 @@ def invoke_async_api(method_name: str,
 
         if resource_pool_id is None:
             logging.info("Task submitted to the default resource pool.")
-            if inductiva.resources.machine_groups.get():
-                warnings.warn(
-                    "Submiting task to default resource pool with other "
-                    "active machine groups.")
         else:
             logging.info("Task submitted to machine group %s.",
                          "api-" + resource_pool_id)
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 import signal
 import time
+import warnings
 from urllib3.exceptions import MaxRetryError, NewConnectionError
 from contextlib import contextmanager
 from typing import Any, Dict, List, Optional, Tuple, Type
@@ -331,6 +332,10 @@ def invoke_async_api(method_name: str,
 
         if resource_pool_id is None:
             logging.info("Task submitted to the default resource pool.")
+            if inductiva.resources.machine_groups.get():
+                warnings.warn(
+                    "Submiting task to default resource pool with other "
+                    "active machine groups.")
         else:
             logging.info("Task submitted to machine group %s.",
                          "api-" + resource_pool_id)
--- a/inductiva/simulators/simulator.py
+++ b/inductiva/simulators/simulator.py
@@ -10,7 +10,7 @@ from inductiva.utils import files
 
 class Simulator(ABC):
     """Base simulator class."""
-    __check_active_resources = True
+    _check_active_resources = True
 
     def __init__(self):
         self.api_method_name = ""
@@ -40,8 +40,8 @@ class Simulator(ABC):
         return input_dir
 
     def _check_resource_pool(self, machine_group):
-        if self.__check_active_resources:
-            self.__check_active_resources = False
+        if self._check_active_resources:
+            self._check_active_resources = False
             if machine_group is None:
                 if inductiva.resources.machine_groups.get():
                     warnings.warn(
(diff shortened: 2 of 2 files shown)


inductiva/api/methods.py
```diff
@@ -331,6 +332,10 @@ def invoke_async_api(method_name: str,
 
         if resource_pool_id is None:
             logging.info("Task submitted to the default resource pool.")
+            if inductiva.resources.machine_groups.get():
```
IvanPombo, line 335: I am not sure if this API call should be in the middle of a simulation workflow. 
This seems like babysitting the users. I think we should provide the tool more clearly, for example, in the CLI so that they are aware of their active machine groups.

This seems to be warning for the mistake of forgetting to add the `machine_group/on` parameter in the simulator. Then, we probably should just make the message above more "visible", so that they are reminded of that in their workflow.

AugustoPeres: So you are suggesting, for example:

```python
    def run(
        self,
        input_dir: types.Path,
        *_args,
        extra_metadata: Optional[dict] = None,
        **kwargs,
    ) -> tasks.Task:

        input_dir = self._setup_input_dir(input_dir)

        # Warn user here

        return tasks.run_simulation(
            self.api_method_name,
            input_dir,
            machine_group=machine_group,
            storage_dir=storage_dir,
            extra_metadata=extra_metadata,
            **kwargs,
        )

```

IvanPombo: What I was trying to say is that we shouldn't make calls to check if there are machines active in the middle of the simulation workflow. Think like this, should we also warn the user if he is launching a new machine group, when he has others active?

I think the user is responsible for this, otherwise we would be responsible for warning of various other stuff.

sergiomsantos: Some interesting points here. It is useful to inform the user that he has allocated resources that he is not using, but I understand that it might become to noisy to inform him about that every time.

I suggest extracting this logic to an external function that gets called everytime, but the call to `inductiva.resources.machine_groups.get()` only gets made once and the warning is only emitted the first time the user runs a task on the default pool even though he has resources allocated. The message should be clear about what is being checked and that the warning will be silenced from that point onwards

sergiomsantos (changes requested):
Lets warn the user only once.

inductiva/api/methods.py
AugustoPeres: So we would like something like this:

```python
simulator = SomeSimulator(...)
simulator.simulate(...)  # -> This warns the user
simulator.simulate(...). # -> This no longer warns the user

simulator = AnotherSimulator(...)
simulator.simulate(...)  # -> This warns the user AGAIN
simulator.simulate(...). # -> This no longer warns the user
```

Or would we like the user to be warned only once even though a new simulator was just created. 

```python
simulator = SomeSimulator(...)
simulator.simulate(...)  # -> This warns the user ONLY here
simulator.simulate(...)

simulator = AnotherSimulator(...)
simulator.simulate(...)
simulator.simulate(...)
```

In the first case we could simply go about setting an internal variable on the simulator. Something like: `self.checked_resource_pool`. In the second case we could make use of environment variables (which would also be exposed to the user).

inductiva/simulators/simulator.py
```diff
@@ -36,6 +39,15 @@ def _setup_input_dir(self, input_dir: types.Path):
                 f"The provided path (\"{input_dir}\") is not a directory.")
         return input_dir
 
+    def _check_resource_pool(self, machine_group):
```
IvanPombo, line 42: Still not sure if I like this, however I will leave here my comments either way.

inductiva/simulators/simulator.py
```diff
@@ -36,6 +39,15 @@ def _setup_input_dir(self, input_dir: types.Path):
...
+        if not self.checked_resource_pool:
+            self.checked_resource_pool = True
+            if inductiva.resources.machine_groups.get(
```
IvanPombo, line 45: This get should be the last thing you do, since otherwise you will call the API unnecessarily, for e.g. when machine_group is None, which is exactly when you want to check it.

inductiva/simulators/simulator.py
```diff
@@ -36,6 +39,15 @@ def _setup_input_dir(self, input_dir: types.Path):
                 f"The provided path (\"{input_dir}\") is not a directory.")
         return input_dir
 
+    def _check_resource_pool(self, machine_group):
+        if not self.checked_resource_pool:
```
IvanPombo, line 43: I suggest the reverse, name it `__check_active_resources` and set it to True first. Moreover, this argument needs to be hidden so users don't mess with it.

inductiva/simulators/simulator.py
```diff
@@ -11,6 +13,7 @@ class Simulator(ABC):
 
     def __init__(self):
         self.api_method_name = ""
+        self.checked_resource_pool = False
```
IvanPombo, line 16: In principle, this can be a class attribute, hence when its value changes when you do `sim = Simulator()` it will be changed for further initializations in the "session".

IvanPombo (commented):
Left a few comments, still not sure about it.

PR #937, opened 2024-01-12
Add new timestamps to tasks and respective list methods

IvanPombo:
This PR adds the correct computation timestamp to the tasks and the respective list methods.

In this way, we and the users can verify what is the correct time that the computation took to run, which doesn't account for the input, queue, and upload to buckets time.

Moreover, we have changed slightly the `seconds_formatter` method and added in the required places where before we had `3.34mins` or something.

Notice that, the Client was updated due to changes on the backend and this is the nature of the large changes.


inductiva/tasks/task.py
```diff
@@ -274,9 +274,34 @@ def _get_path_params(self) -> _PathParams:
...
+        if end_time is None:
+            end_time = datetime.datetime.now(datetime.timezone.utc)
+        else:
+            end_time = datetime.datetime.fromisoformat(
+                info["computation_end_time"])
+
+        return (end_time - start_time).total_seconds()
+
+    def get_total_time(self, fail_if_running: bool = True) -> Optional[float]:
+        """Get the total time the task workflow took to complete.
```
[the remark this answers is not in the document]
IvanPombo: That is what I actually want, given that we already have the computation time. Gonna change the code to address that.

inductiva/tasks/methods.py
```diff
@@ -51,43 +51,59 @@ def _fetch_tasks_from_api(
 
 def _list_of_tasks_to_str(tasks: Sequence["inductiva.tasks.Task"]) -> str:
     columns = [
-        "ID", "Simulator", "Status", "Submitted", "Started", "Duration",
-        "VM Type"
+        "ID", "Simulator", "Status", "Submitted", "Picked-Up",
```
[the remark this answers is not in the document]
IvanPombo: Gonna revert this back.

sergiomsantos (changes requested):

inductiva/tasks/task.py
```diff
@@ -274,29 +272,55 @@ def _get_path_params(self) -> _PathParams:
         """Get dictionary with the URL path parameters for API calls."""
         return {"task_id": self.id}
 
-    def get_execution_time(self,
-                           fail_if_running: bool = True) -> Optional[float]:
-        """Get the time the task took to complete.
+    def get_computation_time(self,
```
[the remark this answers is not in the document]
IvanPombo: All of these times are obtained with an API call in the function `self.get_info()`. Here, we just process the data arriving from there in terms of status and the existence or not of the times.

In the `self.get_info()` method we only call the API again if we haven't retrieved the data yet or the status is not terminal. However, containing this in a `TaskInfo` object is a good approach, but out of the scope of this PR.

inductiva/tasks/task.py
```diff
@@ -274,29 +272,55 @@ def _get_path_params(self) -> _PathParams:
...
+        if info["computation_start_time"] is None:
             return None
 
         # Format the time to datetime type
-        start_time = datetime.datetime.fromisoformat(info["start_time"])
+        start_time = datetime.datetime.fromisoformat(
+            info["computation_start_time"])
+        end_time = info.get("computation_end_time")
+        if end_time is None:
+            end_time = datetime.datetime.now(datetime.timezone.utc)
```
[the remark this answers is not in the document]
IvanPombo: The idea is to provide how long the simulation has been running without much trouble.

inductiva/tasks/README.md
```diff
@@ -132,10 +132,10 @@ import inductiva
...
-#  1695306097851999678      protein solvation         gromacs              success     21 Sep, 15:21:38     21 Sep, 15:21:40      0h 15m 43s      c2-standard-4
-#  1695294928922012701            wind tunnel        openfoam              success     20 Sep, 12:15:31     20 Sep, 12:15:33       0h 0m 50s      c2-standard-4
-#  1695294363618736570            wind tunnel        openfoam              success     20 Sep, 12:06:06     20 Sep, 12:06:06       0h 0m 50s      c2-standard-4
+#                       ID           Simulator     Status          Submitted          Picked-Up Computation Time  Total Duration            VM Type
+#on61z6u9c17uatqnbios7zjo3 openfoam_foundation    success   12 Jan, 10:27:19   12 Jan, 10:27:42        0h 2m 20s       0h 2m 50s   c2d-standard-112
+#u43sobjkk906566xzmm1x8x8t openfoam_foundation     killed   12 Jan, 10:21:51   12 Jan, 10:22:39        0h 1m 42s        0h 2m 8s   c2d-standard-112
+#k56bidnib55bkiv5jpmo0cp3z openfoam_foundation    started   12 Jan, 10:20:14   12 Jan, 10:22:39       *0h 7m 58s      *0h 7m 58s   c2d-standard-112
```
sergiomsantos, line 138: The asterisk needs a legend so people know those timing refer to ongoing tasks.
Also, the prompt `inductiva.tasks.list(5, status="success")` states that we're listing the top 5 successful runs, and that it not what is shown.
We need to be coherent.

inductiva/tasks/methods.py
```diff
@@ -49,55 +51,49 @@ def _fetch_tasks_from_api(
...
 
         row = [
             task.id,
             simulator,
-            task.get_status(),
-            info.get("input_submit_time", None),
-            info.get("start_time", None),
+            status,
+            format_utils.datetime_formatter(info.get("input_submit_time",
+                                                     None)),
```
sergiomsantos, line 89: unnecessary line break. It makes it difficult to read.

inductiva/utils/format_utils.py
```diff
@@ -28,14 +29,15 @@ def bytes_formatter(n_bytes: int) -> str:
...
     local_dt = datetime.datetime.fromisoformat(dt).astimezone()
     return local_dt.strftime("%d %b, %H:%M:%S")
 
 
 def seconds_formatter(secs: float) -> str:
-    return (f"{int(secs // 3600)}h "
-            f"{int((secs % 3600) // 60)}m "
-            f"{int(secs % 60)}s")
+    """Convert seconds to time human readable string."""
+    return str(datetime.timedelta(seconds=round(secs)))
```
sergiomsantos, line 40: why round?

inductiva/_cli/tasks_parsers.py
sergiomsantos, line 1: let's rename this file to `tasks.py`. the `_parsers` suffix doesn't make sense because the module contains non parser related stuff.

inductiva/tasks/methods.py
```diff
@@ -106,8 +102,8 @@ def list(last_n: int = 5,
     # pylint: disable=line-too-long
     """List the last N tasks of a user.
 
-    This function lists info about the last N tasks (with respect to submission
-    time) of a user to stdout, sorted by submission time with the
+    This function returns a dataframe with info about the last N tasks (with
```
sergiomsantos, line 105:
```suggestion
    This function returns a dataframe with information about the last N tasks (with
```

sergiomsantos (commented):
some minor changes before approval

inductiva/tasks/methods.py
IvanPombo: This yapf hitting the boundary 🗡️

inductiva/utils/format_utils.py
IvanPombo: To avoid having `1:30:24.58457`.

sergiomsantos (approved):

IvanPombo:
Merging this, since most of the changes are related with text! Any typos found along the way will be welcome to be fixed!

PR #943, opened 2024-01-16
Quick fix resolve path on mixins

IvanPombo:
This fixes a bug that wasn't thought of which occurs when setting `inductiva.working_dir` to be a path outside the current working directory. For example,
`inductiva.working_dir = "../some_path"`. 

With this setting, the if to check if the directory name already existed was not working inside this new working directory, and was checking inside the local directory where python was being run. 

By setting before the `files.resolve_path(root_directory)` before that if this bug is fixed.

### DIFF
--- a/inductiva/mixins/file_manager.py
+++ b/inductiva/mixins/file_manager.py
@@ -46,7 +46,6 @@ class FileManager:
                 If None, an error is raised.
         """
 
-
         if root_dir is None:
             raise ValueError("Given root directory cannot be None")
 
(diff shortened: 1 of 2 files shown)


AugustoPeres (approved):
LGTM. Thanks Ivan,
Maybe leave a more detailed description of the bug. Also helpful for QA later on

IvanPombo:
> LGTM. Thanks Ivan,
> 
> Maybe leave a more detailed description of the bug. Also helpful for QA later on

Done! Gonna merge anyway since we will do a large QA session later this week.

PR #947, opened 2024-01-17
MachineType to enum

AugustoPeres:
Should handle inductiva/tasks#7.

### DIFF
--- a/inductiva/resources/machine_types.py
+++ b/inductiva/resources/machine_types.py
@@ -1,14 +1,11 @@
-import enum
-
-AVAILABLE_MACHINES = [
-    ("c2-standard-", [4, 8, 16, 30, 60]),
-    ("c3-standard-", [4, 8, 22, 44, 88, 176]),
-    ("c2d-standard-", [2, 4, 8, 16, 32, 56, 112]),
-    ("c2d-highcpu-", [2, 4, 8, 16, 32, 56, 112]),
-    ("e2-standard-", [2, 4, 8, 16, 32]),
-    ("n2-standard-", [2, 4, 8, 16, 32, 48, 64, 80, 96, 128]),
-    ("n2d-standard-", [2, 4, 8, 16, 32, 48, 64, 80, 96, 128, 224]),
-    ("n1-standard-", [1, 2, 4, 8, 16, 32, 64, 96]),
-]
-
-MachineType = enum.Enum("MachineType", AVAILABLE_MACHINES)
+"""Available machine types and their number of cores."""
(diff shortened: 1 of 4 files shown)


inductiva/resources/machine_types.py
```diff
@@ -0,0 +1,11 @@
...
+AVAILABLE_MACHINES = {
+    "c2-standard-": [4, 8, 16, 30, 60],
+    "c3-standard-": [4, 8, 22, 44, 88, 176],
+    "c2d-standard-": [2, 4, 8, 16, 32, 56, 112],
+    "c2d-highcpu-": [2, 4, 8, 16, 32, 56, 112],
+    "e2-standard-": [2, 4, 8, 16, 32],
+    "n2-standard-": [2, 4, 8, 16, 32, 48, 64, 80, 96, 128],
+    "n2d-standard-": [2, 4, 8, 16, 32, 48, 64, 80, 96, 128, 224],
+    "n1-standard-": [1, 2, 4, 8, 16, 32, 64, 96],
+}
```
sergiomsantos, line 11: add an utility method here that returns a generator of all available machines:

```python
def list_available_machines():
    return (machine_type+str(vcpu)
            for machine_type, vcpus in AVAILABLE_MACHINES.items()
            for vcpu in vcpus)
```

 It can then be used elsewhere. Ex:

```python
if 'xpto-standard-3' not in list_available_machines():
    raise Exception()
```

The gist is that we can then replace the body of the `list_available_machines` with an API call whenever the backend implements that endpoint , with minor changes to the overall code.

inductiva/resources/machines_base.py
```diff
@@ -40,6 +40,12 @@ def __init__(self,
                 example, when retrieving with the `machines_groups.get` method.
                 Users should not set this argument in anyway.
         """
+        allowed_machine_types = [
+            machine_type + str(core) for machine_type, cores in
+            inductiva.resources.AVAILABLE_MACHINES.items() for core in cores
+        ]
+        assert machine_type in allowed_machine_types,\
+            "Machine type not supported."
```
sergiomsantos, line 48: no assertions - use exceptions

sergiomsantos, line 48:
```suggestion
        if machine_type no in machine_types.list_available_machines():
            raise ValueError("Machine type not supported.")
```

sergiomsantos (commented):
this task will likely change when we have a backend endpoint to get this information.
Before that happens, we can already start preparing the code to use that endpoint.

PR #951, opened 2024-01-18
Add a websocket consumer for the logs

IvanPombo:
**TLDR** This PR adds a baby websocket consumer to the Client for the purpose of streaming the task logs from the backend endpoint. 

## Implementation
As a first step, we add a new module named `logs` to the Client, with the sole method to connect to the websocket and read from it whatever is being stream.

For the purpose of demonstration, we have created a Flask server that streams some random logs based on the provided `TaskId`. For dummy purposes, the server is as follows:

```python
from flask import Flask
from flask_sock import Sock
import time
import random
import string

app = Flask(__name__)
sock = Sock(app)

@sock.route("/logs/<task_id>")
def logs(ws, task_id):
    timer = 0
    message = f"Logs of Task_ID: {task_id}."
    ws.send(message)
    while True:
        random_string = "".join(random.choices(string.ascii_lowercase, k=10))
        time.sleep(1)
        timer += 1
        message = f"Simulation Time: {timer}s: {random_string}"
        ws.send(message)

@sock.route("/close")
def close(ws):
    ws.close()
```

To run this server install `pip install flask-sock` and it can be launched within the directory it lives with `flask run`. This provides an IP which we are connecting to, in the dummy implementation in the Client. 

On the client, we instantiate a websocket that connects to this IP and that allows users to start streaming the provided logs from the Flask server.

## Running the Logs

The first implementation to read the logs is from the CLI, and users can fetch the logs of a certain task via `inductiva logs {task_id}`.

For example, `inductiva logs 1234`, provides the following output until it's killed with `ctrl+c`:

Screenshot 2024-01-18 at 13 18 11

This PR proposes the MVP presented in inductiva/tasks#29.


inductiva/logs/task_logs.py
sergiomsantos, line 1: The logic in this module can be encapsulated inside a class.

inductiva/logs/__init__.py
```diff
@@ -0,0 +1,2 @@
+#pylint: disable=missing-module-docstring
+from . import task_logs
```
sergiomsantos, line 2: no need to export

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,32 @@
+import websocket
+
+
+def stream_task_logs(task_id):
+    """Stream the logs of a task through a websocket.
+    
+    Args:
+        task_id (int): ID of the task to print the logs of."""
+
+    logs_url = "ws://127.0.0.1:5000/logs"
```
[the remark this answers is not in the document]
IvanPombo: Already changed this to an environment variable, since we may need to change this to different urls depending on the stream from production and development.

With that in mind, should all these environment variables move there?

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,32 @@
+import websocket
+
+
+def stream_task_logs(task_id):
```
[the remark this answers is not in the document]
IvanPombo: I think this will be managed in the `on_message`.

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def setup_websocket(self):
+        """Initialize the websocket app, with the callbacks within the class."""
+        return websocket.WebSocketApp(self.websocket_url,
+                                      on_open=self.on_open,
+                                      on_message=self.on_message,
+                                      on_error=self.on_error,
+                                      on_close=self.on_close)
+
+    def on_message(self, ws, message):  # pylint: disable=unused-argument
```
AugustoPeres, line 32: Why the extra unused arguments? I would expect this if `TaskLogsStream` was inheriting from another class. But that does not seem to be the case

IvanPombo: The `ws` is the websocket connection, which a requirement for the respective attribute method in the `WebSocketApp`.

inductiva/constants.py
```diff
@@ -0,0 +1,5 @@
+"""Constants that can be set for the Inductiva client."""
+import os
+
+LOGS_WEBSOCKET = os.environ.get("INDUCTIVA_TASK_LOGS_URL",
```
sergiomsantos, line 4:
```suggestion
LOGS_WEBSOCKET_URL = os.environ.get("INDUCTIVA_TASK_LOGS_URL",
```

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,50 @@
...
+import websocket
+
+from inductiva import constants
+
+
+class TaskLogsStream:
+    """Logs the streams of a running task through a websocket."""
+
+    def __init__(self, task_id, fout=sys.stdout, ferr=sys.stderr):
+        """Initialize the websocket connection to the task's logs.
```
sergiomsantos, line 13:
```suggestion
        """Initialize a websocket connection to the given task's STDOUT and STDERR streams.
```

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,50 @@
...
+
+        Args:
+            task_id (int): ID of the task to print the logs of."""
```
sergiomsantos, line 16:
```suggestion
            task_id (int): ID of the task for which to get the STDOUT & STDERR streams."""
```

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,50 @@
...
+        self.ferr = ferr
+        self.ws = self.setup_websocket()
+
+    def setup_websocket(self):
+        """Initialize the websocket app, with the callbacks within the class."""
+        return websocket.WebSocketApp(self.websocket_url,
+                                      on_open=self.on_open,
+                                      on_message=self.on_message,
+                                      on_error=self.on_error,
+                                      on_close=self.on_close)
```
sergiomsantos, line 30:
```suggestion
        return websocket.WebSocketApp(self.websocket_url,
                                      on_open=self.on_open,
                                      on_error=self.on_error,
                                      on_close=self.on_close,
                                      on_message=self.on_message)
```

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def on_error(self, ws, error):  # pylint: disable=unused-argument
+        if not isinstance(error, KeyboardInterrupt):
+            print(error, file=self.ferr)
+
+    def on_close(self, ws, status_code, message):  # pylint: disable=unused-argument
+        print(f"Closed stream with {status_code}: {message}.", file=self.fout)
+
+    def on_open(self, ws):  # pylint: disable=unused-argument
+        print("Opening stream of the task logs...", file=self.fout)
```
sergiomsantos, line 43:
```suggestion
        print(f"Opening socket connection to logs of task {self.task_id} ...", file=self.fout)
```

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def stream_task_logs(self):
+        """Stream the logs of a task through a websocket.
+        
+        Args:
+            task_id (int): ID of the task to print the logs of."""
+        self.ws.run_forever(reconnect=5, ping_interval=10, ping_timeout=5)
```
sergiomsantos, line 50: define these constants as properties of the class

sergiomsantos: silence ping/pong messages

inductiva/logs/task_logs.py
hpenedones: can you explain better? I have the same doubt as @AugustoPeres ...
I think pylint was alerting to the fact that you are not using the argument "ws" anywhere in the code -- and ideally we would not disable that warning because it's there for a good reason... Like Augusto said, it seems like if this was derived class from a base class where those methods were declared (and it had those arguments) then ok, but your class does not seem to be inheriting from any base class... So can't you just get rid of it?
If not (because it's indeed implementing a method from a base class) and for some reason you don't use an argument, there are typically a few ways to avoid a linting warning in python (without disabling pylint):
1) delete the variable 
`del myvar`
2) prepend the name with a standard prefix that pylint understand, typically is `unused_` so it would be `unused_myvar` or sometimes is underscore `_myvar`...
See a long thread here:
https://stackoverflow.com/questions/10025680/how-can-i-denote-unused-function-arguments

sergiomsantos: This method is a callback. It needs to comply with the contract expected by the WebSocketApp. It is not used but needs to be there. This being said, I agree with the renaming of the unused variable

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,68 @@
...
+        self.websocket_url = f"{constants.LOGS_WEBSOCKET_URL}/{task_id}"
+        self.task_id = task_id
+        self.reconnect = reconnect
+        self.ping_interval = ping_interval
+        self.ping_timeout = ping_timeout
+        self.fout = fout
+        self.ferr = ferr
+        self.ws = self.setup_websocket()
+
+    def setup_websocket(self):
```
sergiomsantos, line 39: this method can also be made private.
By making these methods private, we just need to document them from the point of view of the developer and not the user.

inductiva/logs/task_logs.py
```diff
@@ -0,0 +1,68 @@
...
+              file=self.fout)
+
+    def stream_task_logs(self):
+        """Stream the logs of a task through a websocket.
+        
+        Args:
+            task_id (int): ID of the task to print the logs of."""
+        self.ws.run_forever(reconnect=self.reconnect,
+                            ping_interval=self.ping_interval,
+                            ping_timeout=self.ping_timeout)
```
sergiomsantos, line 68: When a method is basically a wrapper around a call to another method, I like to name it similarly to the name of the wrapped method. In this case, something simple like `run`.
In this context, and taking onto account what we're doing, I think the name of the class needs to change.
Something like `TaskStreamConsumer`; it basically consumes the streams produced by a task, whatever they are. In that case, renaming the method to `run` would make even more sense:

```python
consumer = TaskStreamConsumer('1234')
consumer.run()
```
 What do you think?

Also, the docstring is not compatible with the signature of the method.

sergiomsantos (commented):
Almost there.
Some comments and minor updates.
Before requesting re-review, make sure all docstrings, names, etc, are coherent.

IvanPombo:
@sergiomsantos Addressing the general picture, and given that this will be always specific for a task, I moved the file to the `tasks` folder under the name of `streams.py`. Hence, we remove another unneeded init file, and the extra name of `task` in the previous naming.

sergiomsantos (approved):

PR #953, opened 2024-01-18
Add sub-dir check

AugustoPeres:
inductiva/tasks#24

This pull requests prevents the user from using `simulator.run(input_dir=some_dir)`when `some_dir` is a parent of the current working directory.

### DIFF
--- a/inductiva/simulators/simulator.py
+++ b/inductiva/simulators/simulator.py
@@ -2,6 +2,8 @@
 from typing import Optional
 from abc import ABC
 
+import pathlib
+
 from inductiva import types, tasks, resources
 from inductiva.utils import files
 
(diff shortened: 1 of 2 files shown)


AugustoPeres:
Just noticed that `is_relative_to` only works for python versions between 3.9 and 3.14. Will change this

inductiva/simulators/simulator.py
```diff
@@ -76,6 +85,9 @@ def run(
             **kwargs: Additional keyword arguments to be passed to the
                 simulation API method.
         """
+        assert self._check_if_input_dir_is_not_parent(
+            input_dir), "Input dir cannot be parent of current_dir"
```
sergiomsantos, line 89: don't use assertions, because these can be skipped. Simply raise an exception.

inductiva/tests/simulators/test_simulator.py
```diff
@@ -105,3 +107,24 @@ def test_valid_resources__mpi_simulators(simulator):
     correctly."""
 
     assert resources.MPICluster in simulator.get_supported_resources()
+
+
+@mark.parametrize("input_dir,expects_assertion_error",
+                  [(pathlib.Path().cwd(), True),
+                   (pathlib.Path().cwd() / "..", True),
+                   (pathlib.Path().cwd() / ".." / "other_dir", False)])
```
sergiomsantos, line 115: what if `input_dir` is a grand+ parent of `cwd`. It should fail in that case as well.

PR #955, opened 2024-01-18
Adds command abstraction

AugustoPeres:
Addresses inductiva/tasks#16.

Not yet updated the READMEs as I want feed back before that.
Example usage:

```python
commands = [
    inductiva.commands.Command("gencase config flow_cylinder -save:all"),
    inductiva.commands.Command(
        "dualsphysics flow_cylinder flow_cylinder -dirdataout data -svres"),
    inductiva.commands.Command(
        "partvtk -dirin flow_cylinder/data -savevtk flow_cylinder/PartFluid -onlytype:-all,+fluid"
    )
]

# Initialize the Simulator
dualsphysics = inductiva.simulators.DualSPHysics()

task = dualsphysics.run(input_dir=input_dir, commands=commands)

```

### DIFF
--- a/inductiva/commands/__init__.py
+++ b/inductiva/commands/__init__.py
@@ -1,0 +1,2 @@
+# pylint: disable=missing-module-docstring
+from .commands import Command
--- a/inductiva/commands/commands.py
+++ b/inductiva/commands/commands.py
@@ -1,0 +1,10 @@
+"""Wrapper class for simulator commands."""
+
+
+class Command(dict):
+    """Abstraction class for commands."""
+
+    def __init__(self, cmd, *prompts):
+        super(Command, self).__init__(cmd=cmd)
+        if prompts:
+            self['prompts'] = list(prompts)
--- a/inductiva/commands/commands.py
+++ b/inductiva/commands/commands.py
@@ -5,6 +5,4 @@ class Command(dict):
     """Abstraction class for commands."""
 
     def __init__(self, cmd, *prompts):
-        super(Command, self).__init__(cmd=cmd)
-        if prompts:
-            self['prompts'] = list(prompts)
+        super(Command, self).__init__(cmd=cmd, prompts=prompts)
--- a/inductiva/__init__.py
+++ b/inductiva/__init__.py
@@ -10,6 +10,7 @@ from . import resources
 from . import storage
 from . import utils
 from . import tasks
+from . import commands
 
 api_url = os.environ.get("INDUCTIVA_API_URL", "https://api.inductiva.ai")
 output_dir = os.environ.get("INDUCTIVA_OUTPUT_DIR", "inductiva_output")


inductiva/commands/commands.py
```diff
@@ -0,0 +1,8 @@
+"""Wrapper class for simulator commands."""
+
+
+class Command(dict):
+    """Abstraction class for commands."""
+
+    def __init__(self, cmd, *prompts):
+        super().__init__(cmd=cmd, prompts=list(prompts))
```
sergiomsantos, line 8: no need to inherit from dict because the class then inherits a lot of methods that do not make sense to exist in this class.

Make it a simple class:


```suggestion
class Command(dict):
    """Abstraction class for commands."""

    def __init__(self, cmd, *prompts):
        self.cmd=cmd
        self.prompts=prompts # <-- keep it a tuple
    
    def to_json(self):
        return self.__dict__ # or just return a dic with the required keys
```


and to make it json serializable, use something like:

```python
>>> c = Command('command', 'p1', 'p2')
>>> json.dumps(c, default=lambda x: x.to_json())
{'cmd': 'command', 'prompts': ('p1', 'p2')}
```

inductiva/commands/__init__.py
```diff
@@ -0,0 +1,2 @@
+# pylint: disable=missing-module-docstring
+from .commands import Command
```
sergiomsantos, line 2: I don't like having `__init__` exporting everything.
Just leave it empty for now.

sergiomsantos (changes requested):
The `commands` argument in the `simulator.run` method should accept 2 things:

1. a list of `str`ings;
2. a list of `str`ings and `Command`s (intermingled)
3. a list of `Command`s

The lower level call to the API should then convert the 1st and 2nd cases into the third. The latter then gets properly serialized to be sent to the backend, respecting the current contract.

inductiva/simulators/simulator.py
```diff
@@ -80,6 +94,8 @@ def run(
 
         self.validate_computational_resources(on)
 
+        kwargs = self._commands_to_json(kwargs)
```
IvanPombo, line 97: In terms of structure, I think we should add a function named `_process_kwargs` where the `_commands_to_json` is used. Otherwise, this seems weird. In case of need, that function can be used for other kwargs in the future.

inductiva/simulators/simulator.py
```diff
@@ -54,6 +55,19 @@ def _setup_input_dir(self, input_dir: types.Path):
...
         return input_dir
 
+    def _commands_to_json(self, kwargs):
+        if not "commands" in kwargs:
+            return kwargs
+
+        def to_cmd(x):
+            if isinstance(x, inductiva.commands.Command):
+                return x
+            return inductiva.commands.Command(x)
```
IvanPombo, line 65: Here, we need an `else` to capture the case when `x` is not a just a str. Moreover, I think we can serialize to json immediately and update the namings as follows:

```suggestion
        def to_cmd_json(cmd):
            if isinstance(cmd, inductiva.commands.Command):
                return cmd.to_json()
            elif not isinstance(cmd, str):
                raise ValueError("Command structure not valid. It should be of type: 'str' or 'inductiva.commands.Command'.")
            
            return inductiva.commands.Command(cmd).to_json()
```

inductiva/simulators/simulator.py
```diff
@@ -54,6 +55,19 @@ def _setup_input_dir(self, input_dir: types.Path):
...
+
+        commands = [to_cmd(cmd).to_json() for cmd in kwargs["commands"]]
```
IvanPombo, line 67: Then, here would just be:

```suggestion
        commands = [to_cmd_json(cmd) for cmd in kwargs["commands"]]
```

inductiva/simulators/simulator.py
```diff
@@ -2,6 +2,7 @@
 from typing import Optional
 from abc import ABC
 
+import inductiva
```
IvanPombo, line 5: Consider the imports mentioned above.

IvanPombo (commented):
First round of comments for clarity purposes only!

inductiva/simulators/simulator.py
AugustoPeres: > Here, we need an `else` to capture the case when `x` is not a just a str. Moreover, I think we can serialize to json immediately and update the namings as follows:

But it should be either a `string` or a `Command` object. As long as that is clear to the user it should be fine, no?

AugustoPeres: I will raise the error regardless

inductiva/commands/__init__.py
IvanPombo: Okay, it is the reverse. We should expose it here, and not on the main `__init__.py`.

So that the above code works.

inductiva/simulators/simulator.py
IvanPombo: Maybe, but it's tickling me that we are passing all kwargs through a `_commands_to_json` method.

inductiva/simulators/simulator.py
IvanPombo: You never know. They may pass a tuple to check if it is possible! And with an explicit error, they will know how to solve it without going to docs.

inductiva/simulators/simulator.py
IvanPombo: I like that!

inductiva/simulators/dualsphysics.py
```diff
@@ -1,8 +1,9 @@
 """DualSPHysics simulator module of the API."""
 
-from typing import Optional, List
+from typing import Optional, List, Union
 
 from inductiva import types, tasks, simulators
+from inductiva import commands as cmds
```
sergiomsantos, line 6: let's use our own API without renaming the namespace

sergiomsantos (changes requested):
We need validation on the Command ctor and tests.

inductiva/commands/commands.py
```diff
@@ -0,0 +1,16 @@
+"""Wrapper class for simulator commands."""
+
+
+class Command:
+    """Abstraction class for commands."""
+
+    def __init__(self, cmd: str, *prompts: str):
+        assert isinstance(cmd, str), "cmd argument must be a string."
```
sergiomsantos, line 8: don't use assertions. They can be turned off

inductiva/commands/commands.py
```diff
@@ -0,0 +1,16 @@
...
+        for prompt in prompts:
+            assert isinstance(prompt, str), "Prompts must be all strings."
```
sergiomsantos, line 10: same here

inductiva/simulators/simulator.py
```diff
@@ -54,6 +55,15 @@ def _setup_input_dir(self, input_dir: types.Path):
...
         return input_dir
 
+    def _commands_to_json(self, commands):
+
+        def to_cmd_json(x):
+            if isinstance(x, cmds.Command):
+                return x.to_json()
+            return cmds.Command(x).to_json()
+
+        return [to_cmd_json(cmd) for cmd in commands]
```
sergiomsantos, line 65:
```suggestion
    def _commands_to_json(self, commands):
        return [(x if isinstance(x, Command) else Command(x)).to_dict() for x in commands]
```

inductiva/tests/commands/test_commands.py
```diff
@@ -0,0 +1,29 @@
...
+    }),
+    ("command without prompts", [], {
+        "cmd": "command without prompts",
+        "prompts": []
+    }),
+])
+def test_to_json_method(cmd, prompts, expected_json):
+    cmd = commands.Command(cmd, *prompts)
+    result = cmd.to_json()
+    assert result == expected_json
```
sergiomsantos, line 20:
```suggestion
def test_to_json__with_prompts(cmd, prompts, expected_json):
    cmd = commands.Command("command with prompts", "y", "y")
    result = cmd.to_json()
    assert result == dict(cmd="command with prompts", prompts=("y", "y"))
    }
    
def test_to_json__without_prompts(cmd, prompts, expected_json):
    cmd = commands.Command("command with prompts", "y", "y")
    result = cmd.to_json()
    assert result == {
        "cmd": "command without prompts",
        "prompts": []
    }
```

inductiva/commands/commands.py
```diff
@@ -0,0 +1,16 @@
...
+
+        self.cmd = cmd
+        self.prompts = list(prompts)
+
+    def to_json(self):
+        return self.__dict__
```
sergiomsantos, line 16:
```suggestion
    def to_dict(self):
        return self.__dict__.copy()
```

inductiva/tests/commands/test_commands.py
sergiomsantos, line 1: add test to ensure Command(xx).to_dict() does not return Command(xx).__dict__

```
x = Command()
assert id(x.to_dict()) != id(x.__dict__)

```

inductiva/commands/commands.py
hpenedones, line 1: A while back, in the first implementation of these "commands", I asked David to make some basic security validations to prevent arbitrary command execution on the server side, e.g running "rm -rf /" or something, just because it's the command sent by the client... Can you guys have a look at the current status of that? Thanks! @sergiomsantos

inductiva/simulators/simulator.py
```diff
@@ -54,6 +55,12 @@ def _setup_input_dir(self, input_dir: types.Path):
                 f"The provided path (\"{input_dir}\") is not a directory.")
         return input_dir
 
+    def _commands_to_json(self, cmds):
```
sergiomsantos, line 58:
```suggestion
    def _commands_to_json(self, cmds: types.Commands):
```

inductiva/simulators/simulator.py
```diff
@@ -54,6 +55,12 @@ def _setup_input_dir(self, input_dir: types.Path):
...
+        return [
+            cmd.to_dict() if isinstance(cmd, commands.Command) else
+            commands.Command(cmd).to_dict() for cmd in cmds
```
sergiomsantos, line 61: keep the if/else statement in the same line (for easier reading):
```suggestion
            (cmd if isinstance(cmd, Command) else Command(cmd)).to_dict()
            for cmd in cmds
```

Also, we can import directly the Command class and reduce the line length.

inductiva/simulators/dualsphysics.py
sergiomsantos, line 1: Just for future reference!
At some point we should create these classes dynamically.
They are all very similar. They can be created using a Meta base.

inductiva/commands/commands.py
```diff
@@ -0,0 +1,18 @@
...
+        if not isinstance(cmd, str):
+            raise ValueError("cmd argument must be a string.")
+
+        if not all(isinstance(prompt, str) for prompt in prompts):
+            raise ValueError("Prompts must be all strings.")
+
+        self.cmd = cmd
+        self.prompts = list(prompts)
+
+    def to_dict(self):
```
sergiomsantos, line 17: missing docstrings in both ctor and this method.

sergiomsantos (commented):
looking good.
just some minor tweaks and we're done.

inductiva/simulators/simulator.py
AugustoPeres: We are using the google style guide lines which are [against importing classes](https://google.github.io/styleguide/pyguide.html#22-imports). Keeping the if in the same line is completely controlled by yapf. I have no say in the matter

sergiomsantos (approved):
just need to clarify the conversion of commands to a list of

PR #980, opened 2024-01-25
Task.kill now waits for 5s for the task to be killed.

Pbarbosa92:


inductiva/tasks/task.py
```diff
@@ -175,11 +176,35 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+
+        while task_status == models.TaskStatusCode.PENDINGKILL:
+
+            time.sleep(1)
+            task_status = self.get_status()
+            if time.time() - start_time > constants.MAX_TIME_WAITING_FOR_TASK:
+                logging.warning(
+                    "Task may not have been killed successfully. Task status: "
+                    "%s", task_status)
+                return
```
AugustoPeres, line 198: I am unsure about this. What happens if we want to kill 100 tasks:

```python
tasks_to_kill = [task_1, task_2, ...]
for task in tasks_to_kill:
    task.kill()
```

Will this hang until each task exists the `PENDINGKILL`? The API should be particularly good in handling large workflows, so this seems strange.

If the main issue is repeated `task.kill()` calls to the same task why not just:

```python
if task.status == models.TaskStatusCode.PENDINGKILL:
    logging.warning("Trying to kill task twice")
```

Pbarbosa92: It should time out after 5 seconds. That is the maximum time we wait (we can decrease that value).

If the issue is the fact that this can take too long, I can decrease both the sleep timer and the timeout.
Most of the time, we won't even enter the while loop. But sometimes it happens.

Do you think that doing a task.kill sleep and another task.kill would be enough?

inductiva/tasks/task.py
```diff
@@ -175,11 +176,35 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
         self._api.kill_task(path_params=self._get_path_params())
 
+        task_status = self.get_status()
+
+        start_time = time.time()
+
+        while task_status == models.TaskStatusCode.PENDINGKILL:
+
+            time.sleep(1)
+            task_status = self.get_status()
```
IvanPombo, line 193: The status here can change to Killed, and the if below can still raise a warning.

inductiva/tasks/task.py
```diff
@@ -175,11 +176,35 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+            if time.time() - start_time > constants.MAX_TIME_WAITING_FOR_TASK:
+                logging.warning(
+                    "Task may not have been killed successfully. Task status: "
+                    "%s", task_status)
+                return
+
+        if task_status == models.TaskStatusCode.KILLED:
+            logging.info("Task killed successfully. Task status: %s",
+                         task_status)
```
IvanPombo, line 202: This check needs to be right after the `get_status()` inside the while and if it is killed you break the loop. Otherwise, you will see check the time condition, which can raise a warning even if the status has changed.

IvanPombo (changes requested):
Left a few comments, than we can think about the reasoning behind @AugustoPeres.

inductiva/constants.py
```diff
@@ -5,3 +5,5 @@
                                     "ws://127.0.0.1:5000/logs")
 
 DEFAULT_QUEUE_MACHINE_TYPE = "c2-standard-4"
+
+MAX_TIME_WAITING_FOR_TASK = 10
```
sergiomsantos, line 9:
```suggestion
TASK_KILL_TIMEOUT_SEC = 10
```

inductiva/tasks/task.py
```diff
@@ -175,11 +176,35 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
     def kill(self) -> None:
```
sergiomsantos, line 176:
```
def kill(self, wait=False, timeout=5, maxretries=5) -> bool:
    self._api.kill_task(path_params=self._get_path_params())
    while maxretries > 0:
        maxretries -= 1
        self._api.kill_task(path_params=self._get_path_params())
        sleep(1)
    if not wait:
        return
    success = True
    start = time.now()
    status = get_status()
    while status == PENDINGKILL:
         if (time.now() - start) >= timeout:
           success = false
             log.error()
           break

        time.sleep(1)
        status = get_status
    return success
```

inductiva/tasks/task.py
```diff
@@ -172,13 +172,65 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
-    def kill(self) -> None:
-        """Kill the task.
-
-        This method issues a request to the API to kill the task. It doesn't
-        block waiting for confirmation if the task was killed.
+    def kill(self,
+             wait: bool = False,
+             timeout: float = 5,
+             maxretries: int = 5) -> bool:
+        """
```
IvanPombo, line 179: Here add a single line describing the functionality. The below you can write all you want describing further the functionality.

inductiva/tasks/task.py
```diff
@@ -172,13 +172,65 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+        Tries to kill a task. If wait is True (default is False)
+        the function sends the kill command and returns.
+        If wait is False the function checks for confirmation that
+        the kill command was successfull.
```
IvanPombo, line 183: Isn't the other way around? If I want to wait, I set it to True.

inductiva/tasks/task.py
```diff
@@ -172,13 +172,65 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+        while maxretries > 0:
+            maxretries -= 1
+            try:
+                self._api.kill_task(path_params=self._get_path_params())
+                break
+            except exceptions.ApiException as e:
+                logging.error("Error while sending the kill command: %s", e)
+                if maxretries == 0:
+                    logging.error(
+                        "Something went wrong while senting the kill"\
```
IvanPombo, line 201: typo *sending.

Below as well.

inductiva/tasks/task.py
```diff
@@ -172,13 +172,65 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+                            "command. Max retries reached."
+                    )
+                    return False
+                logging.error(
+                    "Something went wrong while senting the kill command."\
+                        "Trying again..."
```
IvanPombo, line 207: *Retrying

inductiva/tasks/task.py
```diff
@@ -172,13 +172,65 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+            return True
+
+        success = True
+        start = time.time()
+        status = self.get_status()
+
+        while status == models.TaskStatusCode.PENDINGKILL:
+            if (time.time() - start) >= timeout:
+                success = False
+                logging.error("Kill command timed out.")
```
IvanPombo, line 220: Even if it times out, the status can still change to Killed. Which would raise an error below, since you don't update it again after the loop.

So the `self.get_status()` should be before the time check.

IvanPombo (commented):
Left a few comments to improve readibility, thereafter it should be ready to go.

inductiva/tasks/task.py
Pbarbosa92: Yes. Thanks!

inductiva/tasks/task.py
```diff
@@ -172,13 +172,65 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+        Returns:
+            True if the kill command was successfull, False otherwise.
+        """
+        while maxretries > 0:
+            maxretries -= 1
+            try:
+                self._api.kill_task(path_params=self._get_path_params())
+                break
+            except exceptions.ApiException as e:
+                logging.error("Error while sending the kill command: %s", e)
```
sergiomsantos, line 198:
```suggestion
                logging.error("Failed to send the kill request for task %s. Request failed with the following error:" %s", task_id, e)
```

inductiva/tasks/task.py
```diff
@@ -172,13 +172,65 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
-
-        This method issues a request to the API to kill the task. It doesn't
-        block waiting for confirmation if the task was killed.
+    def kill(self,
+             wait: bool = False,
+             timeout: float = 5,
+             maxretries: int = 5) -> bool:
         """
-        self._api.kill_task(path_params=self._get_path_params())
+        Tries to kill a task.
```
IvanPombo, line 180: Change this to the first line, 179.

inductiva/tasks/task.py
```diff
@@ -172,13 +172,65 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+        If wait is False (default value)the function sends the
```
IvanPombo, line 181: Add a space: `(default value) the`

IvanPombo (commented):
Micro comments.

IvanPombo (approved):
LGTM! On QA now.

sergiomsantos (approved):

inductiva/tasks/task.py
```diff
@@ -172,13 +172,64 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+            if (time.time() - start) >= timeout:
+                success = False
+                break
+
+            time.sleep(1)
+
+        if status != models.TaskStatusCode.KILLED:
+            success = False
+            logging.error(
+                "Failed to kill %s task after %d seconds." /
```
IvanPombo, line 226: Take this bar "/" out, it bugs.

inductiva/tasks/task.py
```diff
@@ -172,13 +172,64 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+        Args:
+            wait: Determines if we wait for the kill command to
+                finish.
+            timeout: Time in seconds to wait for the kill command.
+            maxretries: Number of times to retry the kill command.
+        Returns:
+            True if the kill command was successfull, False otherwise.
         """
-        self._api.kill_task(path_params=self._get_path_params())
+        while maxretries > 0:
```
IvanPombo, line 191: We should check if the task is not in a terminal status. If it is, the endpoint to kill will bug, and print several messages like:

```bash
Something went wrong while sending the kill command.Retrying...
Something went wrong while sending the kill command.Retrying...
Something went wrong while sending the kill command.Retrying...
Something went wrong while sending the kill command.Retrying...
Something went wrong while sending the killcommand. Max retries reached.
False
```

When it should print, `Task has already terminated with status %s.`

inductiva/tasks/task.py
```diff
@@ -172,13 +172,64 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+            maxretries -= 1
+            try:
+                self._api.kill_task(path_params=self._get_path_params())
```
IvanPombo, line 194: Moreover, either we check that maxretries should be > -1 to handle one kill task at least, but still check that  `maxretries>=0`.

IvanPombo (changes requested):
This needs a few updates due to QA. Check my comments.

inductiva/tasks/task.py
Pbarbosa92: Done

inductiva/tasks/task.py
Pbarbosa92: I now raise a ValueError exception.

inductiva/tasks/task.py
Pbarbosa92: Done

inductiva/tasks/task.py
```diff
@@ -172,13 +172,70 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+        if status != models.TaskStatusCode.KILLED:
+            success = False
+            logging.error(
+                "Failed to kill %s task after %d seconds."
+                "The status of the task is %s", self.id, timeout, status)
+
+        if success:
+            logging.info("Successfully killed task %s.", self.id)
+
+        return success
```
sergiomsantos, line 238:
```suggestion
        success = False
        start = time.time()
        while (status := self.get_status()) == models.TaskStatusCode.PENDINGKILL \
            and (time.time() - start) < timeout:
            time.sleep(1)

        if status == models.TaskStatusCode.KILLED:
            logging.info("Successfully killed task %s.", self.id)
            success = True
        else:
            logging.error(
                "Failed to kill task %s after %d seconds."
                "The status of the task is %s", self.id, timeout, status)

        return success
```

inductiva/tasks/task.py
```diff
@@ -174,13 +175,73 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
 
             time.sleep(polling_period)
 
-    def kill(self) -> None:
-        """Kill the task.
+    def kill(
+            self,
+            wait_timeout: Optional[Union[float,
+                                         int]] = None) -> Union[bool, None]:
```
sergiomsantos, line 181: Isn't there a way to prevent these weird line breaks?

inductiva/tasks/task.py
```diff
@@ -174,13 +175,73 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+            None if we don't wait, True if the kill command was successful,
+            False if it was not.
 
-        This method issues a request to the API to kill the task. It doesn't
-        block waiting for confirmation if the task was killed.
         """
-        self._api.kill_task(path_params=self._get_path_params())
+        if not isinstance(wait_timeout, (float, int, type(None))) or (
+                wait_timeout is not None and wait_timeout <= 0.0):
+            raise ValueError("Wait timeout must be a positive number or None.")
```
sergiomsantos, line 198: we actually have2 types of exceptions here:
- TypeError `if not None and not isinstance(wait_timeout, (int, float))`
- ValueError if `wait_timeout < 0.0`

```suggestion
        if wait_timeout is not None:
            if not isinstance(wait_timeout, (float, int)):
                raise TypeError
            if wait_timeout < 0.0:
                raise ValueError
```

inductiva/tasks/task.py
```diff
@@ -174,13 +175,73 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+
+        maxretries = constants.MAX_API_RETRIES
+
+        while maxretries > 0:
+            maxretries -= 1
+            try:
+                task_status = self.get_status()
+
+                if task_status in _TASK_TERMINAL_STATUSES:
+                    break
```
sergiomsantos, line 208: see https://github.com/inductiva/inductiva/pull/980#discussion_r1467956087

move this logic to it's own method.

```python
if self._is_terminal():
    break
```

inductiva/tasks/task.py
```diff
@@ -174,13 +175,73 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+
+        if wait_timeout is None:
+            return None
+
+        success = True
+        start = time.time()
+        status = self.get_status()
+
+        while status == models.TaskStatusCode.PENDINGKILL:
+            status = self.get_status()
```
sergiomsantos, line 228: use the walrus operator to get the status in the if condition:

```suggestion
        while(status := self.get_status()) == models.TaskStatusCode.PENDINGKILL:
```

inductiva/tasks/task.py
```diff
@@ -174,13 +175,73 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+            if (time.time() - start) >= wait_timeout:
```
sergiomsantos, line 229:
```suggestion
            if (time.time() - start) > wait_timeout:
```

inductiva/tasks/task.py
```diff
@@ -174,13 +175,73 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+                success = False
+                break
+
+            time.sleep(1)
```
sergiomsantos, line 233: move to constant

inductiva/tasks/task.py
```diff
@@ -174,13 +175,73 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+                    break
+
+                path_params = self._get_path_params()
+                self._api.kill_task(path_params=path_params)
+                break
+            except exceptions.ApiException as exc:
+                if maxretries == 0:
+                    raise RuntimeError(
+                        "Something went wrong while sending"
+                        "the kill command. Try again Later") from exc
```
sergiomsantos, line 217:
```suggestion
                        " the kill command. Please try again later.") from exc
```

Also, beware of the spaces between strings that are to be concatenated.

inductiva/constants.py
```diff
@@ -5,3 +5,5 @@
                                     "wss://logs.inductiva.ai")
 
 DEFAULT_QUEUE_MACHINE_TYPE = "c2-standard-4"
+
+MAX_API_RETRIES = 5
```
sergiomsantos, line 9:
```suggestion
TASK_KILL_MAX_API_REQUESTS = 5
TASK_KILL_RETRY_SLEEP_SEC = 1.0
```

inductiva/tasks/task.py
```diff
@@ -174,13 +175,73 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+                success = False
+                break
+
+            time.sleep(1)
+
+        if status != models.TaskStatusCode.KILLED:
+            success = False
+            logging.error(
+                "Failed to kill %s task after %f seconds."
+                "The status of the task is %s", self.id, wait_timeout, status)
```
sergiomsantos, line 239:
```suggestion
            logging.error(
                "Unable to ensure that task %s transitioned to the KILLED state after %f seconds. "
                "The status of the task is %s", self.id, wait_timeout, status)
```

sergiomsantos (commented):
just some final comments

inductiva/tasks/task.py
```diff
@@ -174,13 +175,73 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+        request, and return none. If wait_timeout is a positive number,
+        the function will send the kill request and wait (wait_timeout
+        seconds) for the task to be killed. We return True if the task
+        was successfully killed, and False otherwise.
+        Args:
+            wait_timeout: number of seconds to wait for the kill command
+            or none if we don't want to wait.
+        Returns:
+            None if we don't wait, True if the kill command was successful,
+            False if it was not.
```
sergiomsantos, line 193:
```suggestion
        """Request a task to be killed.
        
        This method requests that the current task is remotely killed.
        If `wait_timeout` is None (default), the kill request is sent to the
        backend and the method returns. However, if `wait_timeout` is 
        a positive number, the method waits up to `wait_timeout` seconds
        to ensure that the task transitions to the KILLED state.

        Args:
            wait_timeout (int, float): Optional - number of seconds to wait for the
            kill command or None if only the request is to be sent.

        Returns:
            - None if `wait_timeout` is None and the kill request was
              successfully sent;
            - True if `wait_timeout`> 0 and the task successfully transitioned
              to the `KILLED` state within `wait_timeout` seconds;
            - False if `wait_timeout` > 0 but the task didn't transition
              to the `KILLED` state within `wait_timeout` seconds;
```

inductiva/tasks/task.py
Pbarbosa92: Yapf won't allow me to do it.

inductiva/tasks/task.py
```diff
@@ -169,18 +170,92 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
...
+                        "Something went wrong while sending"
+                        " the kill command. Please try again later.") from exc
+                time.sleep(constants.TASK_KILL_RETRY_SLEEP_SEC)
+
+        if wait_timeout is None:
+            return None
+
+        success = True
+        start = time.time()
+        status = self.get_status()
```
sergiomsantos, line 238: no need for the assignment. The walrus in the `while` condition already performs the assignment.

sergiomsantos (changes requested):
We're just missing unit tests to ensure that the kill method is performing as expected.
You can mock the api call to return whatever you want.

Pbarbosa92:
> We're just missing unit tests to ensure that the kill method is performing as expected. You can mock the api call to return whatever you want.

I've added the unit tests.

Pbarbosa92:
**QA**

@IvanPombo or @AugustoPeres Can you guys QA this code ?

You just need to test the function task.kill with or without timeout.

```
        Args:
            wait_timeout (int, float): Optional - number of seconds to wait
            for the kill command or None if only the request is to be sent.
        Returns:
            - None if `wait_timeout` is None and the kill request was
              successfully sent;
            - True if `wait_timeout`> 0 and the task successfully transitioned
              to the `KILLED` state within `wait_timeout` seconds;
            - False if `wait_timeout` > 0 but the task didn't transition
              to the `KILLED` state within `wait_timeout` seconds;

```

IvanPombo:
## QA Mode On

### Launch a simulation and kill the task with `wait_timeout=None`

No message received, as expected, and the task actually performed the kill. Check with `task.get_status()` afterwards.


### Launch a simulation and set a `wait_timeout=5`

Message received: `Successfully killed task m40p1gweikh8es7kw4eehmae1.` and the return was `True`.


### Launch a simulation and set a `wait_timeout=0.01`

Failed to ensure the kill, message received `Unable to ensure that task iwjj7tsfd0vrp0de0v0q71zuv transitioned to the KILLED state after 0.010000 seconds. The status of the task is pending-kill`. Return was `False`. 


In terms of QA, it is approved. However, I would still like to have a message when `wait_timeout` is None, just to confirm something was submitted.

PR #981, opened 2024-01-25
Add automatic detection of available updates with localization support

sergiomsantos:
Task inductiva/tasks#49 - Automatically check for available updates.

## Description
This PR introduces a mechanism for automatically checking for available package updates.
The check is performed every time the `inductiva` packge is initialized (typically via imports).
The mechanism uses a connection to the PyPi registry to identify the most recent available version of the package and compares it to the version of the package itself. If they are different, a notification is emitted with instructions on how to upgrade.

Simultaneously, this PR introduces a simple localization mechanism that can be used to select the language that is used to notify the user according to the system's locale.

### DIFF
--- a/inductiva/localization/localization.py
+++ b/inductiva/localization/localization.py
@@ -127,8 +127,8 @@ class Translator:
             *args: Positional substitutions for the Template constructor.
             **kwargs: Keyword substitutions for the Template constructor.
         """
-        transtation = self[key]
-        return Template(transtation, *args, **kwargs)
+        translation = self[key]
+        return Template(translation, *args, **kwargs)
 
     def __getitem__(self, key: str) -> str:
         """Get the translation for the given key in the active language.
@@ -147,5 +147,5 @@ class Translator:
 
 
 # pseudo-singleton object that is already initialized to the system's locale
-# or the pakcage's default language if the system's locale is not available.
+# or the package's default language if the system's locale is not available.
 translator = Translator()


inductiva/localization/localization.py
```diff
@@ -0,0 +1,143 @@
...
+            elif isinstance(other, tuple):
+                return self.format(*other)
+            return self.format(other)
+        
+        def __str__(self) -> str:
+            # hack to allow logging to work with this class.
+            # logging uses msg = str(self.msg) in getMessage() before applying
+            # the % operator. Here we return the object itself to avoid route
+            # the string formatting to the __mod__ method
+            return self
```
AugustoPeres, line 44: I am not sure that this is necessary. It is my understanding that this is just being used to format string. But we can use simpler python commands. The following works:

```python
>>> with open("../pt.json", "r") as f: translation = json.load(f)
... 
>>> translation
{'upgrade-available': ["A versão {0} da biblioteca 'inductiva' está disponível (a versão atualmente instalada é {1}).", 'Considere atualizar para a versão mais recente usando:', '  pip install --upgrade inductiva'], 'failed-update-check': ["Não foi possível verificar se há atualizações para a biblioteca 'inductiva'.", 'O seguinte erro ocorreu:\n  {0}']}
>>> translation['upgrade-available'][0].format("2", "3")
"A versão 2 da biblioteca 'inductiva' está disponível (a versão atualmente instalada é 3)."
```

inductiva/localization/localization.py
```diff
@@ -0,0 +1,143 @@
...
+                f"Will use '{_INDUCTIVA_DEFAULT_LANG}' as default language.")
+            return
+
+        with open(pathname, "r") as f:
+            translations = json.load(f)
+
+        for key, translation in translations.items():
+            if isinstance(translation, list):
+                translation = '\n'.join(translation)
+            translations[key] = self._Template(translation)
```
AugustoPeres, line 103: Same as above. I think that we can just use `.format()`

AugustoPeres (commented):
The version check looks good. I am only unsure about 1) thee need of translations 2) If we really need/want a translation I think that it could be a little bit simpler than the current implementation

inductiva/__init__.py
```diff
@@ -26,3 +26,31 @@
...
+                            headers={"Accept": "application/json"})
+        json_response = json.loads(resp.data)
+        new_version = json_response["info"]["version"]
+
+    except Exception as ex:  # pylint: disable=broad-exception-caught
+        logging.warning(t["failed-update-check"], ex)
+
+    if new_version != __version__:
+        msg = t["upgrade-available"].format(new_version, __version__)
+        sys.stderr.write(msg + "\n")
```
IvanPombo, line 53: This should write to the stdout, no? Notice, that this check doesn't mean things won't work, just that the user won't have the latest features. But scripts they had previously wrote, should still work.

inductiva/localization/localization.py
IvanPombo, line 1: This is cool, but is it something normal to add to packages? 
Is the translation happening automatically in general logging messages?

IvanPombo (commented):
Left a few comments to better understand the goal of the translator.

inductiva/localization/localization.py
sergiomsantos: The problem is to ensure we still provide both old-style `%` and lazy string formatting capabilities so that it works with the `logging` module while giving us some more flexibility
on how we define messages using an external file.

translations might not be necessary now, and they might no even be relevant in the future, but this mechanism gives us the tools to have a centralised location for messages.

In any case, your comment made me think a bit more and I simplified the code. Addressed in [5cb88bc](https://github.com/inductiva/inductiva/pull/981/commits/5cb88bc077487bf855d7d5df6090cb3d9fe952d5).

see also https://docs.python.org/3/howto/logging-cookbook.html#use-of-alternative-formatting-styles

inductiva/__init__.py
sergiomsantos: yes, you're correct. It is only printing to stdout so that redirections of the stdout to files do not have this message included. Ex:

```bash
$ python -c 'import inductiva; print(inductiva.__path__)' > tmp.txt
Version 0.4.0 of the 'inductiva' package is available (the currently installed version is 0.3.0).
Consider upgrading to the newer version using:
  pip install --upgrade inductiva
$ cat tmp.txt
['build/__editable__.inductiva-0.3.0-py3-none-any/inductiva']
```

inductiva/localization/localization.py
sergiomsantos: see https://github.com/inductiva/inductiva/pull/981#discussion_r1467053091.

Again, we might not need translations for now, but this can change, specially because we can have outside contributors. More importantly, it provides us with a mechanism to easily review messages without having to go through the code. If translations never become relevant, we can just rename the module and whatever components therein we find necessary.

inductiva/localization/localization.py
```diff
@@ -0,0 +1,151 @@
...
+        If the key doesn't exist in the default language, the key itself will
+        be used as the translation.
+
+        Args:
+            key (str): Key for the translation which will be used as formatting
+                string in the Template constructor.
+            *args: Positional substitutions for the Template constructor.
+            **kwargs: Keyword substitutions for the Template constructor.
+        """
+        transtation = self[key]
```
IvanPombo, line 130: Typo 
```suggestion
        translation = self[key]
```

inductiva/localization/localization.py
```diff
@@ -0,0 +1,151 @@
...
+        return Template(transtation, *args, **kwargs)
```
IvanPombo, line 131:
```suggestion
        return Template(translation, *args, **kwargs)
```

inductiva/localization/localization.py
IvanPombo: So the goal is to actually centralized the messaging in a single file and only use the keys?
I think we have briefly discussed this and is indeed interesting! The purpose of translations got me a bit lost, and still not sure if we should treat other languages since other libraries and simulators require English either way.

inductiva/localization/localization.py
```diff
@@ -0,0 +1,151 @@
...
+
+class Template:
+    """Template for lazy string formatting.
+
+    Formatting is performed using the str.format method only when a string
+    representation of the object is required (typically using str(obj)).
+    """
+
+    def __init__(self, fmt: str, /, *args, **kwargs):
+        """Intialize a Template object using the given formatting string
```
IvanPombo, line 22: This has a typo: `Initialize`.

AugustoPeres (approved):

IvanPombo (approved):
Fixing the typos and is good to go!

Pbarbosa92:
**QA**
I changed __version__  to "0.3.9" at Inductiva.__init__.py and got the following output:

```
A versão 0.4.0 da biblioteca 'inductiva' está disponível (a versão atualmente instalada é 0.3.9).
Considere atualizar para a versão mais recente usando:
  pip install --upgrade inductiva
Downloading from url https://storage.googleapis.com/inductiva-api-demo-files/reef3d-input-example.zip to the local path: [/Users/paulobarbosa/Documents/code/reef3d-input-example.zip](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/reef3d-input-example.zip)
---------------------------------------------------------------------------
ApiException                              Traceback (most recent call last)
File [~/Documents/Gits/inductiva/inductiva/api/methods.py:387](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/api/methods.py:387), in compare_client_and_backend_versions(client_version)
    [386](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/api/methods.py:386) try:
--> [387](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/api/methods.py:387)     api_instance.compare_client_and_backend_versions(
    [388](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/api/methods.py:388)         query_params=query_params)
    [390](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/api/methods.py:390) except (MaxRetryError, NewConnectionError) as exc:

File [~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:258](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:258), in CompareClientAndBackendVersions.compare_client_and_backend_versions(self, query_params, accept_content_types, stream, timeout, skip_deserialization)
    [250](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:250) def compare_client_and_backend_versions(
    [251](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:251)     self,
    [252](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:252)     query_params: RequestQueryParams = frozendict.frozendict(),
   (...)
    [256](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:256)     skip_deserialization: bool = False,
    [257](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:257) ):
--> [258](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:258)     return self._compare_client_and_backend_versions_oapg(
    [259](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:259)         query_params=query_params,
    [260](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:260)         accept_content_types=accept_content_types,
    [261](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:261)         stream=stream,
    [262](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:262)         timeout=timeout,
    [263](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:263)         skip_deserialization=skip_deserialization)

File [~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:202](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:202), in BaseApi._compare_client_and_backend_versions_oapg(self, query_params, accept_content_types, stream, timeout, skip_deserialization)
    [201](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:201) if not 200 <= response.status <= 299:
--> [202](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/client/paths/version_check/get.py:202)     raise exceptions.ApiException(status=response.status,
...
    [401](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/api/methods.py:401)     raise RuntimeError(e) from e
    [403](https://file+.vscode-resource.vscode-cdn.net/Users/paulobarbosa/Documents/code/~/Documents/Gits/inductiva/inductiva/api/methods.py:403) except Exception as e:

RuntimeError: Client version 0.3.9 is not compatible with API version 0.4.0.
Please update the client version.
```

Looks good!

PR #997, opened 2024-01-26
Bug fixed.

Pbarbosa92:
If we have no headers we don't want to enter this if.


IvanPombo (approved):
LGTM! Can you elaborate on what the bug was?

Pbarbosa92:
> LGTM! Can you elaborate on what the bug was?

When we have no headers our data will be empty (cause we do for x in headers). So, no headers = no data.
We had
```
if not headers:
    tabular_data=data
```
not headers when we have no headers will be true and we will end up replacing our tabular data with an empty dict.

inductiva/utils/format_utils.py
```diff
@@ -86,7 +86,7 @@ def get_tabular_str(tabular_data: Union[Mapping[str, Iterable[Any]],
         }
         #if we have no headers data will be empty.
         #So, we want our original tabular_data
-        if not headers:
+        if headers:
```
IvanPombo, line 89: Based on your bug comment, then we should check if there are headers before generating the data, no?

IvanPombo (changes requested):
Clarify me something first, I was a bit reactive at first...

inductiva/utils/format_utils.py
Pbarbosa92: We kinda don't need it because the  `for X in headers` will do nothing.

IvanPombo: Still... better to check first, because for example this bug would have been avoided :D

Can the data be a Mapping and not have headers?

IvanPombo (approved):
LGTM!

sergiomsantos (changes requested):
let's add some unit tests to the `get_tabular_str` method

inductiva/utils/format_utils.py
```diff
@@ -57,6 +57,30 @@ def apply_formatters(table_data: dict, formatters: dict):
     return table_data
 
 
+def get_tabular_data(
```
sergiomsantos, line 60: missing doc string.
You can move part of the docstring from `get_tabular_str` here (making sure to update what needs to be updated) and reference this function in `get_tabular_str` so that there is no need to have doc repetition.

sergiomsantos (commented):
missing docstring in `get_tabular_data`

Pbarbosa92:
**QA**

Build the Inductiva library locally and call the function get_tabular_str and pass a list of lists as rows and list of headers. Use formatters to format any column and check if it works.

```
rows=[["abc",10],["qwe",10]]
headers=["text","number"]
formatters={
     "text": lambda x: x.upper()
}
get_tabular_str(rows, headers, formatters)
```
Check to see if the formatters are being applied.

AugustoPeres:
> @IvanPombo or @AugustoPeres ready for QA

###QA ✅ 

Used the python code:

```python
import inductiva

rows = [["abc", 10], ["qwe", 10]]
headers = ["text", "number"]
formatters = {"text": lambda x: x.upper()}

print(inductiva.utils.format_utils.get_tabular_str(rows, headers, formatters))

print("\n======================\n")

print(inductiva.utils.format_utils.get_tabular_str(rows, [], formatters))

print("\n======================\n")

print(inductiva.utils.format_utils.get_tabular_str([], headers, formatters))

print("\n======================\n")

print(inductiva.utils.format_utils.get_tabular_str(rows, headers, {}))

print("\n======================\n")

print(
    inductiva.utils.format_utils.get_tabular_str(rows, headers,
                                                 {"number": lambda x: x**2}))

print("\n======================\n")
```

and got the expected output

```
text      number
------  --------
ABC           10
QWE           10

======================

---  --
abc  10
qwe  10
---  --

======================

text    number
------  --------

======================

text      number
------  --------
abc           10
qwe           10

======================

text      number
------  --------
abc          100
qwe          100

======================
```

IvanPombo:
# QAing

Your workflow works as expected! I have just asked for the `apply_formatters` to not be working inplace, since we still may want the original data to apply other formatters to it. @Pbarbosa92 already made that change. So everything is validated and ready to merge!

sergiomsantos (approved):

PR #983, opened 2024-01-25
First iteration on Inductiva's Home page

maiyahershey:
This PR addresses documentation, style, text, and changes on Inductiva's Home.

In this iteration, I completely revamped the text, showcased the target audience, the problem and solution inductiva is tackling, and how it works.

The text might have some weaknesses in the technical references, the GIF was an experiment to see how it works and looks, and if adding an infographic brings anything new to the table. 

**Feedback welcome!!**

PR #1000, opened 2024-01-29
Getting Started section 

maiyahershey:
In this section, I replaced Install by Getting Started, added a new layout:
1- Overview of Inductiva, intended audience, and a brief of what it could do
2- Prerequisites
3-  Some features
4- What to read next
5- some recommended sections to add, such as FAQs or Terminology

The links dont lead anywhere but they are for preview purposes, in the future we replace them with the respective links within the read the docs page
Other suggestions could be removed or replaced, but this is a first draft and serves as an informative guide for the reader to set expectations from such project.

Feedback Welcome! 🌞

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -1,17 +1,18 @@
-# Getting Started
-Welcome to Inductiva, a Python API package that simplifies running large-scale 
-physical system simulations in the cloud. Just few lines of code unleash its potential, freeing 
-you from local hardware limits and the technical hassles of scaling simulations. 
+# Getting Started with Inductiva
+Welcome to Inductiva, a Python API package that enables running popular [open-source physical system simulation packages]() in the Cloud. Experience HPC-standard scalability, optimized performance, and cost-efficiency with minimal setup, all through Inductiva's streamlined interface.
 
 ### What We'll Cover:
 
-* [Inductiva API Installation]()
+* [Installing Inductiva API]()
 * [What to read next]()
-* [Troubleshooting]()
-* [Frequently Asked Questions (FAQs)]()
-* [Useful Terminology]()
 
-## Inductiva API Installation
+
+## Installing Inductiva API with pip
+You will need Python 3.8 or higher installed:
+
+```
+pip install --upgrade pip
+```
 
 To install the latest Inductiva package release on PyPI with pip, just open your 
 terminal and run:
@@ -19,10 +20,6 @@ terminal and run:
 ```
 pip install --upgrade inductiva
 ```
----
-
-> **Tip:** To ensure smooth installation and operation of our API, we recommend 
-keeping your Python package manager, Pip, up-to-date. Visit [Pip's official installation guide](https://pip.pypa.io/en/stable/installation/) for detailed instructions on updating Pip.
 
 
 Encountering issues? Don’t worry. Head over to [Troubleshooting]() to work it out.
@@ -31,62 +28,14 @@ Otherwise, you're all set! You can now decide [what to read next]().
 
 ## What to read next
 
-You can start experimenting with Inductiva by choosing from a variety of ready-to-use, 
-[open-source simulators]() that are immediately available for use on our default 
-machines.
+[Run your existing simulations through Inductiva API]().
 
-You can set up custom, dedicated machine groups for more complex or [larger-scale simulation]() 
-tasks, tapping into the more extensive [computational resources]() of Inductiva.
+[Scale your simulation by customizing your hardware with Inductiva API]().
 
-You can get inspired by how other researchers have used Inductiva API to configure, 
-execute, and scale their simulations with minimal coding effort by visiting our 
-[example gallery]().
+[Get inspired]() by how other researchers have used Inductiva API to configure, 
+run, and scale their own simulations.
 
-You can troubleshoot installation problems that you might encounter with Inductiva 
-API by checking [troubleshooting](#troubleshooting), [FAQs](), or [getting in touch]() 
+Troubleshoot installation problems that you might encounter with Inductiva 
+API by checking our [troubleshooting guide](#troubleshootingguide), [FAQs](), or [getting in touch]() 
 with our support team directly.
 
-## Troubleshooting
-
-If installing the package fails, you can retry it on a new Python virtual environment. 
-A [virtual environment](https://docs.python.org/3/library/venv.html) allows you to 
-have a fresh Python environment with isolated dependencies. In your shell, run:
-
-```
-python -m venv <venv>
-```
-
-In that command, you should replace `<venv>` with the path (*e.g.*, `.venv`) in 
-which you would like to create the environment. Then, to activate the environment 
-(again, correctly replacing `<venv>`), run:
-
-For `bash`/`zsh`:
-
-```
-source <venv>/bin/activate
-```
-
-For `cmd.exe` (Windows):
-
-```
-<venv>\Scripts\activate.bat
-```
-
-For `PowerShell` (Windows):
-```
-<venv>\Scripts\Activate.ps1
-```
-
-After activating the virtual environment, you can install the package as described 
-below:
-
-```
-pip install --upgrade inductiva
-```
-
-## Frequently Asked Questions
-This section could be removed or updated later.
-## Useful Terminology
-This documentation uses the following terms:
-
-https://packaging.python.org/en/latest/glossary/ 
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -1,6 +1,17 @@
-# Installing Inductiva
+# Getting Started
+some text here to intoduce this section
 
-Inductiva package is simple to install, just run on your terminal:
+## Prerequisites
+You will need [Python](https://www.python.org/) 3.8 or higher installed. If you do not have [Python](https://www.python.org/) installed already, here are the instructions to [download Python](https://www.python.org/downloads/) and install it.
+
+> **Tip:** New versions of Python are released annually in October, and it can take a few months for the scientific Python ecosystem to catch up. If you have trouble installing plasmapy on the most recent Python version between October and March, then try installing it on the second most recent version.
+
+
+
+## Installation
+
+To install the latest Inductiva package release on PyPI with pip, just open your 
+terminal and run:
 
 ```
 pip install --upgrade inductiva
@@ -8,6 +19,9 @@ pip install --upgrade inductiva
 
 This will provide the core functionalities of the API, which allows you to submit 
 jobs, control machines, and run simulations.
+
+**You're all set!** Visit the [example gallery]() to get inspired on the many ways you can use Inductiva. It 
+takes about 10 minutes to browse.
 
 For any issues with the installation, see [Troubleshooting](#troubleshooting).
 
@@ -48,3 +62,5 @@ below:
 ```
 pip install --upgrade inductiva
 ```
+
+## Frequently Asked Questions
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -1,14 +1,97 @@
 # Getting Started
-some text here to intoduce this section
+Welcome to Inductiva, where large-scale simulations meet simplicity, through a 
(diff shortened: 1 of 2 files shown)


docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
+# Getting Started
+Welcome to Inductiva, where large-scale simulations meet simplicity, through a
```
sarmento, line 2: I would be even more straight to the point, and reduce adjectives as much as possible.
Something like:

Welcome to Inductiva, a Python API for running large-scale simulations on the Cloud.

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
+# Getting Started
+Welcome to Inductiva, where large-scale simulations meet simplicity, through a 
+seamless, user-friendly interface that brings cutting-edge computational resources 
+to your fingertips—effortlessly.
+
+This section is designed to provide you with all the necessary information you 
+need to get started with Inductiva API. Here, we've condensed everything so you
```
sarmento, line 7: I like the first sentence, but I would remove the second sentence. No need for that.

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+Welcome to Inductiva, where large-scale simulations meet simplicity, through a 
+seamless, user-friendly interface that brings cutting-edge computational resources 
+to your fingertips—effortlessly.
+
+This section is designed to provide you with all the necessary information you 
+need to get started with Inductiva API. Here, we've condensed everything so you 
+can install Inductiva, grab usage tips, and find some of its essential features, 
+with ease.
+
+It's all you need to get familiar and make the most out of Inductiva from the
```
sarmento, line 11: I would also remove this sentence for now. Let's go straight to the point.

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+need to get started with Inductiva API. Here, we've condensed everything so you 
+can install Inductiva, grab usage tips, and find some of its essential features, 
+with ease.
+
+It's all you need to get familiar and make the most out of Inductiva from the 
+get-go.
+
+In this section you will find:
+
+* [Overview]()
```
sarmento, line 16: Love this!

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+* [Installing Inductiva]()
+* [What to read next]()
+* [Troubleshooting]()
+* [Frequently Asked Questions (FAQs)]()
+* [Useful Terminology]()
+
+
+## Overview
+### What is Inductiva?
+Inductiva is a state-of-the-art Python package that simplifies running large-scale
```
sarmento, line 31: remove adjectives as much as possible (e.g. "state-of-the-art").

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+* [Frequently Asked Questions (FAQs)]()
+* [Useful Terminology]()
+
+
+## Overview
+### What is Inductiva?
+Inductiva is a state-of-the-art Python package that simplifies running large-scale 
+physical system simulations. Just two lines of code unleash its potentional, freeing 
+you from local hardware limits and the technical hassles of scaling simulations. 
+It's all about making big computational tasks a breeze.
```
sarmento, line 34: I like the idea, but I would go "dryer". Let's try to quantify. @sergiomsantos : it would be great to talk about number.

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Who is it for?
+Inductiva is perfect for curious minds in research, science, and engineering who
```
sarmento, line 37: I am not sure this is about "curious minds" :)

Inductiva helps scientist and engineers scaling ....

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+## Overview
+### What is Inductiva?
+Inductiva is a state-of-the-art Python package that simplifies running large-scale 
+physical system simulations. Just two lines of code unleash its potentional, freeing 
+you from local hardware limits and the technical hassles of scaling simulations. 
+It's all about making big computational tasks a breeze.
+
+### Who is it for?
+Inductiva is perfect for curious minds in research, science, and engineering who 
+are engaged in intensive computational work. Whether you're unraveling the mysteries
```
sarmento, line 38: I think this stile would work great for the Genesis program, which is more targeted at students. I think here we need to talk more about optimal design or generating synthetic data at scale.

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+1. Visit the [Python download page](https://www.python.org/downloads/).
+2. Select and download the appropriate Python version for your operating system.
+3. Follow the installation instructions to set up Python on your system.
+
+> **Tip:** Python updates are released annually in October. It's worth noting that 
+it may take a few months for the scientific Python ecosystem to fully adapt to new 
+releases. Therefore, for the best compatibility with Inductiva, consider using a 
+Python version that has been stable for a few months.
+
+### Visualization Tools
```
sarmento, line 85: I think this is extra for now. Let's assume users know enough about this.

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+### Why choose Inductiva API?
+Because choosing Inductiva means opting for simplicity and efficiency in your 
+simulation work. With just two lines of code, you can activate its powerful features. 
+No need to tangle with technical headaches or a web of dependencies. Inductiva 
+streamlines the process, so can you cut through the clutter and focus on what 
+really matters – *the science*.
+
+## Prerequisites
+
+### Python Installation
```
sarmento, line 72: I would not add this to the main page.
Our users will be python users almost for sure, or will have familiarity with it.
I think this could be part of a "Trouble Shooting" section, but I would not use such valuable real estate for this.

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+
+1. Download the results from Inductiva.
+2. Utilize any visualization tool of your choice to analyze and interpret the data. 
+For large-scale, complex 3D visualizations, our team recommends open-source tools 
+like [ParaView](https://www.paraview.org/download/) or [VTK](https://vtk.org/download/).
+
+## Installing Inductiva API
+
+To install the latest Inductiva package release on PyPI with pip, just open your 
+terminal and run:
```
sarmento, line 97: should we ask for people to update pip before (@sergiomsantos : I was having that trouble)

docs/Getting Started.md
```diff
@@ -0,0 +1,171 @@
...
+
+```
+pip install --upgrade inductiva
+```
+
+This will provide the core functionalities of the API, which is designed to facilitate
```
sarmento, line 103: Do we need this here. This is actually a good paragraph for "What is Inductiva API".

sarmento (commented):
Maya: this is a great push fwd.
This is a good opportunity to talk about the target audience for this repo and also the "tone" we want to have for this document.

docs/Getting Started.md
maiyahershey: for a nice comparison, you can check the style/tone of this section vs the home page iteration PR: https://github.com/inductiva/inductiva/pull/983
This one has a completely different style, so we can pick and choose what we like!

sarmento: This is great, because of the Genesis program (I will tell you what it is) we will need this style. Don't loose track of this style because we will need it elsewhere.

docs/Getting Started.md
```diff
@@ -0,0 +1,92 @@
...
+
+To install the latest Inductiva package release on PyPI with pip, just open your 
+terminal and run:
+
+```
+pip install --upgrade inductiva
+```
+---
+
+> **Tip:** To ensure smooth installation and operation of our API, we recommend
```
sarmento, line 24: If you add the previous line, you don't need this.

docs/Getting Started.md
```diff
@@ -0,0 +1,92 @@
...
+keeping your Python package manager, Pip, up-to-date. Visit [Pip's official installation guide](https://pip.pypa.io/en/stable/installation/) for detailed instructions on updating Pip.
+
+
+Encountering issues? Don’t worry. Head over to [Troubleshooting]() to work it out.
+
+Otherwise, you're all set! You can now decide [what to read next]().
+
+## What to read next
+
+You can start experimenting with Inductiva by choosing from a variety of ready-to-use,
```
sarmento, line 34: Let's maybe call each of these something.

This one is "Running your own simulation through Inductiva API" (see below).

sarmento: I think it's not really "choosing a simulator". You have a few simulators available -- 7 or 8 -- and then you can run *your* simulations scripts prepared for those simulators. So, we need to put the focus on the fact that users already are running simulations of their own. What they will be able to do is to run their existing simulations using the Inductiva API.

So, the very first section is about exactly that: Running your own simulation through Inductiva API

sarmento: Once we show the user how to run their existing simulations via the API we can then show how to scale them.

docs/Getting Started.md
```diff
@@ -0,0 +1,92 @@
...
+
+Otherwise, you're all set! You can now decide [what to read next]().
+
+## What to read next
+
+You can start experimenting with Inductiva by choosing from a variety of ready-to-use, 
+[open-source simulators]() that are immediately available for use on our default 
+machines.
+
+You can set up custom, dedicated machine groups for more complex or [larger-scale simulation]()
```
sarmento, line 38: So, this is about "Choosing your Hardware" or maybe about "Scaling your Simulation"

The title here should be an action point for the user.

docs/Getting Started.md
```diff
@@ -0,0 +1,92 @@
...
+## What to read next
+
+You can start experimenting with Inductiva by choosing from a variety of ready-to-use, 
+[open-source simulators]() that are immediately available for use on our default 
+machines.
+
+You can set up custom, dedicated machine groups for more complex or [larger-scale simulation]() 
+tasks, tapping into the more extensive [computational resources]() of Inductiva.
+
+You can get inspired by how other researchers have used Inductiva API to configure,
```
sarmento, line 41: Do we have this? @IvanPombo ?

sarmento: Or is this about the other repos? (it could be!)

docs/Getting Started.md
```diff
@@ -0,0 +1,92 @@
...
+
+You can get inspired by how other researchers have used Inductiva API to configure, 
+execute, and scale their simulations with minimal coding effort by visiting our 
+[example gallery]().
+
+You can troubleshoot installation problems that you might encounter with Inductiva 
+API by checking [troubleshooting](#troubleshooting), [FAQs](), or [getting in touch]() 
+with our support team directly.
+
+## Troubleshooting
```
sarmento, line 49: I would make this a separate document.

docs/Getting Started.md
```diff
@@ -0,0 +1,92 @@
+# Getting Started
+Welcome to Inductiva, a Python API package that simplifies running large-scale 
+physical system simulations in the cloud. Just few lines of code unleash its potential, freeing
```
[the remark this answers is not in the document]
maiyahershey: Excellent, let me edit a few bits

docs/Getting Started.md
maiyahershey: so this section should be able to act as a standalone that would link you to different parts of the repo, could be other projects. I used the play of words but can make it more explicit with titles. 
so the logic or journey is:
1- you understand and learn what inductiva is and who it is for
2- you install it 
3- you then go to "what to read next" to choose what you want to do with it. 
it is for ease of navigation. looked like good practice on other repos.

docs/Getting Started.md
sergiomsantos: I think this should go to the troubleshooting section.
Otherwise we should also say that the user should be using an up-to-date version of python, etc.

maiyahershey: i really agree with this sergio, seems best practice on all repos i've seen so far

sarmento (commented):
LGTM!

sarmento (approved):
LGTM!

PR #1008, opened 2024-01-29
Makes `working_dir` and `output_dir` ContextVariables

AugustoPeres:
This pull request makes `working_dir` and `output_dir` ContextVariables. This makes setting the `working_dir` and `output_dirs` thread safe. Take a look at [this comment](https://github.com/inductiva/tasks/issues/24#issuecomment-1914556893).

### QA testing

This can be tested by launching simulations setting different `output/working_dir`s. For example, on my side I used:

```python
import inductiva

mg = inductiva.resources.MachineGroup(
    machine_type="c2-standard-4",
    num_machines=1,
    disk_size_gb=70,
)
mg.start()

    "splishsplash-input-example.zip",
    unzip=True)

splishsplash = inductiva.simulators.SplishSplash()

task = splishsplash.run(input_dir=input_dir,
                        on=mg,
                        sim_config_filename="config.json")

task.wait()
task.download_outputs()

inductiva.set_output_dir("dummy_dir")
task.download_outputs()

# Changing working dir
inductiva.set_working_dir(input_dir)
task = splishsplash.run(input_dir="./",
                        on=mg,
                        sim_config_filename="config.json")

task.wait()
task.download_outputs()
```

For this particular case, you should observe that the tasks are downloaded first to the default `output_dir` and then to the directory `dummy_dir`. Finally, because we then set `working_dir` to `input_dir="splish_splash_input_example"` you should see the final task being downloaded to `"splish_splash_input_example"/dummy_dir`.

On your own testing you can play around with this/simulator type and the behaviour should be maintained.


inductiva/__init__.py
```diff
@@ -15,10 +16,12 @@
...
 
 api_url = os.environ.get("INDUCTIVA_API_URL", "https://api.inductiva.ai")
-output_dir = os.environ.get("INDUCTIVA_OUTPUT_DIR", "inductiva_output")
+_output_dir = contextvars.ContextVar("INDUCTIVA_OUTPUT_DIR")
+_output_dir.set(os.environ.get("INDUCTIVA_OUTPUT_DIR", "inductiva_output"))
 api_key = os.environ.get("INDUCTIVA_API_KEY")
 
-working_dir = None
+_working_dir = contextvars.ContextVar("INDUCTIVA_WORKING_DIR")
+_working_dir.set(None)
```
sergiomsantos, line 24: let's remove all related to `_working_dir`

inductiva/tasks/task.py
```diff
@@ -231,7 +231,7 @@ def download_outputs(
                 files are downloaded.
             output_dir: Directory where to download the files. If None, the
                 files are downloaded to the default directory. The default is
-                {inductiva.working_dir}/{inductiva.output_dir}/{task_id}.
+                {inductiva._working_dir}/{inductiva._output_dir}/{task_id}.
```
sergiomsantos, line 234: The default should be:

```suggestion
                {inductiva.get_output_dir()}/{output_dir}/{task_id}
```

if `output_dir` is None, files are written to:

```suggestion
                {inductiva.get_output_dir()}/{task_id}
```

if `inductiva.get_output_dir() is None` and `output_dir is not None`:

```suggestion
                ./{output_dir}/{task_id}
```

if bot are None:

```suggestion
                ./{task_id}
```

inductiva/utils/files.py
```diff
@@ -34,7 +34,7 @@ def resolve_path(path: Optional[types.Path]) -> pathlib.Path:
     """Resolve a path relative to the Inductiva package working directory.
 
     This takes in a path and returns it resolved relative to the
-    "inductiva.working_dir" setting.
+    "inductiva._working_dir" setting.
```
sergiomsantos, line 37:
```suggestion
    "inductiva.get_output_dir()" setting.
```

inductiva/utils/files.py
```diff
@@ -49,8 +49,8 @@ def resolve_path(path: Optional[types.Path]) -> pathlib.Path:
     """
     resolved_path = pathlib.Path.cwd()
 
-    if inductiva.working_dir:
-        resolved_path = pathlib.Path(inductiva.working_dir)
+    if inductiva.get_working_dir():
+        resolved_path = pathlib.Path(inductiva.get_working_dir())
```
sergiomsantos, line 53:
```suggestion
    if inductiva.get_output_dir():
        resolved_path = pathlib.Path(inductiva.get_output_dir())
```

inductiva/tasks/task.py
```diff
@@ -250,8 +250,8 @@ def download_outputs(
         response = api_response.response
 
         if output_dir is None:
-            output_dir = files.resolve_path(inductiva.output_dir).joinpath(
-                self.id)
+            output_dir = files.resolve_path(
+                inductiva.get_output_dir()).joinpath(self.id)
```
sergiomsantos, line 254:
```suggestion
            output_dir = files.resolve_path(output_dir).joinpath(self.id)
```

sergiomsantos (changes requested):
Some clarifications:
1. Let's remove the `working_dir` variable altogether. It will make everything easier.
2. `files.resolve_path` should, perhaps, be renamed to `files.resolve_output_path` to ensure that only the output is relative to `inductiva.get_output_dir()`.
3. Add a test to ensure that `inductiva.get_output_dir()` is thread-safe.

IvanPombo:
@sergiomsantos @AugustoPeres thinking about this a bit more and reflecting on how this will be used, I think that both `working_dir` and `output_dir` serve for when launching multiple simulations. And from this and its complexity, two keys aspects surge:

1 - This is doing nothing more than a join of paths between the current directory/`working_dir` and a path we pass. Notice that, while the current directory is the absolute path, when attaching the `working_dir` (or the `output_dir` for that matter) it is a relative path to current directory. So in general, this is already not working as expected. That is, when passing any extra resolving directory (`working_dir` and/or `output_dir`) this need to be in the current working dir. This is irrelevant, since users can do this with simple variables on their side, and manage the paths as they want.
2 - The main workflow for this variables is for the management of many simulations within the same script. This is technically what we want to do with the concept of `projects`, and there it will make sense to set the `working_dir` and/or `output_dir` for managing everything.

Henceforth, I think the general concept of these variables doesn't make sense moving forward, and the `projects` will achieve the same goal.

AugustoPeres:
> @sergiomsantos @AugustoPeres thinking about this a bit more and reflecting on how this will be used, I think that both `working_dir` and `output_dir` serve for when launching multiple simulations. And from this and its complexity, two keys aspects surge:
> 
> 1 - This is doing nothing more than a join of paths between the current directory/`working_dir` and a path we pass. Notice that, while the current directory is the absolute path, when attaching the `working_dir` (or the `output_dir` for that matter) it is a relative path to current directory. So in general, this is already not working as expected. That is, when passing any extra resolving directory (`working_dir` and/or `output_dir`) this need to be in the current working dir. This is irrelevant, since users can do this with simple variables on their side, and manage the paths as they want. 2 - The main workflow for this variables is for the management of many simulations within the same script. This is technically what we want to do with the concept of `projects`, and there it will make sense to set the `working_dir` and/or `output_dir` for managing everything.
> 
> Henceforth, I think the general concept of these variables doesn't make sense moving forward, and the `projects` will achieve the same goal.

@IvanPombo @sergiomsantos At the moment I think I side with Ivan and before completely removing `working_dir` we need to think more deeply about those implications. I will address the rest of the comments.

IvanPombo:
## QAing

To QA these changes I have started from your script, and everything was in its respective folder as the script mentions.

As a second step, I have attempted to `inductiva.set_output_dir(something)` before the run, and indeed, the `task_metadata.json` goes to that folder. 

As a third attempt, since I was getting confused with something, I tried to not set the output dir, but select the `output_dir` within the `download_outputs(output_dir="LEGO")`. From this, I expected to obtain the downloaded data within `LEGO` on my cwd, and nothing else. However, now we are obtaining the data within `LEGO/task_id`. This doesn't make sense, since it means that the `output_dir` in the `download_outputs` has the same role as `set_output_dir`.

Before, this only occurred when `output_dir` was None in `download_outputs`.

So I get mixed feelings, and I only semi-approve these changes.

AugustoPeres:
> As a third attempt, since I was getting confused with something, I tried to not set the output dir, but select the `output_dir` within the `download_outputs(output_dir="LEGO")`. From this, I expected to obtain the downloaded data within `LEGO` on my cwd, and nothing else.

This should now go to `inductiva_outpts/tLEGO`, no?

IvanPombo:
> > As a third attempt, since I was getting confused with something, I tried to not set the output dir, but select the `output_dir` within the `download_outputs(output_dir="LEGO")`. From this, I expected to obtain the downloaded data within `LEGO` on my cwd, and nothing else.
> 
> This should now go to `inductiva_outpts/tLEGO`, no?

Just "LEGO". In particular, if we do `inductiva.set_output_dir("Awesome")` then we would have the data go to `Awesome\LEGO` and not `Awesome\LEGO\<task_id>`.

AugustoPeres:
> > > As a third attempt, since I was getting confused with something, I tried to not set the output dir, but select the `output_dir` within the `download_outputs(output_dir="LEGO")`. From this, I expected to obtain the downloaded data within `LEGO` on my cwd, and nothing else.
> > 
> > 
> > This should now go to `inductiva_outpts/tLEGO`, no?
> 
> Just "LEGO". In particular, if we do `inductiva.set_output_dir("Awesome")` then we would have the data go to `Awesome\LEGO` and not `Awesome\LEGO\<task_id>`.

Addressed these comments

inductiva/__init__.py
```diff
@@ -30,6 +30,14 @@
 __version__ = "0.4.1"
 
 
+def set_output_dir(new_output_dir):
+    _output_dir.set(new_output_dir)
+
+
+def get_output_dir():
+    return _output_dir.get()
```
sergiomsantos, line 38: missing docstrings

inductiva/tests/test_output_dir_thread_safety.py
```diff
@@ -0,0 +1,25 @@
...
+    """Tests that threads do not make definite changes to context."""
+    inductiva.set_output_dir("test_dir")
+    output_dir = inductiva.get_output_dir()
+
+    num_threads = 5
+    output_dirs = [f"output_dir_{i}" for i in range(num_threads)]
+
+    with ThreadPoolExecutor(max_workers=num_threads) as executor:
+        futures = [
+            executor.submit(inductiva.set_output_dir, output_dirs[i])
```
sergiomsantos, line 17: we also need to check that the output dir is set to the correct value in each thread. Something along the lines:

```suggestion
            def test(out_dir):
                inductiva.set_output_dir(out_dir)
                assert inductiva.get_output_dir() == out_dir
                sleep some ms here
            executor.submit(test, output_dirs[i])
```

Also, try running with a bit more threads and adding a random sleep (tens to hundreds of  ms) in each test call to scramble things a bit and eventually discover potential concurrency issues.

sergiomsantos (commented):
almost there. We're just missing the doc strings for the get/set_output_dir methods.

IvanPombo:
# QA Approval

Script used: 
```python
import inductiva

    "splishsplash-input-example.zip",
    unzip=True)

splishsplash = inductiva.simulators.SplishSplash()

task = splishsplash.run(input_dir=input_dir,
                        sim_config_filename="config.json")

task.wait()
task.download_outputs()   # inductiva_output/task_id

inductiva.set_output_dir("dummy_dir")
task.download_outputs() # dummy_dir/task_id

inductiva.set_output_dir(None)
task = splishsplash.run(input_dir=input_dir,
                        sim_config_filename="config.json")

task.wait()
task.download_outputs() #task_id

task = splishsplash.run(input_dir=input_dir,
                        sim_config_filename="config.json")

task.wait()
task.download_outputs(output_dir="dummy_inside_dir") #dummy_inside_dir
```

PR #1016, opened 2024-01-31
Get_status was calling is_terminal and is_terminal was calling get_status

Pbarbosa92:


IvanPombo (approved):
LGTM! We may need to rethink all of this in the future, but for now it is okay to move forward.

PR #1018, opened 2024-01-31
Removes references to `inductiva.api_key`

AugustoPeres:
The exception message becomes:

```
No API Key specified. Set the INDUCTIVA_API_KEY environment variable.
  Exiting with code 1.
```


IvanPombo (commented):
Can you check the problem that was occuring also when using the CLI without a key set up?

AugustoPeres:
> Can you check the problem that was occuring also when using the CLI without a key set up?

Yes, fixed now

sergiomsantos (approved):

PR #1019, opened 2024-01-31
Getting Started final iteration

maiyahershey:
In this PR:
- Shortened intro, added tutorial with clear process steps
- Added missing steps such as API access token
- Added placeholders for example and testing the API
- Made some edits on the read next section
- Moved things around to Troubleshooting

Feedback Welcome!

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -28,11 +28,10 @@ export INDUCTIVA_API_KEY="YOUR_API_KEY"
 
 
 ## Install the Inductiva API with pip
-Now that you've set up the API access token, you're ready to install the the 
+Now that you've set up the API access token, you're ready to install the 
 latest Inductiva package release on PyPI with pip. 
 
-Simply open your 
-terminal and run:
+Simply open your terminal and run:
 
 ```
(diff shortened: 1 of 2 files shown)


docs/Getting Started.md
```diff
@@ -1,39 +1,63 @@
...
+before running your first simulation project. Should you encounter any errors or 
+issues along the way, our [troubleshooting guide]() is readily available to assist 
+you in resolving them efficiently.
 
-* [Installing Inductiva API]()
-* [What to read next]()
+### Steps We'll Cover:
 
+1. [Request Your API Access Token]()
+2. [Install the Inductiva API]()
```
sarmento, line 13: I would do reverse the order. 1 <-> 2

Let's assume that the person has already requested the key, so it's just a matter of configuring it. 
We will still leave the link to request the key if you don't have it. But I think if we have the key request first, we will loose people because our onboarding is still pretty bad... :(

And actually, we will be "sending" keys to people for now...

sarmento (approved):
LGTM!

sergiomsantos (approved):

PR #1021, opened 2024-01-31
Add first example to Getting-started

IvanPombo:
This adds a first example with Reef3d, the simplest simulator, to the Getting-started readme.

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -41,9 +41,32 @@ export INDUCTIVA_API_KEY="YOUR_API_KEY"
 
 ## Verify the Installation with a Test Run
 Next, make sure everything is set up correctly with a quick test run.
-## Project Example
+
+### Running your first example simulation
 Finally, check out this Project Example to demonstrate how you can run simulations 
 through the Inductiva API.
+
+For the first example, you will run a classical dam break simulation with the
+REEF3D simulator. 
+
+```python
+import inductiva
+
+# Download the configuration files for a REEF3D simulation
+input_dir = inductiva.utils.download_from_url(
+    "https://storage.googleapis.com/inductiva-api-demo-files/"
+    "reef3d-dambreak-dir.zip", unzip=True)
+
+# Initialize the REEF3D simulator object
+simulator = inductiva.simulators.REEF3D()
+
+# Launch the simulation with the downloaded configuration files
+task = simulator.run(input_dir=input_dir)
+
+# Wait for the simulation to finish and download the outputs
+task.wait()
+task.download_outputs()
+```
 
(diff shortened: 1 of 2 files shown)


docs/Getting Started.md
```diff
@@ -44,37 +44,71 @@ Next, make sure everything is set up correctly with a quick test run.
...
+|- ctrl.txt
+|- grid-000001.dat
+|- grid-000002.dat
+|- stderr.txt
+|- stdout.txt
 ```
 
+In case you want to analyze the results of the simulation, you can use the
+Paraview software.
+
```
maiyahershey, line 111: While using the Inductiva API, keep in mind that it does not provide built-in 
visualization tools for the simulators. For such complex 3D visualizations, our team recommends open-source tools 
like [ParaView](https://www.paraview.org/download/) or [VTK](https://vtk.org/download/).

Something like this?

IvanPombo: Made your words mine :D

docs/Getting Started.md
```diff
@@ -44,37 +44,71 @@ Next, make sure everything is set up correctly with a quick test run.
...
+In this example, you will use the open-source hydrodynamics REEF3D simulator to simulate a [dam break scenario](https://github.com/REEF3D/REEF3D/tree/master/Tutorials/REEF3D_CFD/9_1%202D%20Dam%20Break).
 
-Running this simulation requires substantial computational resources and precise 
-configuration, challenges that have previously slowed down your project.
+With this first example, you learn that running simulations via Inductiva API is
+not much different from running them in your local machine. That's the magic
+of it all! 
 
-With Inductiva, you have the opportunity to run the dam break simulation with 
-your existing configuration files.
```
maiyahershey: In this first example, you'll discover the simplicity of running simulations with the Inductiva API – it allows you to run the dam break simulation with your existing configuration files just like you would on your local machine, except it's on the Cloud. 

what do you think? (merging lines 51 till 55)

IvanPombo: I think this was deprecated!

docs/Getting Started.md
sarmento: I would avoid the negative tone. Why would you have to say that you don't have something you are actually not supposed to have :)
Why not simply say that:

Now, the all the data produced by the simulator has been transfered to your local machine where you can post-process it and, in some cases, even visualize it. For example, you can visualize the contents of file X and Y using standard visualization tools suchs as A and B.

sarmento (approved):
Just left a comment (regarding no need to ask for forgiveness for something we didn't do), but I think it's good to go.

PR #1022, opened 2024-01-31
76 remove user confirmation from terminate all if there are no resources

Pbarbosa92:


Pbarbosa92:
Will re-do from a develop branch to avoid conflicts.

PR #1023, opened 2024-01-31
Removed user confirmation when we have no resources.

Pbarbosa92:
We were checking if active_machines is none. But it's never None. machine_groups.get always return a list.

### DIFF
--- a/inductiva/_cli/cmd_resources/terminate.py
+++ b/inductiva/_cli/cmd_resources/terminate.py
@@ -9,7 +9,6 @@ def terminate_machine_group(args):
     counter = 0
 
     if len(active_machines) == 0:
-        print("No active computational resources found.")
         return
 
     # Doesn't run in case --all is not passed.
--- a/inductiva/_cli/cmd_resources/terminate.py
+++ b/inductiva/_cli/cmd_resources/terminate.py
@@ -8,7 +8,7 @@ def terminate_machine_group(args):
     active_machines = resources.machine_groups.get()
     counter = 0
 
-    if len(active_machines) == 0:
+    if not active_machines:
         return
 
     # Doesn't run in case --all is not passed.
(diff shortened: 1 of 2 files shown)


AugustoPeres (approved):
LGTM. Can you add a screenshot showing the behaviour fix?

Pbarbosa92:
> LGTM. Can you add a screenshot showing the behaviour fix?

Yes!
Captura de ecrã 2024-01-31, às 16 20 14

Well, now I see I need to change something.

Pbarbosa92:
Done

IvanPombo (approved):
LGTM!

inductiva/_cli/cmd_resources/terminate.py
```diff
@@ -8,8 +8,7 @@ def terminate_machine_group(args):
     active_machines = resources.machine_groups.get()
     counter = 0
 
-    if active_machines is None:
-        print("No active computational resources found.")
+    if len(active_machines) == 0:
```
sergiomsantos, line 11:
```suggestion
    if not active_machines:
```

inductiva/resources/machine_groups.py
```diff
@@ -84,7 +84,7 @@ def _fetch_machine_groups_from_api():
         api = compute_api.ComputeApi(inductiva.api.get_client())
         response = api.list_active_user_instance_groups()
         if len(response.body) == 0:
-            print("No active machine groups found.")
+            print("No active computational resources found.")
```
sergiomsantos, line 87: we should remove these print statements in API calls and move them to the consumers.

sergiomsantos (approved):

inductiva/resources/machine_groups.py
Pbarbosa92: Create an issue for that. [here](https://github.com/inductiva/tasks/issues/89)

PR #1025, opened 2024-01-31
80 cli remove spot spot from commands

Pbarbosa92:
No more --spot spot

### DIFF
--- a/inductiva/_cli/cmd_resources/cost.py
+++ b/inductiva/_cli/cmd_resources/cost.py
@@ -17,8 +17,8 @@ def estimate_machine_cost(args):
     if num_machines == 1:
         print(f"Estimated cost of machine: {cost} $/h.")
     else:
-        print(f"Estimated total cost (per machine)"
-              ": {cost*num_machines} ({cost}) $/h.")
+        print("Estimated total cost (per machine)"
+              f": {cost*num_machines} ({cost}) $/h.")
 
 
 def register(parser):
--- a/inductiva/_cli/cmd_resources/cost.py
+++ b/inductiva/_cli/cmd_resources/cost.py
@@ -7,12 +7,19 @@ def estimate_machine_cost(args):
     """Estimate the cost of a certain machine type."""
     machine_type = args.machine_type
     spot = args.spot
+    num_machines = args.num_machines
 
     cost = resources.estimate_machine_cost(
         machine_type=machine_type,
         spot=spot,
     )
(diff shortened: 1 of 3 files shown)


IvanPombo (approved):
Confirm the behavior and let me know before merging.

Pbarbosa92:
Working as expected!
Captura de ecrã 2024-02-01, às 09 11 36

inductiva/_cli/cmd_resources/cost.py
IvanPombo, line 1: As you are it here: Can you round here the cost result to three decimal places and moreover add the `-n` parameter to compute the cost for `n` number of machines.

IvanPombo (approved):
While we are at it, please address the decimal places for the cost and add a `n` parameter for the number of machines. See my comment below.

Else...

# QA Approved

inductiva/resources/machine_groups.py
```diff
@@ -31,7 +31,7 @@ def estimate_machine_cost(machine_type: str, spot: bool = False):
     else:
         estimated_cost = instance_price.body["on_demand_price"]
 
-    return round(float(estimated_cost), 5)
+    return round(float(estimated_cost), 3)
```
sergiomsantos, line 34: this is not the right place to round the values. They should be returned as is and only round when printing.

inductiva/_cli/cmd_resources/cost.py
```diff
@@ -7,12 +7,18 @@ def estimate_machine_cost(args):
...
+    num_machines = args.num_machines
 
     cost = resources.estimate_machine_cost(
         machine_type=machine_type,
         spot=spot,
     )
-    print(f"Estimated cost of machine: {cost} $/h.")
+
+    if num_machines == 1:
+        print(f"Estimated cost of machine: {cost} $/h.")
```
sergiomsantos, line 18:
```suggestion
        print(f"Estimated cost of machine: {cost:.3f} $/h.")
```

inductiva/_cli/cmd_resources/cost.py
```diff
@@ -7,12 +7,18 @@ def estimate_machine_cost(args):
...
+    else:
+        print("Estimated total cost (per machine)"
+              f": {cost*num_machines} ({cost}) $/h.")
```
sergiomsantos, line 21:
```suggestion
              f": {cost*num_machines} ({cost:.3f}) $/h.")
```

sergiomsantos (changes requested):

PR #1028, opened 2024-02-01
83 check for api key in  Inductiva init

Pbarbosa92:
I'm not 100% happy with this one. I wanted to validate the key in the init file, but I couldn't do that due to the recursive imports of inductiva. 

Só I had to just check if we had an API key. I've repeated the text from validate_api_key [here](https://github.com/inductiva/inductiva/pull/1018).

When we merge both branches, we need to move this text to the single text file.


IvanPombo:
Can you elaborate on the recursive imports of the Inductiva?

inductiva/__init__.py
```diff
@@ -5,6 +5,8 @@
 
 import absl
 
+from inductiva.api.methods import validate_api_key
+
 from . import api
```
sergiomsantos, line 10: you already import the api module 2 lines below.
You can use it.

```suggestion
from . import api
```

inductiva/__init__.py
```diff
@@ -65,4 +68,13 @@ def _check_for_available_package_update():
         print(msg, file=sys.stderr)
 
 
+def _check_key():
+    global _checked_key
+    if not _checked_key and os.getenv("GITHUB_ACTIONS",
+                                      "false").lower() != "true":
+        validate_api_key(api_key)
```
sergiomsantos, line 75:
```suggestion
        api.methods.validate_api_key(api_key)
```

inductiva/__init__.py
```diff
@@ -65,4 +68,13 @@ def _check_for_available_package_update():
         print(msg, file=sys.stderr)
 
 
+def _check_key():
+    global _checked_key
+    if not _checked_key and os.getenv("GITHUB_ACTIONS",
```
IvanPombo, line 73: Use `utils.format_utils.getenv_bool` to avoid a comparison with a string that is meant to be a boolean.

IvanPombo:
Minor comment on the env variable.

Pbarbosa92:
> Can you elaborate on the recursive imports of the Inductiva?

I got an error that hinted about recursive imports. I can't replicate it now. Maybe it was a problem with my imports.

sergiomsantos (approved):

inductiva/__init__.py
```diff
@@ -68,9 +68,10 @@ def _check_for_available_package_update():
 
 def _check_key():
     global _checked_key
-    if not _checked_key and os.getenv("GITHUB_ACTIONS",
-                                      "false").lower() != "true":
-        validate_api_key(api_key)
+
+    if not _checked_key and utils.format_utils.getenv_bool(
+            "GITHUB_ACTIONS", False) is not True:
```
IvanPombo, line 73: So it is False?

IvanPombo:
Moreover, this needs QA. I will look into it now.

inductiva/__init__.py
Pbarbosa92: If running on Github we will get True, and we won't enter the if statement. If we run on our pc we will get False and we will enter the if statement. 
Github action was failing because we had no api key, so we had to disable this check for GitHub actions.

Pbarbosa92:
> Moreover, this needs QA. I will look into it now.

My bad, I forgot to ask for QA and merged. You'll have to QA on dev and let me know if there is any issue. Thanks

inductiva/__init__.py
IvanPombo: I just wanted to say the `getenv_bool` gets you a boolean, so you don't need to compare with `is not True`, you just need to negate it. My logic classes are also tingling me to use morgan's law and apply a general not like:
`not (_checked_key or utils.format_utils.getenv_bool(...)`

IvanPombo:
In terms of QA, both aspects are working:

- Not setting an API Key, gives error on import and in CLI.
- Setting the API Key, changes the status of _checked_key.
- Setting the GITHUB_ACTIONS env var to True, doesn't perform a verification.

To finish the related issue, it is necessary to go over the code and remove the other api key verifications.

Pbarbosa92:
> In terms of QA, both aspects are working:
> 
> * Not setting an API Key, gives error on import and in CLI.
> * Setting the API Key, changes the status of _checked_key.
> * Setting the GITHUB_ACTIONS env var to True, doesn't perform a verification.
> 
> To finish the related issue, it is necessary to go over the code and remove the other api key verifications.

I'll create a task for that.

PR #1035, opened 2024-02-01
85 added multiple paths to cli remove - also added --all

Pbarbosa92:
This PR makes some cli commands more consistent.

inductiva tasks kill
inductiva storage rm
inductiva resources terminate

Added the ability to kill(rm/terminate) multiple instances by id. Added the ability to kill(rm/terminate) --all instances. 

Made all text across the 3 commands identical.


inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,35 @@
...
+                           " path(s)? (y/[N])")
         confirm = prompt.lower() in ["y", "ye", "yes"]
 
     if confirm:
-        print("Removing %s in the remote storage", path)
-        storage.rmdir(path, confirm=confirm)
+        if all_paths:
+            storage.rmdir("*", confirm=confirm)
+        for path in paths:
+            storage.rmdir(path, confirm=confirm)
```
sergiomsantos, line 32: Just to be extra explicit:

```suggestion
            storage.rmdir("*", confirm=True)
        for path in paths:
            storage.rmdir(path, confirm=True)
```

inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,35 @@
...
+    if paths and all_paths:
+        print("inductiva storage remove: error: "
+              "argument path not allowed with argument --all")
+        sys.exit(1)
 
     if not confirm:
-        prompt = input(f"Are you sure you want to remove {path}? (y/n)")
+        if all_paths:
+            prompt = input("Are you sure you want to remove everything "
+                           "from your storage(y/[N])")
```
sergiomsantos, line 22: Missing question mark and space after last character:

```suggestion
            prompt = input("Are you sure you want to remove everything "
                           "from your remote storage space (y/[N])? ")
```

inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,35 @@
...
+        else:
+            prompt = input(f"Are you sure you want to remove {len(paths)}"
+                           " path(s)? (y/[N])")
```
sergiomsantos, line 25: not in the same format as the previous one:

```suggestion
            prompt = input(f"Are you sure you want to remove {len(paths)}"
                           " path(s) from your remote storage space (y/[N])? ")
```

inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,35 @@
...
     """Remove user's remote storage contents."""
-    path = args.path
+    paths = args.path
     confirm = args.confirm
+    all_paths = args.all
+
+    if paths and all_paths:
+        print("inductiva storage remove: error: "
+              "argument path not allowed with argument --all")
+        sys.exit(1)
```
sergiomsantos, line 17: errors should go to stderr; also, 2 `:` in the same sentence:

```suggestion
    if paths and all_paths:
        print("inductiva storage remove error:\n"
              "argument path not allowed with argument --all", file=sys.stderr)
        sys.exit(1)
```

inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,35 @@
...
 def remove(args):
     """Remove user's remote storage contents."""
-    path = args.path
+    paths = args.path
     confirm = args.confirm
+    all_paths = args.all
+
+    if paths and all_paths:
+        print("inductiva storage remove: error: "
+              "argument path not allowed with argument --all")
```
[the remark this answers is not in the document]
Pbarbosa92: I copied the behaviors of argparser. 
Captura de ecrã 2024-02-01, às 17 35 15

inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,35 @@
...
+        sys.exit(1)
 
     if not confirm:
-        prompt = input(f"Are you sure you want to remove {path}? (y/n)")
+        if all_paths:
+            prompt = input("Are you sure you want to remove everything "
+                           "from your storage(y/[N])")
+        else:
+            prompt = input(f"Are you sure you want to remove {len(paths)}"
```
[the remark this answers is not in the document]
Pbarbosa92: Since our id's have no meaning and are just random numbers and letters I don't think that showing a list of ids like this: n1tuy5pe5wo32ihzz8zar58cn,tgnasiw09a8jprbwwkq2jxiry,ir7rex4rwo4nktv9qjikr0uf3,se4ozwx5uqpq5i3gymp1jkcoa 
would be very helpful. 

Furthermore, we are using the same method for the tasks kill
Do you confirm you want to kill {len(args.id)} tasks (y/[N])?

I can change it on both, but I think a list of id's won't help the users.

inductiva/_cli/cmd_storage/remove.py
Pbarbosa92: I copied the same text from the argparser error when we do
inductiva tasks list -n 5 -id sadase
We need to make these kinds of messages consistent. Should I change it on the list too?

inductiva/_cli/cmd_storage/remove.py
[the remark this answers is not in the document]
IvanPombo: That is because we have not yet started organising our simulation into named folders. But when running a simulation you can pass a `storage_dir="happily_ever_after"` and the simulations will be saved in that folder in the bucket. 

I haven't seen the `kill` method. But I really think we could do it on both. Not in a single line though. Something like:
```bash
"Path to be removed: 
> n1tuy5pe5wo32ihzz8zar58cn
> tgnasiw09a8jprbwwkq2jxiry
> ir7rex4rwo4nktv9qjikr0uf3
> se4ozwx5uqpq5i3gymp1jkcoa
Are you sure you want to remove the above  paths from your storage? (y/[N])"
```

inductiva/_cli/cmd_storage/remove.py
IvanPombo: Yes, apply it in every place you consider necessary :D

inductiva/_cli/cmd_storage/remove.py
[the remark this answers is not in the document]
Pbarbosa92: Changing this here would mean we have to change kill too (kill was just merged). @sergiomsantos do I have the green light to change it?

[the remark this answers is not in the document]
Pbarbosa92: Changed it

inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,45 @@
...
+            "inductiva storage remove: error: "
+            "argument path not allowed with argument --all",
+            file=sys.stderr)
+        sys.exit(1)
 
     if not confirm:
-        prompt = input(f"Are you sure you want to remove {path}? (y/n)")
+        if all_paths:
+            prompt = input("Are you sure you want to remove everything "
+                           "from your remote storage space (y/[N])? ")
```
IvanPombo, line 24: This is just being picky at the moment, but can this message be similar to the ones below and the `everything` be in CAPS LOCK for awarenes..

inductiva/_cli/cmd_tasks/kill.py
```diff
@@ -1,13 +1,21 @@
 """Kills a tasks by id via CLI."""
 
 import inductiva
+from inductiva import constants
 
 
 def kill_task(args):
```
IvanPombo, line 7: The following can be addressed in another PR, but should we permit killing also all running tasks? Wdyt?

IvanPombo (approved):
LGTM, just a minor picky comment and a general open question.

inductiva/_cli/cmd_tasks/kill.py
sergiomsantos: maybe at some point.
Open a task for this.

This is why that document is important. It should be kept up-to-date so that we can check for consistency

IvanPombo (approved):
Thanks for the effort to clean everything @Pbarbosa92! LGTQA!

AugustoPeres:
> **QA**
> 
> There are a few things to test.
> 
> * Let's start with the kill command. We need to check if we have the proper warnings when we are about to kill multiple tasks.
> * to kill <= 5 tasks it should list all the task id's
> * to kill > 5 tasks it should ask if you want to kill X number of tasks
> * Test the same scenario for cli resources remove (they should result in similar messages).
> * Check if there is no duplicated text when removing a folder.
> 
> @IvanPombo and @AugustoPeres

QA

I observe the desired behaviour for tasks and for storage:

Screenshot 2024-02-02 at 10 55 37
Screenshot 2024-02-02 at 10 56 19
Screenshot 2024-02-02 at 10 56 40

But not for machine groups. Is this the desired behaviour:
Screenshot 2024-02-02 at 10 57 33

inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,46 @@
...
+            "argument path not allowed with argument --all",
+            file=sys.stderr)
+        sys.exit(1)
 
     if not confirm:
-        prompt = input(f"Are you sure you want to remove {path}? (y/n)")
+        if all_paths:
+            prompt = input("You are about to remove EVERYTHING from your"
+                           " remote storage space.\n"
+                           "Are you sure you want to proceed (y/[N])? ")
```
sergiomsantos, line 25:
```suggestion
            print("You are about to remove EVERYTHING from your remote storage space.")
```

inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,46 @@
...
+        else:
+            if len(paths) > constants.MAX_CONFIRMATION_LINES:
+                prompt = input(f"You are about to remove {len(paths)} "
+                               "paths from your remote storage space.\n"
+                               "Are you sure you want to proceed (y/[N])? ")
```
sergiomsantos, line 30:
```suggestion
                print(f"You are about to remove {len(paths)} paths from your remote storage space.")
```

inductiva/_cli/cmd_storage/remove.py
```diff
@@ -1,20 +1,46 @@
...
+            else:
+                print("You are about to remove the following paths from "
+                      "your remote storage space: ")
+                for path in paths:
+                    print(f"  - {path}")
+                prompt = input("Are you sure you want to proceed (y/[N])? ")
```
sergiomsantos, line 36: Just a single prompt:

```suggestion
            prompt = input("Are you sure you want to proceed (y/[N])? ")
```

inductiva/utils/input_functions.py
```diff
@@ -0,0 +1,44 @@
...
+        list_costum_message (str): Custom message when the list of things
+            is small and we are about to list them all.
+            Example: with list_of_things with 3 elements and
+            list_costum_message="kill the following tasks" we will print:
+            "You are about to kill the following tasks:"
+    Returns:
+        bool: Whether the user has confirmed the action.
+    """
+    if all_thins:
+        print(f"You are about to {all_custom_message}.")
```
sergiomsantos: I suggest the caller gives the entire message and not just a part.
It will be easier to have messages translated. It will be impossible to translate messages as is.

inductiva/_cli/cmd_resources/terminate.py
```diff
@@ -1,54 +1,69 @@
...
+        confirm = input_functions.user_confirmation_prompt(
+            all_names, names, __("user-prompt-terminate-all"),
+            __("user-prompt-terminate-big", len(names)),
+            __("user-prompt-terminate-small"))
+    if confirm:
+        for machine in active_machines:
+            if all_names:
+                machine.terminate()
+            elif machine.name in names:
+                machine.terminate()
```
sergiomsantos, line 45:
```suggestion
    # dict to map from name to machine
    name_to_machine = {machine.name:machine for machine in active_machines}
    
    active_machine_names = name_to_machine.keys()
    target_machine_names = set(names) # the user can give the same name multiple times!!
    invalid_names = target_machine_names.difference(active_machine_names)

    if invalid_names:
        for name in invalid_names:
            print(f"Resource {name} does not exist.")
        print('aborting)
        return 1

    confirm = confirm or input_functions.user_confirmation_prompt(...)
    if not confirm:
        return 1
    
    machines_to_kill = active_machine_names is all else target_machine_names
    for name in machines_to_kill:
        machine = name_to_machine[name]
        machine.terminate()
    
    return 0
```

Pbarbosa92:
resources remove should now behave in a similar manner to tasks kill and storage rm.

For future QA
test

- inductiva tasks kill
- inductiva storage rm
- inductiva resources terminate

In those three cases test with a few items, a lot of items (or change the constant to trigger the effect with less items), --all items. Test the -y/--yes options. Check if refusing the confirmation works ok.

inductiva/_cli/cmd_resources/terminate.py
```diff
@@ -1,54 +1,78 @@
...
+    active_machine_names = name_to_machine.keys()
+    target_machine_names = set(
+        names)  # the user can give the same name multiple times!!
+    invalid_names = target_machine_names.difference(active_machine_names)
+
+    if invalid_names:
+        for name in invalid_names:
+            print(f"Resource {name} does not exist.")
+        print("Aborting.")
+        return 1
```
IvanPombo, line 38: TBH, I'm not sure what should be the default behaviour, but it seems weird that we would abort everything because one name failed.

IvanPombo: Moreover, thinking about this a bit more, a `invalid_name` may come from an error on the name, or the machines being already terminated. At the moment, we cannot know if it has been terminated, because that is not saved in the DB, but I see a possibility where the same configuration (including the name) are re-usable by the user, so the information of it being already terminated is important if they want to raise again.

Just a bit of rambling... WDYT?

IvanPombo (commented):
This has been a great push @Pbarbosa92, however, as we change stuff around more questions can be raised :D

One thing that catches my eye immediately is the fact that on the termination of resources we are confirming that machines actually exist to throw messages otherwise. But in both the tasks and storage, we don't seem to be doing that.. Would you care to comment on your decision?

Pbarbosa92:
> This has been a great push @Pbarbosa92, however, as we change stuff around more questions can be raised :D
> 
> One thing that catches my eye immediately is the fact that on the termination of resources we are confirming that machines actually exist to throw messages otherwise. But in both the tasks and storage, we don't seem to be doing that.. Would you care to comment on your decision?

Well, tasks and folders can be A LOT, like in the hundreds of thousands. Machines should not be that high. For us to check if a task exists, we would have to do it one of two ways:

- Get all tasks and check if my IDs are inside the list of all tasks. All tasks can be huge and requires multiple calls to the API to get.
- Instantiate a task for every if to kill do the get status and based on that result, we can know if the task exists. That takes one API call per task.

When checking for resources, we just do one api call to get all resources (which should be small) and check if our IDs are inside that list of all resources.

Storage suffers from a similar problem, the output can be huge if we just do a list and check if our paths are there.

inductiva/localization/translations/en.json
```diff
@@ -11,5 +11,44 @@
...
+    ],
+    "user-prompt-terminate-small":
+    [
+        "You are about to terminate the following resources:"
+    ],
+    "user-prompt-remove-small":
+    [
+        "You are about to remove the following paths from your remote storage space:"
+    ],
+    "user-prompt-kill-small":
```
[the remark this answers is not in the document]
Pbarbosa92: I agree. Changed it.

inductiva/_cli/main.py
```diff
@@ -46,6 +46,8 @@ def main():
 
     # Call the function associated with the subcommand
     try:
-        args.func(args)
+        status = args.func(args)
     except Exception as e:  # pylint: disable=broad-except
         print(e)
+        status = 1
+    return status
```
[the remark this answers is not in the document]
Pbarbosa92: Maybe @sergiomsantos can shine a light on this topic.

IvanPombo:
> > This has been a great push @Pbarbosa92, however, as we change stuff around more questions can be raised :D
> > One thing that catches my eye immediately is the fact that on the termination of resources we are confirming that machines actually exist to throw messages otherwise. But in both the tasks and storage, we don't seem to be doing that.. Would you care to comment on your decision?
> 
> Well, tasks and folders can be A LOT, like in the hundreds of thousands. Machines should not be that high. For us to check if a task exists, we would have to do it one of two ways:
> 
> * Get all tasks and check if my IDs are inside the list of all tasks. All tasks can be huge and requires multiple calls to the API to get.
> * Instantiate a task for every if to kill do the get status and based on that result, we can know if the task exists. That takes one API call per task.
> 
> When checking for resources, we just do one api call to get all resources (which should be small) and check if our IDs are inside that list of all resources.
> 
> Storage suffers from a similar problem, the output can be huge if we just do a list and check if our paths are there.

Notice that your second suggestion is what you will be doing anyway. I just want to make sure that we pass the information of error or non-existence correctly to the user. That is all that matters. You don't need to check for every path in existence, just to throw an appropriate logging error mentioning that to the user. First, start by looking into what happens when you either send one single task that doesn't exist (or path for the storage).

inductiva/_cli/cmd_resources/terminate.py
Pbarbosa92: Well, I'm a fan of terminating everything if one fails. As a user, I would want my command to be fully complete or none at all. I don't want my environment in a gray area where I don't really know what was done or wasn't (I mean, I can know if I read the full logs, but that can be massive).

IvanPombo: That is fine by me, but then we should not do a loop to add more logs to the table... It should just be a single message on top and then list the invalid names below.

Pbarbosa92:
> > > This has been a great push @Pbarbosa92, however, as we change stuff around more questions can be raised :D
> > > One thing that catches my eye immediately is the fact that on the termination of resources we are confirming that machines actually exist to throw messages otherwise. But in both the tasks and storage, we don't seem to be doing that.. Would you care to comment on your decision?
> > 
> > 
> > Well, tasks and folders can be A LOT, like in the hundreds of thousands. Machines should not be that high. For us to check if a task exists, we would have to do it one of two ways:
> > 
> > * Get all tasks and check if my IDs are inside the list of all tasks. All tasks can be huge and requires multiple calls to the API to get.
> > * Instantiate a task for every if to kill do the get status and based on that result, we can know if the task exists. That takes one API call per task.
> > 
> > When checking for resources, we just do one api call to get all resources (which should be small) and check if our IDs are inside that list of all resources.
> > Storage suffers from a similar problem, the output can be huge if we just do a list and check if our paths are there.
> 
> Notice that your second suggestion is what you will be doing anyway. I just want to make sure that we pass the information of error or non-existence correctly to the user. That is all that matters. You don't need to check for every path in existence, just to throw an appropriate logging error mentioning that to the user. First, start by looking into what happens when you either send one single task that doesn't exist (or path for the storage).

I've added some exception-catching for tasks and storage. Now, if something fails, we will continue with the others.
@sergiomsantos What is the desired behavior? All or nothing? Or just do what we can and log for failures?

inductiva/_cli/cmd_tasks/kill.py
```diff
@@ -1,20 +1,49 @@
...
+        all_ids = []
+        for status in constants.TASK_RUNNING_STATUSES:
+            all_ids.extend(get_all(status=status))
+        ids = all_ids
 
-    for task_id in args.id:
-        inductiva.tasks.Task(task_id).kill(wait_timeout=args.wait_timeout)
+    confirm = args.yes or user_confirmation_prompt(
+        ids, __("task-prompt-kill-all"), __("task-prompt-kill-big", len(ids)),
+        __("task-prompt-kill-small"), kill_all)
```
sergiomsantos, line 37: please consider readability.

Example:

```suggestion
    confirm = args.yes or \
        user_confirmation_prompt(ids,
                                 __("task-prompt-kill-all"),
                                 __("task-prompt-kill-big", len(ids)),
                                 __("task-prompt-kill-small"),
                                 kill_all)
    if not confirm:
        return 1
```

inductiva/_cli/cmd_tasks/kill.py
```diff
@@ -1,20 +1,49 @@
 """Kills a tasks by id via CLI."""
+import sys
 
 import inductiva
+from inductiva import constants
+from inductiva.tasks.methods import get_all
+from inductiva.utils.input_functions import user_confirmation_prompt
+from ...localization import translator as __
+
+#             "pending-input": "PENDINGINPUT",
```
AugustoPeres, line 10: Remove this comment

inductiva/tasks/methods.py
```diff
@@ -143,3 +143,31 @@ def get(
...
+            will be returned.
+    Returns:
+        List of Task objects.
+    """
+    status = models.TaskStatusCode(status) if status is not None else None
+
+    all_tasks = []
+    page_counter = 1
+
+    while tasks_fetched := _fetch_tasks_from_api(status,
```
AugustoPeres, line 167: The `:=` is only supported from python3.8 forward. Is this in agreement with the versions we claim to support?

AugustoPeres (approved):

inductiva/tasks/methods.py
Pbarbosa92: Yes. I think we support from 3.9 forward.

inductiva/localization/translations/en.json
```diff
@@ -8,8 +8,52 @@
...
         "The following error occurred:\n  {0}"
     ],
-    "kill-request-sent":["Successfully sent kill request for task {0}.",
+    "task-request-sent":["Successfully sent kill request for task {0}.",
         "> Consider tracking the task status by calling task.get_status()",
         "> or via the CLI using 'inductiva tasks -id {0}'."
+    ],
+    "user-prompt-confirmation": [
+        "Are you sure you want to proceed (y/[N])? "
+    ],
```
sergiomsantos, line 17: single lined messages can be inlined (no need to be wrapped in an array):

```suggestion
    "user-prompt-confirmation": "Are you sure you want to proceed (y/[N])? ",
```

inductiva/tasks/methods.py
sergiomsantos: yep 3.8+

inductiva/tasks/task.py
```diff
@@ -35,13 +35,50 @@ class Task:
...
+        models.TaskStatusCode.SPOTINSTANCEPREEMPTED,
+        models.TaskStatusCode.ZOMBIE
+    }
+
+    TERMINAL_STATUSES = {models.TaskStatusCode.SUCCESS}.union(FAILED_STATUSES)
+
+    RUNNING_STATUSES = {
+        models.TaskStatusCode.PENDINGINPUT, models.TaskStatusCode.SUBMITTED,
+        models.TaskStatusCode.STARTED
+    }
```
sergiomsantos, line 52: SUBMITTED is not running

inductiva/tasks/task.py
```diff
@@ -35,13 +35,50 @@ class Task:
...
         self._api = tasks_api.TasksApi(api.get_client())
         self._info = None
         self._status = None
 
+    def is_running(self) -> bool:
+        """Validate if the task is running.
+
+        This method issues a request to the API.
+        """
+        return self.get_status() in self.RUNNING_STATUSES
```
sergiomsantos, line 66: this is not the same as the previous implementation

inductiva/localization/translations/en.json
```diff
@@ -8,8 +8,52 @@
         "Unable to check for available updates for the 'inductiva' package.",
         "The following error occurred:\n  {0}"
     ],
-    "kill-request-sent":["Successfully sent kill request for task {0}.",
+    "task-request-sent":["Successfully sent kill request for task {0}.",
```
sergiomsantos, line 11: `task-request-sent` (and `kill-request-sent`) do not give provide enough context on what the msg is used for. Perhaps (make sure it is used in the right place because I didn't see an update for it's usage):

```suggestion
    "task-kill-request-sent":["Successfully sent kill request for task {0}.",
```

inductiva/tasks/task.py
Pbarbosa92: I created a KILLABLE_STATUSES.

inductiva/tasks/task.py
Pbarbosa92: I reverted to the previous implementation.

inductiva/localization/translations/en.json
Pbarbosa92: I've changed the name of the variable and forgot to change the place where it was called. Fixed now.

sergiomsantos (approved):

AugustoPeres:
### QA

#### Terminate 1 machine group
Screenshot 2024-02-05 at 17 19 29

#### Terminate +5 machine groups
Screenshot 2024-02-05 at 17 24 58

#### Rm <5 storage folders
Screenshot 2024-02-05 at 17 28 18

#### RM >5 storage folders

Screenshot 2024-02-05 at 17 30 03

#### Kill >5 tasks
Screenshot 2024-02-05 at 17 31 18

#### Kill <5 tasks
Screenshot 2024-02-05 at 17 32 29

Everything seems to be working as expected

PR #1044, opened 2024-02-02
Add hooks for missing sections

sergiomsantos:
This PR adds some hooks for missing sections in the MVP documentation.


docs/index.md
```diff
@@ -13,33 +13,52 @@ Getting Started
...
 maxdepth: 2
-hidden: true
+hidden: false
 ---
-Storage Levels and Data Flow
-Managing Storage with the Inductiva API
+Ivan_Computational_Resources
+Shared VM Pool Integration
+Machine Groups
+Setting up an MPI
```
IvanPombo, line 23: I think the `Ivan_Computational_Resources` should be at the end since it is more technical and in-depth.

IvanPombo (approved):
Looking great! Do we already have a placeholder for the Shared Pool readme?

IvanPombo (changes requested):
It actually seems that the toctree is not working correctly, since it is only appearing the `Computational Resources` and not the tutorials on machines.

docs/index.md
maiyahershey: I think we do have a placeholder for the Shared Pool in the Machines.md 
[Task Queue and Shared VM Pool Integration]

IvanPombo: I am not finding that ReadMe and is not rendering it on the ReadTheDocs. Moreover, the Machine Groups name should change to Machines or the file to `Machine Groups.md`

sarmento:
Let's merge for now and then let's synchronize monday.
I think we are starting to have enough material to then create the appropriate TOC

sarmento (approved):

PR #1059, opened 2024-02-06
Bugfix - tasks list prints 3 empty lines when we have no tasks

Pbarbosa92:
Problem:
When we have no tasks our list will give the wrong output
```
(inductiva-dev) paulobarbosa@Computador-portatil-de-Paulo inductiva % inductiva tasks list -n 10

    

(inductiva-dev) paulobarbosa@Computador-portatil-de-Paulo inductiva % 
```
Solution:
Now to_dict returns a dict with all keys with empty arrays if no tasks are found. If we decide that we want to print an empty table with only column names, we should be able to do that easily now.
inductiva tasks list now prints "no tasks" found if no tasks found

### DIFF
--- a/inductiva/_cli/cmd_tasks/list.py
+++ b/inductiva/_cli/cmd_tasks/list.py
@@ -19,6 +19,10 @@ def list_tasks(args):
         last_n = 5 if args.last_n is None else args.last_n
         task_list = tasks.get(last_n=last_n)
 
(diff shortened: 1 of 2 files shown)


AugustoPeres (approved):
Thanks Paulo. Tested on my side and it looks good:
Screenshot 2024-02-06 at 15 20 46

PR #1060, opened 2024-02-06
We now only allow 20GB for disk size.

Pbarbosa92:
Problem:
If I created a machine group with 30gb disk size we get:
```
Registering MachineGroup configurations:
Registering machine group failed with exception (500)
Reason: Internal Server Error
HTTP response headers: HTTPHeaderDict({'content-type': 'application/json', 'X-Cloud-Trace-Context': '94734d08d9768b552f2402a705f9b602', 'Date': 'Tue, 06 Feb 2024 12:24:58 GMT', 'Server': 'Google Frontend', 'Content-Length': '47'})
HTTP response body: b'{"detail":"Disk size must be less than 80 GB."}'
```
Solution:
We now pass the disk size as is to the back-end. All checks and logic are done on the backend. We just present the possible exception or not.


inductiva/resources/machines_base.py
```diff
@@ -47,14 +47,12 @@ def __init__(self,
         if machine_type not in inductiva.resources.list_available_machines():
             raise ValueError("Machine type not supported")
 
-        if data_disk_gb < 0 or data_disk_gb > 100:
+        if data_disk_gb < 0 or data_disk_gb > 1940:
```
AugustoPeres, line 50: According to the description of the pull request shouldn't we use `data_disk_gb > 20` here?

inductiva/resources/machines_base.py
```diff
@@ -180,7 +177,7 @@ def terminate(self, **kwargs):
                     id=self.id,
                     name=self.name,
                     machine_type=self.machine_type,
-                    disk_size_gb=self._true_disk_size_gb,
+                    disk_size_gb=self.data_disk_gb,
```
AugustoPeres, line 180: Since now we are not sending `_true_disk_size_gb` but `data_disk_gb` and the backend handles the rest would it make sense to also rename the arguments of the functions making the requests accordingly. That is:

```python
inductiva.client.models.GCPVMGroup(
                    id=self.id,
                    name=self.name,
                    machine_type=self.machine_type,
                    disk_size_gb=self._true_disk_size_gb,
                    # Change to data
                    data_size_gb=self.data_disk_gb,)
```

AugustoPeres (commented):
LGTM, Left just a few comments

inductiva/resources/machines_base.py
Pbarbosa92: > According to the description of the pull request shouldn't we use `data_disk_gb > 20` here?

We just need to check if data_disk_gb is positive. I fixed that on the wrong branch and ended up forgetting to change it. Will be fixed on the next commit.

inductiva/resources/machines_base.py
Pbarbosa92: _true_disk_size_gb no longed exists. All the math is handled on the back-end.

IvanPombo (approved):
Change the description of the PR for future reference. Else, good to QA.

Pbarbosa92:
**QA**

Try to launch multiple types of machine groups. Try multiple disk sizes.
We expect:
- disk size <= 0 - Error from the client
- disk size > 0 and <=100 - No problem
- disk size > 100 - error from the backend

sergiomsantos (approved):

PR #1067, opened 2024-02-06
T96  Add `watch` option to selected commands 

sergiomsantos:
inductiva/tasks#96

### Description
This PR adds the ability to periodically run some CLI commands _à la_ `watch` style.
The new functionality is based on a custom pager that implements file-like interface to which content can be printed.
The pager uses ANSI escape codes to provide screen management and navigation.


inductiva/_cli/pager.py
```diff
@@ -0,0 +1,312 @@
...
+    FOOTER = 1
+
+    def __init__(self, name=None):
+        self._lock = threading.RLock()
+        self._init()
+        self._set_header(name)
+        self.flush()
+
+    def _init(self):
+        """Intialize the display."""
```
IvanPombo, line 26: Typo
```suggestion
        """Initialize the display."""
```

inductiva/_cli/pager.py
```diff
@@ -0,0 +1,312 @@
...
+        self._line_count = 0
+        self._height, self._width = self._window.getmaxyx()
+
+        curses.noecho()
+        curses.cbreak()
+        # hide the cursor
+        curses.curs_set(0)
+
+        # receive keypad events and make 1-millisecond
+        # timeouts on `getch`
```
IvanPombo, line 38: Is this fetch?

inductiva/_cli/pager.py
```diff
@@ -0,0 +1,312 @@
...
+    HEADER = 1
+    FOOTER = 1
+
+    def __init__(self, name=None):
+        self._lock = threading.RLock()
+        self._init()
+        self._set_header(name)
+        self.flush()
+
+    def _init(self):
```
IvanPombo, line 25: Maybe:
```suggestion
    def _init_display(self):
```

inductiva/_cli/pager.py
```diff
@@ -0,0 +1,312 @@
...
+        Line separators are not added, so it is usual for each of the
+        lines provided to have a line separator at the end.
+        Note: The display is only updated after a call to the `flush` method.
+        """
+        self.write("".join(lines))
+
+    def flush(self) -> None:
+        """Flush the content of the file-like object to the display."""
+        with self._lock:
+            self._header.refresh(0, 0, 0, 0, self.HEADER, self._width)
```
IvanPombo, line 169: Will we always update the header or just the first time? Or given the trouble, it os just easier to just update it always?

inductiva/_cli/pager.py
```diff
@@ -0,0 +1,312 @@
...
+        Always returns False.
+        """
+        return False
+
+    def seekable(self) -> bool:
+        """Return whether the object supports random access.
+
+        Always returns False.
+        """
+        return False
```
IvanPombo, line 221: What is the actual goal of having these functions?

inductiva/_cli/pager.py
```diff
@@ -0,0 +1,312 @@
...
+            "File-like object does not support the truncate method")
+
+    def __next__(self) -> AnyStr:
+        raise TypeError("File-like object is not an iterator")
+
+    def __iter__(self) -> Iterator[AnyStr]:
+        raise TypeError("File-like object is not iterable")
+
+
+class Scheduler(threading.Thread):
```
IvanPombo, line 257: So, this is the Watcher, no?

Scheduler is the common name?

inductiva/_cli/pager.py
```diff
@@ -0,0 +1,312 @@
...
+    every = 5.0  # seconds
+
+    def producer(fout: PagedOutput, schedulr: Scheduler):
+        fout.clear()
+        for i in range(50):
+            fout.write(f"--- line {schedulr.count} {i+1} ---\n")
+        fout.scroll_to(0)
+
+    with PagedOutput(f"> every {every} s: inductiva tasks list -n 10") as f:
+        scheduler = Scheduler(every, producer, args=())
```
IvanPombo, line 303: A bit confused here... The `Scheduler` takes in the producer, which itself takes a Scheduler...

So, how is it actually initialized?

IvanPombo (commented):
Looking good Terminal King! Just left a few comments to better understand things.

IvanPombo:
@sergiomsantos can you please review your own PR? :D

inductiva/_cli/main.py
```diff
@@ -54,9 +55,14 @@ def main():
...
 
+    exit_code = 0
     # Call the function associated with the subcommand
     try:
-        exit_code = args.func(args)
+        if args.watch is None:
+            exit_code = args.func(args)
+        else:
+            method = " ".join(sys.argv[3:])
+            _cli.utils.watch(args, method)
```
sergiomsantos, line 65:
```suggestion
        if getattr(args, "watchable", False) and args.watch is not None and :
            cmd = " ".join(sys.argv[1:])
            watch(args.func, args.watch, args, cmd)
        else:
            exit_code = args.func(args)
```

sergiomsantos (commented):
Just some minor comments to make the code simpler and more resilient.
I think we need to investigate why the `stty sane` is needed. I don't need it on my side, so it just might be symptom of some problem with your system.

inductiva/_cli/ansi_pager.py
```diff
@@ -0,0 +1,464 @@
...
+
+    def close(self) -> None:
+        """Close the file-like object.
+
+        The method closes the file-like object and resets the terminal
+        to its previous state."""
+        self._resizer.stop()
+        self._resizer.join()
+        self._exit_alt_screen()
+        os.system("stty sane")  # Required to return terminal to normal.
```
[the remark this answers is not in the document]
IvanPombo: Which terminal are you using? This error appears to me in bash, I have tried in zsh and I don't have the problem. It seems that the terminal needs to be reset. For example, after commenting this and executing one prompt that creates a pager, I get the error. If I type reset (even though invisibly) it goes back to how it was.

In fact, if it `os.system("reset")` it refactors back again, but it takes longer and clears the terminal fully.

inductiva/_cli/utils.py
```diff
@@ -11,6 +12,35 @@
...
+    subparser.set_defaults(watchable=True)
+    subparser.add_argument("-w",
+                           "--watch",
+                           nargs="?",
+                           const=2.0,
+                           type=positive_float,
+                           help="Prompt the command every N seconds.")
+
+
+def remove_flags(string, flags):
```
sergiomsantos, line 36: this function has a bug and will not work all the time:

```python
>>> remove_flags("inductiva hi -w -t 10", ["w", "watch"])
'inductiva hi 10'

>>> remove_flags("inductiva hi -w -t", ["w", "watch"])
'inductiva hi'
```

inductiva/_cli/main.py
```diff
@@ -42,6 +44,25 @@ def get_main_parser():
...
+def watch(func, every, args, cmd):
+    """Run the function at regular intervals and display results in a pager."""
+    cmd = _cli.utils.remove_flags(cmd, ["watch", "w"])
+    header = f"> every {every}s: inductiva {cmd}"
+
+    def action(fout: TextIO = sys.stdout):
+        fout.clear()
+        func(args, fout=fout)
+
+    with ansi_pager.PagedOutput(header, "Press 'q' to close.") as pager:
```
sergiomsantos, line 56: There is more to the paginator than simply pressing `q` to exit; you can scroll line by line and page by page.
Hence the default footer.

inductiva/_cli/ansi_pager.py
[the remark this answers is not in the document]
sergiomsantos: I've tried it with zsh and bash in iterm2, terminal.app and vscode's terminal.

inductiva/_cli/cmd_tasks/list.py
```diff
@@ -14,40 +17,40 @@ def list_tasks(args):
...
     header_formatters = [
-        lambda x: emph_formatter(x.upper(), utils.format_utils.Emphasis.BOLD)
+        lambda x: emph_formatter(x.upper(), format_utils.Emphasis.BOLD)
     ]
 
-    print(
-        utils.format_utils.get_tabular_str(table_dict,
-                                           formatters=formatters,
-                                           header_formatters=header_formatters))
+    print(format_utils.get_tabular_str(table_dict,
```
sergiomsantos, line 49: this print is adding a blank line at the end that prevents the paginator from being able to scroll to the last line.
Well, not really, because it is indeed scrolling but the last line is an empty line so it gives the idea that the content of the file went away.

sergiomsantos (commented):
there's a bug in the remove flag function. I think we can get away without it!
unit tests would have caught this.

inductiva/_cli/cmd_resources/list.py
```diff
@@ -81,9 +83,16 @@ def register(parser):
                                   help="List currently active resources.",
                                   formatter_class=argparse.RawTextHelpFormatter)
 
+    subparser.add_argument("-w",
+                           "--watch",
+                           nargs="?",
+                           const=2.0,
+                           type=float,
+                           help="List the tasks every N seconds.")
```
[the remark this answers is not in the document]
sergiomsantos: ok, I see. You want to be able to pass `--watch` without needing to specify the default sleep time.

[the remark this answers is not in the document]
sergiomsantos: I also think 2s is to frequent. 5s would sound better

inductiva/_cli/ansi_pager.py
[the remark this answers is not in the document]
IvanPombo: To be fair, I don't understand the reason to not work on my computer...
But, to be completely sure, I will comment this on QA so that we can verify if it works for other people.

inductiva/_cli/cmd_resources/list.py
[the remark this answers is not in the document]
IvanPombo: It is the default of the watch method, so I wanted to keep it as 2.0, since it is the expected behavior for knowledgable people.

inductiva/_cli/utils.py
IvanPombo: Found a hacky way to fix the bug, now some extra spaces may still appear.

IvanPombo:
@sergiomsantos pinging you again, for a final review.

IvanPombo (approved):
LGTM @sergiomsantos !

PR #1070, opened 2024-02-07
first example edits

maiyahershey:
Fixed gif, edited run your first simulation page to flow better with less interruptions.


sarmento (approved):
This is great! I really like it!
Let's keep going in this direction in the other files!

PR #1071, opened 2024-02-07
FIX-changing-kill-text

Pbarbosa92:
If our task status is `success` we will have the message:
```
Unable to ensure that task r7wq2mwm551zqxdgp62nmfvjq transitioned to the KILLED state after 5 seconds. The status of the task is success
```
immediately. So it does not make sense to have "after 5 seconds".


AugustoPeres (approved):

IvanPombo:
> If our task status is `success` we will have the message:
> 
> ```
> Unable to ensure that task r7wq2mwm551zqxdgp62nmfvjq transitioned to the KILLED state after 5 seconds. The status of the task is success
> ```
> 
> immediately. So it does not make sense to have "after 5 seconds".

Do you mean if the task finished in between and reached any terminal status?

Pbarbosa92:
> > If our task status is `success` we will have the message:
> > ```
> > Unable to ensure that task r7wq2mwm551zqxdgp62nmfvjq transitioned to the KILLED state after 5 seconds. The status of the task is success
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > immediately. So it does not make sense to have "after 5 seconds".
> 
> Do you mean if the task finished in between and reached any terminal status?

Yes, if the task leaves pending kill to any state other than killed we will get that message, even if the time has not passed.

IvanPombo (approved):
LGTM! The only thing is if possible to decrease the amount of lines in the prints.

sergiomsantos (approved):

AugustoPeres:
### QA

When `status = models.TaskStatusCode.PENDINGKILL`:
Screenshot 2024-02-09 at 12 38 13

When `status = models.TaskStatusCode.KILLED`
Screenshot 2024-02-09 at 12 39 12

When `status = models.TaskStatusCode.EXECUTERTERMINATEDBYUSER or ZOMBIE`:
Screenshot 2024-02-09 at 12 39 52
Screenshot 2024-02-09 at 12 40 20

PR #1072, opened 2024-02-07
Edited introduction

maiyahershey:
Edited introduction to the functionalities, applied some edits to the skeleton, and added infographic


docs/index.md
```diff
@@ -14,7 +14,7 @@ first_example
 
 ```{toctree}
 ---
-caption: Overview
+caption: API Functionalities
```
sarmento, line 17: I am not sure if Functionalities is the best word here.
I understand the Overview is maybe too generic, and needs to be changed.
But Functionalities is a lot about what it does, and this is more about how it works.
WDYT?

docs/introduction/index.md
```diff
@@ -1,16 +1,41 @@
...
+- **Shared vs. Dedicated Resources:** Understanding how tasks get executed on 
+computational resources that are either **shared** among all users or **dedicated** 
+to individual users.
+
+- **Storage and Data Flow:** Understanding the flow of data across platforms, from 
+sending your input files to remote computational resources to accessing the 
+simulation outputs.
+
+- **Infrastructure:** A further look under the hood, representing the underlying 
+infrastructure powering your simulations, currently leveraging [Google Cloud Platform](https://cloud.google.com/compute/docs/machine-resource), with more options on the horizon.
```
sarmento, line 41: I would point to our new page about infra-structure, not outbound.
In fact, I think each of the these would point to the content that we have already there.
Or not, because we have the side navigation, but I would remove the outbound link.

docs/index.md
IvanPombo: Agreed here, I also thought functionalities was to specific to the methods and exactly to what it does. 
I think Overview is still fine since these sets the stage for the more nitty gritty details.

maiyahershey: I get you, how about any of these?
- api workflow
- api architecture 
- usage

docs/introduction/index.md
maiyahershey: I think we can open a PR tomorrow just for links, the toc and the titles still need work imo.
I will link these to their respective places now

docs/index.md
sarmento: What do the readthedocs of similar systems usually show?
If we don't find something really strong, I think it's fine if we keep overview for now...

maiyahershey: Usually what I found as best practice:

- capabilities
- User guide/ api usage
- features

Overview is too broad, and confusing because I feel we have so much intros now. In this part of the doc, we are in the specificities and we need to make that explicit for the reader so they don't get frustrated when navigating the overall doc 

I'll send you some examples to the chat

sarmento: OK! I see.
But it's not about capabilities, functionalities (what it does), or features (what it has). 
It could be something like "how it works". It's about allowing users to build  a mental image of what is going on, and why things happen in certain ways. 

"Understanding the API"?

maiyahershey: I love it! I was worried we were passing on using the verbs. this is quite cool and it would flow well with the rest of the stuff. I'll adjust stuff.

PR #1075, opened 2024-02-07
Task overview document

sergiomsantos:


docs/introduction/tasks.md
```diff
@@ -1,148 +1,238 @@
 # Tasks
 
-The **Inductiva API** client provides the ability to run complex simulations and
-contains easy-to-use tools to manage them. Each simulation was submitted - with
-`simulator.run()` - runs asynchronously and returns a `Task` object that contains
-methods to manage the simulation.
+At the heart of the API lies the concept of a `Task`.
+A task is an abstraction that encapsulates all the information about the
+computational workload generated by the user. A task is generated when a
```
sarmento, line 5: "generated" twice

docs/introduction/tasks.md
```diff
@@ -1,148 +1,238 @@
 # Tasks
 
-The **Inductiva API** client provides the ability to run complex simulations and
-contains easy-to-use tools to manage them. Each simulation was submitted - with
-`simulator.run()` - runs asynchronously and returns a `Task` object that contains
-methods to manage the simulation.
+At the heart of the API lies the concept of a `Task`.
```
sarmento, line 3: I think we are using this image ("at the heart") elsewhere. I think it is fine now, but then Maya can help with cleaning up this sort of repeated pattarns.

docs/introduction/tasks.md
```diff
@@ -1,148 +1,238 @@
...
+A task is an abstraction that encapsulates all the information about the
+computational workload generated by the user. A task is generated when a
+simulation is submitted to the API and exists until the user deletes it.
+With a task at hand, the user can manage the simulation, retrieve its status,
```
sarmento, line 7: Not sure if we should address it as *T*ask. We are doing the same for other core concepts, such as "Machine Groups"

docs/introduction/tasks.md
```diff
@@ -1,148 +1,238 @@
...
-# Kill a task that hasn't completed yet.
-if status == "submitted" or status == "started":
-    task.kill()
 ```
 
+A task is uniquely identified by an alphanumeric identifier which is unique across
+all tasks submitted to the API by the user. However, a user can create multiple
+`Task` objects that point to the same task by instantiating the `Task` class with
+the same identifier without this resulting in the creation of new tasks on the
+API; they will be different Python objects, but they will refer to the same task
```
sarmento, line 54: Not sure if you need to address this repetition for now. I would just focus on retrieving the task by id.

The other situation could be addressed in a FAQ such as "What happens is I create two task objects for the same Task id?"

But as it is, it seems we are tryng to ask a question that the users still hasn't asked.

docs/introduction/tasks.md
```diff
@@ -1,148 +1,238 @@
...
+i4ir3kvv62odsfrhko4y8w2an
+>>> print(task2.id)
+i4ir3kvv62odsfrhko4y8w2an  # the same as task1.id
+```
 
-# Block the call until the simulation completes, but
-# without interrupting the remote task if the local process
-# is interrupted. 
-task.wait()
+### Managing a task
```
sarmento, line 73: I think some of these entries make more sense in the "How To" section.
Or maybe the title as it is may not be the most appropriate. We are leaving the "Manage X" to the "How To".

docs/introduction/tasks.md
```diff
@@ -1,148 +1,238 @@
...
+>>> task.get_status()
+<DynamicSchema: 'success'>
+
+# Get the computational resource type where the task is
+# running or was submitted to
+>>> task.get_machine_type()
+'c2-standard-4'
+
+>>> task.get_computation_time()
+'00:03:12
```
sarmento, line 103: missing '

Btw, if this a string maybe we want to rethink. A string is something quite unhelpful because I can only print it. Maybe I want to have a "ytime" object or I want have:

task.get_computation_time_in_secs()

which returns an float.

docs/introduction/tasks.md
```diff
@@ -1,148 +1,238 @@
...
 ```
 
-```python
-# Download all the generated files.
-output_dir = task.download_outputs()
+Most of the methods of the `Task` class in the snippet above accept multiple
+arguments to configure the action to be performed. For example, the `download_outputs`
+method allows the user to specify which files to download and where to save
+the downloaded files.
+Please refer to the [Task API documentation](../api/inductiva.tasks.html) for more
```
sarmento, line 128: Where is this?

docs/introduction/tasks.md
```diff
@@ -1,148 +1,238 @@
...
+information on the methods available in the `Task` class.
+
+#### Downloading outputs
```
sarmento, line 131: I think this is How To

sarmento (commented):
I think this is a great push, but I would probably move part of this content to the How To section, as a "Manage Task" subsection. 

Here I would focus more on the "life of the Task", and maybe in explaining concepts such synchronous and asynchronous tasks.

sarmento (approved):

PR #1076, opened 2024-02-08
115-Give feedback on number on running machines on machine group

Pbarbosa92:
This PR aims to show the current active machines in a machine group.
```
NAME                                MACHINE TYPE         ELASTIC         TYPE           # MACHINES
api-2iwt33f8gqw3d1zyv4ru4vymo       c3-standard-22       False           standard       1/10 <--- requested 10 but only 1 is alive
api-43etg4t6153mn3ky9nnjfvndp       c3-standard-8        False           standard       5/10 <--- requested 10 but only 5 are alive
```


inductiva/resources/machine_cluster.py
```diff
@@ -52,7 +53,8 @@ def __init__(self,
     @classmethod
     def from_api_response(cls, resp: dict):
         machine_group = super().from_api_response(resp)
-        machine_group.num_machines = int(resp["num_vms"])
+        machine_group.num_machines = int(resp["max_vms"])
```
Pbarbosa92, line 56: Previously, num_machines always had the number of machines requested, this means that to keep this variable with the same number, number_machines now has max_vms.
I did this in hopes of not breaking any code that was counting on num_machines having the maximum number of machines.

inductiva/resources/machine_cluster.py
```diff
@@ -52,7 +53,8 @@ def __init__(self,
...
+        machine_group.__dict__["_current_machines"] = int(resp["num_vms"])
```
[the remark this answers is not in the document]
AugustoPeres: I agree that this should be the way to go. For me it seems strange that I can do:

```
import inductiva

mg = inductiva.resources.ElasticMachineGroup(...2, 20)
# submit here 1e10 tasks
mg.current_machines_to_str() # ---> 0/20
```

Because the var is never updated from a backend response. However, if the backend still does not support this then I have no issues with this approach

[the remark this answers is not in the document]
Pbarbosa92: We don't have a get machine group by id. Once they implement that we can implement a method called update_current_machines().

inductiva/resources/machine_groups.py
```diff
@@ -54,12 +54,10 @@ def _machine_group_list_to_str(machine_group_list) -> str:
...
 
         if isinstance(machine_group, resources.ElasticMachineGroup):
-            num_active_machines = machine_group.num_active_machines
-            max_machines = machine_group.max_machines
-            num_active_machines = f"{num_active_machines}/{max_machines}"
+            num_active_machines = machine_group.current_machines_to_str()
             is_elastic = True
         else:
-            num_active_machines = machine_group.num_machines
+            num_active_machines = machine_group.current_machines_to_str()
```
sergiomsantos, line 60: these 2 lines are the same. they can be moved out of the 1f/else body

inductiva/resources/machines.py
```diff
@@ -173,6 +177,12 @@ def from_api_response(cls, resp: dict):
         machine_group.num_active_machines = int(resp["num_vms"])
         return machine_group
 
+    def current_machines_to_str(self) -> str:
+        """Returns a string representation of the 
+        number of machines currently running.
+        """
+        return f"{self.num_active_machines}/{self.max_machines}"
```
sergiomsantos, line 184: maybe give some feedback that the request refers to the max number of machines:

```suggestion
        return f"{self.num_active_machines}/{self.max_machines} (max)"
```

inductiva/resources/machines_base.py
```diff
@@ -104,6 +106,11 @@ def _register_machine_group(self, **kwargs):
     def __repr__(self):
         pass
 
+    def current_machines_to_str(self) -> str:
+        """Return the number of machines currently running in the MPI cluster.
```
sergiomsantos, line 110: Is this comment for MPI?

inductiva/resources/machines.py
```diff
@@ -42,6 +42,7 @@ def __init__(
                          data_disk_gb=data_disk_gb,
                          register=register)
         self.num_machines = num_machines
+        self._current_machines = 0
```
sergiomsantos, line 45: maybe rename to `_active_machines` and add a comment about this being the number of active running machines
at the time of the request?

```suggestion
        self._active_machines = 0
```

inductiva/resources/machine_cluster.py
sergiomsantos: please confirm with @begaiym-k this change to be sure nothing is breaking

inductiva/resources/machine_cluster.py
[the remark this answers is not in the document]
sergiomsantos: maybe the `_current_machines` attribute can make a backend request everytime to update its value.

inductiva/resources/machines_base.py
Pbarbosa92: No, copy pasted and forgot to change the docs.

IvanPombo (changes requested):
Nice push @Pbarbosa92! I think we should change a bit the nomenclature and make sure everything is on point, for example on registration and so on.

inductiva/resources/machine_groups.py
Pbarbosa92: Changed it to active_machines_to_str

inductiva/resources/machine_cluster.py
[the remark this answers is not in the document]
Pbarbosa92: > maybe the `_current_machines` attribute can make a backend request everytime to update its value.

We don't have that request yet. We could get all active machine groups and filter them to get to our machine group, but that does not make sense. We should have an api call to get a machine group by id.

inductiva/resources/machine_cluster.py
Pbarbosa92: > please confirm with @begaiym-k this change to be sure nothing is breaking

I had it wrong. num_vms never changed. It always had the current active machines. I will have num_machines for the active machines and _requested_machines for the max_vms

inductiva/resources/machine_cluster.py
```diff
@@ -79,7 +81,8 @@ def terminate(self):
 
     def _log_machine_group_info(self):
         super()._log_machine_group_info()
-        logging.info("> Number of machines: %s", self.num_machines)
+        logging.info("> Number of machines: %s of %s", self._current_machines,
+                     self.num_machines)
```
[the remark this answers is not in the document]
Pbarbosa92: You are correct. Changed

inductiva/resources/machines.py
```diff
@@ -155,6 +158,7 @@ def __init__(
         self.num_active_machines = min_machines
         self.__is_elastic = True
         self.spot = spot
+        self.num_machines = max_machines
```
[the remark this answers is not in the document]
Pbarbosa92: I was mistaken. I thought num_machines always had the max number of machines, but I was wrong. Now num_machines has the exact same value as before (num_vms) and I created _requested_machines that gets the max_vms

Pbarbosa92:
Fixes #1076

inductiva/resources/machine_cluster.py
```diff
@@ -38,6 +38,9 @@ def __init__(self,
                          data_disk_gb=data_disk_gb,
                          register=register)
         self.num_machines = num_machines
+        #Number of active machines at the time of
+        #the request machine_groups.get()
+        self._requested_machines = 0
```
IvanPombo, line 43: I am getting confused...

We set `num_machines`, which is technically the requested number of machines... Then, you establish a variable that is named `_requested_machines`, which technically is the actual number of machines requested. So it makes no sense, on the api_response to fetch the active number of machines to `num_machines`. 

So, to make it simpler, we should have a variable `num_machines` that always refers to the number of machines requested and can be used to reinitialize the class back again, from the number of `max_vms`. Then, you can have the `_active_machines` variable that receives from the backend the `num_vms` which refers to the active machines. But in this way, there are no conflicts, since the class created from the response is exactly the same. 

I think the names coming from the backend should also be formatted to address this, but we can deal with that after a more general refactor.

IvanPombo (changes requested):

inductiva/resources/machine_cluster.py
Pbarbosa92: The way it is now, ´num_machines´ is doing exactly the same as it was before the change. I don't want to change it because I am afraid it will break something somewhere.
So, by keeping ´num_machines´ the exact same, I hope to introduce fewer points of failure.

IvanPombo: It is in the object initialization, but not in the `api_response` where it seems to be receiving the `num_vms` which is the active machines, no?

Pbarbosa92: > It is in the object initialization, but not in the `api_response` where it seems to be receiving the `num_vms` which is the active machines, no?

Yes, but previously `num_vms` was also the active machines. Sometimes we had 0 machines on a machine group cause the machines were not active yet.

IvanPombo: I think it was until @begaiym-k changed it in the backend, because last week or so we weren't even capturing this.

inductiva/resources/machines.py
```diff
@@ -155,6 +160,7 @@ def __init__(
...
```
IvanPombo, line 163: I think this more extra for the elastic class, no? Because, we are doing everything with min, max, active machines. Especially if we override the àctive_machines_to_str`.

IvanPombo (commented):
Almost there, just one minor comment, and should be good to go!

inductiva/resources/machines.py
begaiym-k: I agree, here the num_machines is unnecessary.

inductiva/resources/machines_base.py
```diff
@@ -104,6 +108,11 @@ def _register_machine_group(self, **kwargs):
     def __repr__(self):
         pass
 
+    def active_machines_to_str(self) -> str:
+        """Return the number of machines currently running.
+        """
+        return f"{self.num_machines}/{self._active_machines}"
```
begaiym-k, line 114: Shouldn't it be the other way around?

inductiva/resources/machines_base.py
```diff
@@ -59,6 +59,10 @@ def __init__(self,
         self.create_time = None
         self._started = False
         self.register = register
+        #Number of active machines at the time of
+        #the request machine_groups.get()
+        self._active_machines = 0
```
begaiym-k, line 64: There is already a variable called `num_active_machines` in child classes, so you can whether set that one here instead of `_active_machines` for the consistency, or set the the ladder in the child classes.

Pbarbosa92: Changed num_active_machines to _active_machines

IvanPombo (approved):
LGTM! Minor comment regarding the string with (max). take this part for consistency.

inductiva/resources/machines.py
```diff
@@ -170,9 +175,15 @@ def from_api_response(cls, resp: dict):
...
         machine_group.min_machines = int(resp["min_vms"])
-        machine_group.num_active_machines = int(resp["num_vms"])
+        machine_group.__dict__["_active_machines"] = int(resp["num_vms"])
         return machine_group
 
+    def active_machines_to_str(self) -> str:
+        """Returns a string representation of the 
+        number of machines currently running.
+        """
+        return f"{self._active_machines}/{self.max_machines} (max)"
```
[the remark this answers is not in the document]
Pbarbosa92: @sergiomsantos suggested this change.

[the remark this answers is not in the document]
IvanPombo: Ok, so just merge.

Pbarbosa92:
**QA**

Machine Group
Captura de ecrã 2024-02-09, às 13 29 54

Elastic Machine Group
Captura de ecrã 2024-02-09, às 13 36 20

MPI Cluster
Captura de ecrã 2024-02-09, às 13 38 54

inductiva/resources/machines.py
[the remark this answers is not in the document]
Pbarbosa92: I just did QA but I don't know what is the policy on self-QA.
Should I wait for @sergiomsantos to review? Or can I merge it now ?

PR #1085, opened 2024-02-09
64-check all arguments to create machines and prevent unreasonable values

Pbarbosa92:
Check all arguments to create machine groups and make sure they have decent values.

For instance, we can create machine groups with -1 machines or with 0 machines.

Find the cases where that can happen or not, and raise exceptions when we need them.


IvanPombo (approved):
LGTM!

Pbarbosa92:
**QA**

Try to launch machine groups with negative numbers of instances.

AugustoPeres:
Tested for both MPI, MachineGroup and ElasticMachineGroup:
All raised the exception 
Screenshot 2024-02-16 at 09 53 06

sergiomsantos (approved):

PR #1093, opened 2024-02-14
[Simulator] Add SWAN client interface

IvanPombo:
TLDR: Adds the SWAN simulator to the client.

## Description

In this PR, we add the client interface to execute simulations with SWAN simulator that was added to the backend in the PR inductiva/inductiva-web-api#632.

This simulator runs exactly as SWASH with a single binary and MPI, so the user only needs to pass the input directory and the name of `sim_config_filename`. An example input dir was added to our respective cloud folder to allow for testing.

## Implications

The main application, that should be addressed soon and that here I was not strict is with the `api_method_name`. In a next PR, I will make sure to uniformize all of them together with the backend.

There are few threads open with SWAN, respectively the usage of OpenMP vs MPI when running the simulator. One discussion in the "literature" is that OpenMP is more effective in a single machine, while MPI is more effective over several. Here, we would need to separate images for each, since SWAN doesn't seem to allow for compilation of both at the same time. Due to our current structure and simplicity, I have followed with the implementation with MPI, and further conversations with users may decide our next steps.

A README will follow on a next iteration.

## Test

To test this simulator you can deploy the backend locally or wait for it to be merged into development. To do so, use the following script:

```python
import inductiva

    "swan-input-example.zip", True)

# Initialize the Simulator
swan = inductiva.simulators.SWAN()

task = swash.run(
    input_dir=input_dir, sim_config_filename="a11refr.swn")
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -58,6 +58,7 @@ simulators/GROMACS
 simulators/SWASH
 simulators/XBeach
 simulators/Reef3D
+simulators/SWAN
 simulators/FDS
 ```
 
--- a/docs/simulators/SWAN.md
+++ b/docs/simulators/SWAN.md
@@ -28,3 +28,6 @@ task = swash.run(
 task.wait()
 task.download_outputs()
 ```
+
+Check the [official documentation](https://swanmodel.sourceforge.io/) of SWAN to know 
+more about the configuration details specific of the simulator.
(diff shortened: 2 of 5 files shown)


Pbarbosa92 (approved):

AugustoPeres (approved):

sergiomsantos (commented):
We can also add the associated doc page

PR #1094, opened 2024-02-14
122 change n cores to n_vcpus and allow user to send number of vcpus

Pbarbosa92:
In this pr we we give the user more control over the simulations.

We now allow the user to send as a parameter to some simulations the number of cpus that he wants to run (n_vcpu).
Because some machines might have (or not) hyperthreading, we also allow the user to send a flag that will define if we use hyperthreading or not (use_threads_cpus - True by default).

With this the user can have a machine with 10 cores (20 threads) and decide if he wants his simulation:
- To run on 10 physical cores (n_vcpus=10, use_threads_cpus=False)
- To run on 10 threads (n_vcpus=10, use_threads_cpus=True)
- On all available threads (n_vcpus=20, use_threads_cpus=True)

**Possible problem**
The user can ask for 15 threads and ask for threading off, and it will get an error like this
"There are not enough slots available in the system to satisfy the 19
slots that were requested by the application:"

The problem is that this error only shows on the stderr in the task output files. The user will only see that the task failed.

We can let the error do the mistake **OR** create a back-end endpoint that tells us the number of cores for a machine type and we can generate an exception. 
The big issue with this is that the task can go to a queue for 2 hours, and the user will only have the task failed after 2 hours of waiting.

**Task** [here](https://github.com/inductiva/tasks/issues/122)


inductiva/utils/meta.py
```diff
@@ -59,3 +60,32 @@ def get_method_name(function_ptr) -> str:
     module_name = ".".join(module_name_useful_segments)
 
     return f"{module_name}.{function_ptr.__name__}"
+
+
+# pylint: disable=unused-argument
+def custom_format_warning(msg, *args, **kwargs):
+    """Custom format for warnings."""
+    return str(msg) + "\n"
```
sergiomsantos, line 68: why is this necessary?

inductiva/utils/meta.py
```diff
@@ -59,3 +60,32 @@ def get_method_name(function_ptr) -> str:
...
+
+
+def deprecated_arg(**kwargs_decorator):
```
sergiomsantos, line 71: doc string?

inductiva/utils/meta.py
```diff
@@ -59,3 +60,32 @@ def get_method_name(function_ptr) -> str:
...
+
+    def wrapper(func):
+
+        def inner_wrapper(*args, **kwargs):
+            decorator_keys = kwargs_decorator.keys()
+            function_keys = kwargs.keys()
+
+            warnings.formatwarning = custom_format_warning
```
sergiomsantos, line 79: why is this necessary?

inductiva/simulators/xbeach.py
```diff
@@ -12,19 +13,21 @@ def __init__(self):
...
+            on: Optional[types.ComputationalResources] = None,
+            storage_dir: Optional[types.Path] = "",
+            extra_metadata: Optional[dict] = None,
+            **kwargs) -> tasks.Task:
         """Run the simulation.
 
         Args:
             input_dir: Path to the directory of the simulation input files.
             sim_config_filename: Name of the simulation configuration file.
+            n_vcpus: Number of vCPUs (all by default) to use for the simulation.
```
sergiomsantos, line 30:
```suggestion
            n_vcpus: Number of vCPUs to use in the simulation. If not provided (default), all vCPUs will be used.
```

inductiva/utils/meta.py
```diff
@@ -59,3 +60,32 @@ def get_method_name(function_ptr) -> str:
...
+
+            for decorator_key in decorator_keys:
+                if decorator_key in function_keys:
+                    warnings.warn(f"{decorator_key} is deprecated, use "
+                                  f"{kwargs_decorator[decorator_key]} instead.")
+                    kwargs[
+                        kwargs_decorator[decorator_key]] = kwargs[decorator_key]
```
sergiomsantos, line 86: this can be made cleaner: no need to get values multiple times from the dict.
Also, to get the common keys you can use `set( kwargs_decorator) & set(kwargs)`

inductiva/simulators/fds.py
```diff
@@ -12,21 +13,22 @@ def __init__(self):
...
+            on: Optional[types.ComputationalResources] = None,
+            storage_dir: Optional[types.Path] = "",
+            extra_metadata: Optional[dict] = None,
+            **kwargs) -> tasks.Task:
         """Run the simulation.
 
         Args:
             input_dir: Path to the directory of the simulation input files.
             sim_config_filename: Name of the simulation configuration file.
+            n_vcpus: Number of vCPUs (all by default) to use for the simulation.
```
sergiomsantos, line 31: see my other comment

inductiva/utils/meta.py
Pbarbosa92: Without this the warning shows like this:
```
/Users/paulobarbosa/Documents/Gits/inductiva/inductiva/utils/meta.py:83): UserWarning: n_cores is deprecated, use n_vcpus instead.
  warnings.warn(f"{decorator_key} is deprecated, use "
```

Since meta.py is the code of the decorator that is meaningless to the user.

inductiva/simulators/xbeach.py
```diff
@@ -12,19 +13,26 @@ def __init__(self):
...
-        sim_config_filename: Optional[str] = "params.txt",
-        on: Optional[types.ComputationalResources] = None,
-        storage_dir: Optional[types.Path] = "",
-        extra_metadata: Optional[dict] = None,
-    ) -> tasks.Task:
+    @meta.deprecated_arg(n_cores="n_vcpus")
+    def run(self,
+            input_dir: types.Path,
+            n_vcpus: Optional[int] = None,
+            use_hwthread: bool = None,
```
IvanPombo, line 20: The default value of `use_hwthread` should be `True` everywhere. Let's assume users want to use all vcpus they have available.

inductiva/utils/meta.py
```diff
@@ -59,3 +60,35 @@ def get_method_name(function_ptr) -> str:
...
+        def inner_wrapper(*args, **kwargs):
+
+            for key in set(deprecated) & set(kwargs):
+                new_key = deprecated[key]
+                value = kwargs.pop(key)
+                kwargs[new_key] = value
+
+                warnings.warn(
+                    f"{key} is deprecated, use "
+                    f"{new_key} instead.",
```
IvanPombo, line 87:
```suggestion
                    f"The {key} argument was deprecated, and substituted with {new_key}.",
```

IvanPombo (commented):
Almost ready to go!

inductiva/simulators/openfoam.py
```diff
@@ -23,19 +24,26 @@ def __init__(self, version: str = "foundation"):
...
 
-    def run(
-        self,
-        input_dir: types.Path,
-        commands: types.Commands,
-        on: Optional[types.ComputationalResources] = None,
-        storage_dir: Optional[types.Path] = "",
-        extra_metadata: Optional[dict] = None,
-    ) -> tasks.Task:
+    @meta.deprecated_arg(n_cores="n_vcpus")
```
IvanPombo, line 27: This decorator should not be here, since the argument didn't exist at all for this simulator.

inductiva/simulators/reef3d.py
```diff
@@ -13,17 +14,24 @@ def __init__(self):
...
         self.api_method_name = "reef3d.reef3d.run_simulation"
 
-    def run(
-        self,
-        input_dir: types.Path,
-        on: Optional[types.ComputationalResources] = None,
-        storage_dir: Optional[types.Path] = "",
-        extra_metadata: Optional[dict] = None,
-    ) -> tasks.Task:
+    @meta.deprecated_arg(n_cores="n_vcpus")
```
IvanPombo, line 17: Same as above.

inductiva/simulators/swash.py
```diff
@@ -12,19 +13,26 @@ def __init__(self):
...
 
-    def run(
-        self,
-        input_dir: types.Path,
-        sim_config_filename: str,
-        on: Optional[types.ComputationalResources] = None,
-        storage_dir: Optional[types.Path] = "",
-        extra_metadata: Optional[dict] = None,
-    ) -> tasks.Task:
+    @meta.deprecated_arg(n_cores="n_vcpus")
```
IvanPombo, line 16: here as well.

inductiva/simulators/xbeach.py
```diff
@@ -12,19 +13,26 @@ def __init__(self):
...
 
-    def run(
-        self,
-        input_dir: types.Path,
-        sim_config_filename: Optional[str] = "params.txt",
-        on: Optional[types.ComputationalResources] = None,
-        storage_dir: Optional[types.Path] = "",
-        extra_metadata: Optional[dict] = None,
-    ) -> tasks.Task:
+    @meta.deprecated_arg(n_cores="n_vcpus")
```
IvanPombo, line 16: here as well

Pbarbosa92:
**QA**

To be testes at the same time as https://github.com/inductiva/inductiva-web-api/pull/633

inductiva/utils/meta.py
```diff
@@ -59,3 +60,35 @@ def get_method_name(function_ptr) -> str:
     module_name = ".".join(module_name_useful_segments)
 
     return f"{module_name}.{function_ptr.__name__}"
+
+
+# pylint: disable=unused-argument
+def custom_format_warning(msg, *args, **kwargs):
```
sergiomsantos, line 66: is this still necessary?

Pbarbosa92: Removed

PR #1123, opened 2024-02-22
Remove signal handler when submitting tasks

luispcunha:
The `blocking_task_context` context manager sets a custom signal handler. This becomes problematic when using the client in a multi threaded environment, since the `signal.signal` function can only be called in the main thread of an application.

I implemented this quick fix to have a branch of the client to use in a specific use-case where multi threaded is desired. We can discuss if this solution could be made definitive or discuss alternatives @IvanPombo @sergiomsantos 

Edit: I noticed now that the `blocking_task_context` uses `signal.signal` to set the default `SIGINT` handler. From the comment there, I seem to recall that the reason for this was that at the time we were using a dependency that set a custom `SIGINT` handler. I don't think that's the case anymore, so maybe we could keep the usage of the `blocking_task_context` but change it to not explicitly set the default `SIGINT` handler.

### DIFF
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -296,15 +296,14 @@ def submit_task(api_instance, method_name, request_params, resource_pool,
     task_id = task["id"]
     log_task_info(task_id, method_name, params, resource_pool)
 
-    with blocking_task_context(api_instance, task_id):
-        if task["status"] == "pending-input":
+    if task["status"] == "pending-input":
 
-            upload_input(
-                api_instance=api_instance,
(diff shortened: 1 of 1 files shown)


hpenedones:
Cool, sorry! I got the email from GitHub and didn't notice it was still in draft mode!

sergiomsantos:
Capturing SIGINT in multithreaded apps will aways be a problem.
For now, we can accept this change and start thinking about a wider approach to thread safety.

sergiomsantos (approved):

PR #1136, opened 2024-02-27
65 log task end after no more logs and logs pick up on the correct message when reconnecting from timeout

Pbarbosa92:
Previously, our logs would hang connected forever after a task ended. The user had to use control+c to leave the logs.

Now the logs end when the stream ends. And when the user has a timeout, the logs pick up on the last line received by the client, preventing the user from losing important logs.

**The user can still lose logs because we have a limit of 100000 lines of logs to query. So, if the timeout was long enough that the stream generated more than 100000 lines the user would lose some.**

[inductiva/tasks#65](https://github.com/inductiva/tasks/issues/65)

### DIFF
--- a/inductiva/tasks/streams.py
+++ b/inductiva/tasks/streams.py
@@ -10,7 +10,8 @@ $ induction logs <task_id>
                                                     `SPACING` black lines
  ● 0:00:02 Streaming output from task <task_id>     connection status footer
 """
-from threading import RLock, Thread
+from threading import Thread, RLock
+from urllib.parse import urlencode
 from datetime import timedelta
 from typing import IO
 import logging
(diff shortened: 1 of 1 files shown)


AugustoPeres (approved):
LGTM!! Call me for Q/A time

Pbarbosa92:
**QA**

- Start a task. Stream the logs to the cli and use ctrl+c to check if keyboard interruptions are working.
- Go to streams.py and add time.sleep(0.5) to the end of the on_message function (this will generate a timeout after 50 seconds). Now run an open foam task and log the results. It should give a timeout after 50 seconds. After some time, we will reconnect, and we will keep getting the logs from the previous line where we lost connection. To check if there are no logs missing, download the output files and check the line where you had the timeout on the terminal and the stdout file. See if the following lines on the file are equal to the ones on the terminal.

Pbarbosa92:
**QA**

Connects back when we get a timeout (can connect multiple times). ✅ 
Stops stream when we get end of stream. ✅ 
Redirected output to txt file (with timeouts). ✅ 
Redirected output to txt and redirect err to another txt file (with timeouts). ✅

sergiomsantos (approved):

PR #1178, opened 2024-03-07
Fix pager and watch bugs

sergiomsantos:
This PR:
- fixes a bug in the scrolling mechanism that allowed for negative lines when the buffer was empty;
- ensures that the body and footer are properly cleared when the `clear` method is called;
- ensures all threads run as daemon of the main thread and sets a timeout when joining the scheduler thread to allow for better user experience.


IvanPombo (approved):
On the one side it LGTM, but I will QA tomorrow for further verification!

PR #1180, opened 2024-03-08
[Fenicsx] provide the mesh as an input

Joao97ribeiro:
This PR introduces the ability to provide a mesh as input for the simulator. This enhancement eliminates the need to generate the mesh within the API, offering users the flexibility to supply their own. This feature is particularly valuable when the same mesh is required across different scenarios.


IvanPombo (approved):
Top! LGTM!

PR #1182, opened 2024-03-08
Refactor template mechanism to be more user friendly

IvanPombo:
This PR refactors that templating mechanism to be more user-friendly and with less ambiguity in what the functions do.

TLDR: The templating mechanism is now fully done with one class and the mechanisms itself is inherited from the JinjaRendered and the FileManager.

## Description

The main strike points of the previous mechanism were:
- The `FileManager` was not so intuitive to document and explain the template mechanism;
- The functions `add_dir` and `add_file` were doing two things at the same time and weren't clear what is doing what.

A first iteration of a refactor was proposed in inductiva/inductiva#1160. After discussion we decided that was still complicated and decided with the following reasoning.

With this two conditions, we broke apart the `TemplateEngine` into two pieces: the functions that just manage files as they are - copying files and directories - and; the functions that perform the rendering.
- The former is still performed by a cleaner `FileManager` interface, which the nomenclature remains, but now it has the functions `copy_file` and `copy_dir` which do exactly what the name describes.
- The latter is embedded in the `TemplateEngine` class, but the actual mechanism that performs the rendering can be changed. That is, the `TemplateEngine` class has the `render_file` and `render_dir` methods that perform the rendering in a higher-level abstraction. This functions use a `rendered` object that can be modified on initialisation of  the engine and where the default is the `JinjaRenderer`.

In this case, the `JinjaRenderer` is the low-level class that actually implements the rendering mechanism based on how `jinja2` works.

## How it feels

Let's imagine that we now have a template directory prepared for a wind tunnel simulation with OpenFOAM. This directory is named as `windtunnel_templates` and is sitting in our current working dir. The template files have two tags: `wind_speed` and `resolution` that can be modified. Moreover, we want to add an object file from outside this directory.

In the end, we want an input directory ready to run the simulation. The code workflow is as follows:

```python
import inductiva

# Initialise the TemplateEngine with the directory of templates and a root directory where all files will go to.
template_engine = inductiva.TemplateEngine(template_dir="windtunnel_templates", base_dir="windtunnel_input_dir")

# Render full directory, which keeps the structure in the base directory.
template_engine.render_dir(wind_speed=10, resolution=3)
# Copy file from my current working dir to the root_dir.
template_engine.copy_file("vehicle.obj")

simulator = inductiva.simulators.OpenFOAM()
comands = [...]

task = simulator.run(input_dir=template_engine.get_root_dir(), commands=commands)
```

This is a simple example. But we could do more specific things, like render the folder `constant` from within `windtunnel_templates` with:
```python
template.render_dir(source_dir="constant", wind_speed=10)
````

Moreover, we can just render files and copy directories.

## Tests

A full-set of tests was added to complement the structure of the templating mechanism. It is broken into:
- `tests/utils/test_file_manager.py`;
- `test/templating/test_template_engine.py`;
- `test/templating/test_renderer.py`.

### CODE REVIEW

File: inductiva/tests/templating/test_template_engine.py [hunk 1]
```diff
@@ -0,0 +1,129 @@
...
+
+
+@pytest.fixture
+def fixture_templatr():
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        os.chdir(tmpdirname)
+        yield TemplateEngine(ASSETS_DIR)
+
+
+@pytest.fixture(name="templatr")
```
Pbarbosa92, line 20: Typo "template"

Pbarbosa92 (approved):
LGTM! Jost one minor comment

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+
+        return utils.files.map_dir_files(source_dir, cls.TEMPLATE_EXTENSION)
+
+
+class JinjaRenderer(BaseRenderer):
+    """Jinja2 renderer for .jinja templates."""
+
+    TEMPLATE_EXTENSION = ".jinja"
+
+    @classmethod
```
sergiomsantos, line 58: why a class method?

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+    def render_dir(cls, source_dir: types.Path, target_dir: types.Path,
+                   **render_args):
```
sergiomsantos, line 60:
```suggestion
                   **render_args) -> Path:
```

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+                        target_dir,
+                        dirs_exist_ok=True,
+                        symlinks=True)
+
+        template_dir_struct = cls.get_template_files(target_dir)
+
+        # Render the template files from within the target directory.
+        environment = jinja2.Environment(
+            loader=jinja2.FileSystemLoader(target_dir),
+            undefined=jinja2.StrictUndefined)
```
sergiomsantos, line 86: this should be unique to the instance. the shouldn't be any need to instantiate a new environment every time a file is rendered. Look at how it was done in https://github.com/inductiva/inductiva/pull/1160/files. It was pretty stable and clean.

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+            target_file (types.Path): Path to the target file.
+            source_file (types.Path): Path to the source file.
+            render_args: Arguments to render the template file.
+        """
+
+        source_path = pathlib.Path(source_file)
+
+        environment = jinja2.Environment(loader=jinja2.FileSystemLoader(
+            source_path.parent),
+                                         undefined=jinja2.StrictUndefined)
```
sergiomsantos, line 113: move to ctor.

inductiva/tests/templating/test_template_engine.py
```diff
@@ -0,0 +1,129 @@
...
+import tempfile
+
+import pytest
+
+from inductiva import TemplateEngine
+
+ASSETS_DIR = pathlib.Path(__file__).parent / "assets"
+
+
+@pytest.fixture
```
sergiomsantos, line 13:
```suggestion
@pytest.fixture(name="templatr")
```

inductiva/tests/templating/test_template_engine.py
sergiomsantos, line 20: this one (and the ones below) are not fixtures; they are using a fixture. They shouldn't have the decorator.

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
+"""Base and specific renderers packages to manage template files."""
+from abc import abstractmethod
+from typing import Iterable, Union, IO
```
sergiomsantos, line 3:
```suggestion
from typing import Iterable, Union, IO
from abc import abstractmethod
```

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @abstractmethod
+    def render_dir(self, target_dir: types.Path, source_dir: types.Path,
+                   **render_args):
+        """Render a template to a file."""
+        pass
+
+    @abstractmethod
+    def render_file(self, target_file: types.Path,
+                    source_file: Union[types.Path, IO], **render_args):
```
sergiomsantos, line 26:
```suggestion
    def render_file(self, target_file: types.Path,
                    source_file: Union[types.Path, IO], **render_args): -> Path
```

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+        """Render a file."""
+        pass
+
+    @classmethod
+    def is_template(cls, file: types.Path) -> bool:
```
sergiomsantos, line 31:
```suggestion
    def is_template(cls, file: types.PathOrStr) -> bool:
```

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+        """Check if the given file is of template type."""
+        if isinstance(file, pathlib.Path):
+            return file.suffix == cls.TEMPLATE_EXTENSION
```
sergiomsantos, line 34: Not sure how this works with `cls.TEMPLATE_EXTENSION` that start with a dot (ex. `.jinja`)

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+        if isinstance(file, pathlib.Path):
+            return file.with_suffix("")
+        return os.path.splitext(file)[0]
+
+    @classmethod
+    def get_template_files(cls, source_dir: types.Path) -> Iterable[types.Path]:
+        """Return a list of template files in the given source directory."""
+
+        return utils.files.map_dir_files(source_dir, cls.TEMPLATE_EXTENSION)
+
```
sergiomsantos, line 51: I don't like this method here, nor do I like having this `map_dir_files` function in utils. It should be part of the file manager.

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+
+class JinjaRenderer(BaseRenderer):
+    """Jinja2 renderer for .jinja templates."""
+
+    TEMPLATE_EXTENSION = ".jinja"
```
sergiomsantos, line 56: check how this relates with the comment in the `renderer.is_template` classmethod.

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+import pathlib
+import shutil
+import os
+
+import jinja2
+
+from inductiva import types, utils
+
+
+class BaseRenderer:
```
sergiomsantos, line 13: Please use the original name. This is not a base renderer.

inductiva/templating/renderers.py
sergiomsantos, line 1: Please use the original implementation from the other draft PR. It was clean, working and decoupled.

inductiva/templating/engine.py
```diff
@@ -0,0 +1,142 @@
...
+        if not overwrite:
+            try:
+                self._validate_destination(source_dir, target_dir)
+            except FileExistsError as e:
+                msg = f"{e}; set `overwrite=True` to overwrite existing files."
+                raise FileExistsError(msg) from e
+
+        self.renderer.render_dir(source_dir=source_dir,
+                                 target_dir=target_dir,
+                                 **render_args)
```
sergiomsantos, line 122: let's move the original logic here. The purpose of the Renderer adapter is simply to interface the actual rendering engine that works based on a single file. This class should not offload the work there, otherwise we have coupled concerns.

sergiomsantos (changes requested):
why didn't you use my branch as starting point? I see that some of the decoupling efforts I made were reverted.

IvanPombo:
> why didn't you use my branch as starting point? I see that some of the decoupling efforts I made were reverted.

The main reason was that I already had a branch also prepared with a different version of the templating engine, before we did the final design. This new branch ends up combining the two.

inductiva/types.py
```diff
@@ -1,9 +1,10 @@
 """Type definitions."""
 import os
+import pathlib
 from typing import Union, List
 from typing_extensions import TypeAlias
 
-Path: TypeAlias = Union[os.PathLike, str]
+Path: TypeAlias = Union[os.PathLike, pathlib.Path, str]
```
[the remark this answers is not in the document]
IvanPombo: This can break a lot of things, since it is used elsewhere so I will open an issue with it.

inductiva/templating/engine.py
```diff
@@ -0,0 +1,142 @@
...
+
+        return target_dir
+
+    def _validate_destination(self, source_dir: pathlib.Path,
```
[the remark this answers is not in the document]
IvanPombo: There is an equivalent one in the FileManager, here we are specifying the ones with the Template Extension, since we want to check if there aren't any inside the target_dir already with the same as a template file without the extension. So, the reason to fail is slight different.

inductiva/templating/renderers.py
IvanPombo: This is done, because we can template different dirs now, with the `source_dir` being passed. In the PR you link, we were stucking the user to have to instantiate a new object of Template type, every time we would want to change the template_dir. So this brings the best of both worlds and allows the user to have a single base template dir with all his jinja files.

inductiva/templating/renderers.py
IvanPombo: Can you elaborate? I felt that this was a general function, so there wasn't a reason to attach it to the file_manager. Moreover, this is used in a different manner inside the TemplateEngine.

inductiva/templating/renderers.py
sergiomsantos: The `TemplateEngine` class is instantiated with a template dir which is supposed to be the source of all templates. It makes sense to point the environment to that dir.

inductiva/types.py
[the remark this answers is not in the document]
sergiomsantos: Just create a new type called `PathOrStr: TypeAlias = Union[os.PathLike, pathlib.Path, str]` and use that in the new code. Latter we'll clean the other ones.

inductiva/templating/engine.py
```diff
@@ -0,0 +1,142 @@
...
+                              target_dir: pathlib.Path):
+        """Check if the target directory exists and is empty."""
+
+        source_map = self.renderer.get_template_files(source_dir)
```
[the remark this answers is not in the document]
sergiomsantos: why? the map maps all file and subdirs.

inductiva/templating/renderers.py
```diff
@@ -0,0 +1,120 @@
...
+    """Abstract base class for template renderers."""
+
+    TEMPLATE_EXTENSION = None
+
+    @abstractmethod
+    def render_dir(self, target_dir: types.Path, source_dir: types.Path,
+                   **render_args):
```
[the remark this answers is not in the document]
sergiomsantos: > I noticed your implementation, and it gets stuck to a single BASE TEMPLATE DIR, so it doesn't allow versatility to only use subdirs inside that base TEMPLATE Dir.

yes it does.

inductiva/templating/engine.py
IvanPombo: Yes, but to be fair, we are not even aware how other renderers may work. So, I don't see the problem of offloading the work there, since each specific adapter works in their own ways to render everything. Maybe some renders, even are able to render a directory without much concerns to iterate over the files.

inductiva/templating/renderers.py
IvanPombo: This works!

sergiomsantos (approved):

IvanPombo:
## QA - Passed

To QA the current implementation, I have used the current version here presented to test in the windtunnel-app.

The windtunnel scenario performs a rendering of a full directory containing the templates and copies an object into the input directory. Everything worked as expected and the simulation performed wonderfully.

PR #1199, opened 2024-03-14
Add google analytics to docs

tiagoafpereira:
This should provide some insights on the current usage of Inductiva's documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -207,6 +207,7 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+docs/.idea/
 
 # PyBuilder
 .pybuilder/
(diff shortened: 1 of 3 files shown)


hpenedones (approved):
Great, let's see if this starts showing up on the dashboards... :)

IvanPombo (approved):
This is super cool!! This ID refers to our general account correct? So we can use a single dashboard to monitor the analytics of the website and docs, etc...

tiagoafpereira:
> This is super cool!! This ID refers to our general account correct? So we can use a single dashboard to monitor the analytics of the website and docs, etc...

That is correct, it uses the same ID we are using on the website

PR #1204, opened 2024-03-14
Fix docs path

Pbarbosa92:


.github/workflows/build-docs.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       - uses: ammaraskar/sphinx-action@master
         with:
-          docs-folder: "docs/"
+          docs-folder: "source/docs/"
           build-command: "sphinx-build . ../build"
```
hpenedones, line 29: but then here probably you have to do `sphinx-build . ../../build`

hpenedones (approved):

PR #1208, opened 2024-03-15
Fix bug of `argument of type "NoneType" is not iterable` detected on Windows

IvanPombo:
This PR addresses issue inductiva/tasks#144 where Windows machines were not being able to terminate their resources or kill their tasks from the CLI.

This bug arises from the usage of the `localization` module only on the CLI and just being prompted in Windows machines.

The reason was that when `translations` had a None value, we were still checking if a key was inside a NoneType.
Apparently, if I understood correctly, Windows doesn't handle this cases properly and throws the error to the user. At least according to this: https://stackoverflow.com/questions/6762446/typeerror-argument-of-type-nonetype-is-not-iterable-in-python.

Check first if it is None fixes the bug.

### DIFF
--- a/inductiva/localization/localization.py
+++ b/inductiva/localization/localization.py
@@ -77,19 +77,21 @@ class Translator:
             if lang in self._translations:
                 return
 
-            self._load(lang)
-            self._lang = lang
+            if self._load(lang):
+                self._lang = lang
 
-    def _load(self, lang: str) -> None:
+    def _load(self, lang: str) -> bool:
         """Load the translation file for the given language.
 
+        Returns True if the language file was successfully
+        loaded, False otherwise.
         NOTE: This method is not thread-safe.
 
         Args:
             lang (str): Name of the language to load.
         """
         if not lang:
-            return
+            return False
 
         lang = lang.lower()
         pathname = self.TRANSLATIONS_DIR / f"{lang}.json"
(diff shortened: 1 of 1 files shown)


inductiva/localization/localization.py
```diff
@@ -141,7 +141,7 @@ def __getitem__(self, key: str) -> str:
             key (str): Key for the translation to get.
         """
         translations = self._translations.get(self._lang)
-        if key not in translations:
+        if translations is None or key not in translations:
```
sergiomsantos, line 144: strange, I wonder what the value of `self._lang` is.

sergiomsantos: At this point, if `locale` is not set, `self._lang` should be `_INDUCTIVA_DEFAULT_LANG` 🤔

IvanPombo: Cannot address the lang part now, but the highlight is that the `translation` was None. That maybe pointing to a deeper bug where the language is set correctly, but it may not appropriately refer to the language we have. For example, it could be `pt-br`.

@Pbarbosa92 can you try on your windows computer what is the language here? Just put a print for the self._lang and call a terminal command of the terminate a resource.

IvanPombo: Changing the default language on the Mac keeps working and it prints `self._lang` as pt, so it needs to be something more inherent of Windows.

AugustoPeres (approved):

AugustoPeres:
Do we have a QA procedure for windows?

Pbarbosa92 (approved):

PR #1218, opened 2024-03-20
Add ice to client

IvanPombo:
This PR adds the ability to launch simulations in ICE (Inductiva Computing Cluster). There were a lot of small tweaks to be added, but in general the interface is super simple, one just needs to pass the provider on the `MachineGroup` class.

In general, I don't really like the current interface, because the ICE workflow is very different than GCP, that is, by launching a machine group in ICE, that machine is not reserved completely to me, and other people can submit to the same machine group.

Moreover, the names also don't seem appropriate at the moment, since in GCP `c2` refers to Intel Xeon, here it is Intel i9, and the vCPUs also break the logic of the GCP interface.

### DIFF
--- a/inductiva/resources/machine_types.py
+++ b/inductiva/resources/machine_types.py
@@ -20,25 +20,22 @@ def get_available():
     return machine_types
 
 
-def list_available_machines():
+def list_available_machines(provider: str):
     """List all available machines types."""
 
     resources_available = get_available()
-
+    provider_resources = resources_available[provider]
     machine_types = []
 
-    # Fetch the provider information
-    for _, provider_resources in resources_available.items():
-        # Fetch the available machine CPU series
-        for cpu_series, series_info in provider_resources["cpu-series"].items():
-            # Fetch the available RAM types and vCPUs info
-            for ram_type, type_info in series_info["types"].items():
-                vcpus = type_info["vcpus"]
-                machine_types.extend(
-                    [f"{cpu_series}-{ram_type}-{vcpu}" for vcpu in vcpus])
-                if type_info["lssd"]:
-                    for vcpu in vcpus:
-                        machine_types.append(
-                            f"{cpu_series}-{ram_type}-{vcpu}-lssd")
+    # Fetch the available CPU series for the given provider
+    for cpu_series, series_info in provider_resources["cpu-series"].items():
+        # Fetch the available RAM types and vCPUs info
+        for ram_type, type_info in series_info["types"].items():
+            vcpus = type_info["vcpus"]
+            machine_types.extend(
+                [f"{cpu_series}-{ram_type}-{vcpu}" for vcpu in vcpus])
+            if type_info["lssd"]:
+                for vcpu in vcpus:
+                    machine_types.append(f"{cpu_series}-{ram_type}-{vcpu}-lssd")
 
     return tuple(machine_types)
--- a/inductiva/resources/machines.py
+++ b/inductiva/resources/machines.py
@@ -1,7 +1,18 @@
 """Classes to manage different Google Cloud machine group types."""
 from absl import logging
 
+from typing import Literal
 from inductiva.resources import machines_base
+
+def _check_ice_args(num_machines: int, spot: bool):
+
+    if num_machines > 1:
+        raise ValueError(
+            "ICE provider only supports launching one machine at a time.")
+
+    if spot:
+        raise ValueError(
+            "ICE provider only supports persistent machine launch.")
 
 
 class MachineGroup(machines_base.BaseMachineGroup):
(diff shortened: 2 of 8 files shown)


inductiva/resources/machines.py
```diff
@@ -15,6 +27,7 @@ class MachineGroup(machines_base.BaseMachineGroup):
     def __init__(
         self,
         machine_type: str,
+        provider: Literal["GCP", "ICE"] = "GCP",
```
AugustoPeres, line 30: Why not lower case for these defaults as well

Pbarbosa92 (approved):
LGTM!

inductiva/resources/machines.py
sergiomsantos, line 30: why not an enum?

inductiva/tasks/methods.py
```diff
@@ -45,7 +45,7 @@ def to_dict(list_of_tasks: Iterable[Task]) -> Mapping[str, List[Any]]:
         if executer is None:
             resource_type = None
         else:
-            resource_type = executer["vm_type"]
+            resource_type = executer["host_type"] + " " + executer["vm_type"]
```
sergiomsantos, line 48: I've seen this in multiple places (starting at the backend). It should be `executor` and not `executer` 😉

inductiva/tasks/task.py
```diff
@@ -488,9 +488,8 @@ def get_machine_type(self) -> Optional[str]:
             return None
 
         machine_info = info["executer"]
-        if "vm_type" not in machine_info:
-            return "inductiva-machine"
 
+        machine_provider = machine_info["host_type"]
```
sergiomsantos, line 492: I've seen `provider`, `provider_id` and here `host_type` referring to the provider. Can't we have a common name?
Do you know why the provider is being returned as `host_type`?

inductiva/resources/machines_base.py
```diff
@@ -43,13 +45,15 @@ def __init__(self,
                 example, when retrieving with the `machines_groups.get` method.
                 Users should not set this argument in anyway.
         """
-        if machine_type not in inductiva.resources.list_available_machines():
-            raise ValueError("Machine type not supported")
+        if machine_type not in inductiva.resources.list_available_machines(
+                provider.lower()):
+            raise ValueError(f"Machine type not supported in {provider}")
```
sergiomsantos, line 50: just to prevent the weird line break:

```suggestion
        provider = provider.lower()
        available = inductiva.resources.list_available_machines(provider)
        if machine_type not in available:
            raise ValueError(f"Machine type not supported in {provider}")
```

inductiva/resources/machine_types.py
```diff
@@ -20,25 +20,22 @@ def get_available():
     return machine_types
 
 
-def list_available_machines():
+def list_available_machines(provider: str):
```
sergiomsantos, line 23: can't we use get this listing directly from the BE?

```json
[
  {
    "machine_type": "n2d-highcpu-128",
    "num_cpus": 128,
    "ram_gb": 128,
    "price": 1.21344,
    "provider": "GCP",
    "spot": true,
    "region": "europe-west1"
  },
  ...
]
```

inductiva/tasks/task.py
sergiomsantos: FYI @lneves-ai
Can we standardize this nomenclature?

inductiva/resources/machines.py
IvanPombo: Because, the backend established it like this...

inductiva/resources/machine_types.py
IvanPombo: Will do in another PR.

IvanPombo:
## QA

For anyone that would like to try ICE - pointing to @AugustoPeres and @Pbarbosa92 here - you can launch a simulation as it is normal, but you should start an ICE machine as follows:
```python
machine = inductiva.resources.MachineGroup("c2-highmem-4", provider="ICE")
machine.start()
```

Moreover, you can check the machines available for ICE with `inductiva resources available -p ice`.

To check we are validating correctly, please try to launch a different machine type for ICE like `c2-standard-4` and that a different provider like  `AWS` does not work. To finish, you can attempt to launch as before, and pass the variables `spot=True` and `num_machines=2` to see if we capture an error to not allow users to launch these configs.

AugustoPeres:
## QA

I can launch the machines and I can submit simulations. However, the simulations fail. Is this expected?
Also I am having this strange behaviour on the `cli`:

Screenshot 2024-03-21 at 09 55 52

When using `provider=aws` an error is raise (as expected)
Screenshot 2024-03-21 at 09 57 19

inductiva/tasks/methods.py
sergiomsantos: As far as I can tell, _executer_ doesn't exist [in the Cambridge dictionary](https://dictionary.cambridge.org/spellcheck/english/?q=executer). The [Oxford one](https://www.oed.com/dictionary/executor_n?tab=meaning_and_use#4988318) points to an _executor_ as being  _an agent, doer_, and the same as executer. [Wikipedia](https://en.wikipedia.org/wiki/Executer) points to a band.
All uses of the _execut*r_ expression in tech refer to _executor_.

PR #1229, opened 2024-03-22
Add timestamps to auto-terminating machines

IvanPombo:
This PR adds the capability for users to add a `max_idle_time` and `auto_terminate` args in the `start` methods of their machines, in order to control their termination time.

## Implications

Now, users can configure their machine groups as they usually did, and when deciding to start their machines they can
add the following args:
- `max_idle_time`: Time in mins the machine shall be up without running any simulations;
- `auto_terminate`: Timestamp or float representing the hours until a machine group will automatically terminate itself, independently of any tasks being running on it or not.

At the moment, I have added these features to the elastic machine group, but in the future, with the autoscaling, I am not so sure if we should maintain these.

## QA:

Create a machine of any type, and start it with variations of the above parameters. Wait patiently on termination watch.


inductiva/resources/machines.py
```diff
@@ -221,13 +236,28 @@ def __str__(self):
...
+    def start(self,
+              max_idle_time: float = None,
+              auto_terminate: Union[float, str] = None):
+        """Starts minimum number of machines.
+
+        Args:
+            max_idle_time (float): Time in minutes that the machine can remain
+                idle.
+            auto_terminate (float, str): Time to automatically terminate the
+                machines independently of any simulations being running there.
```
AugustoPeres, line 248: Maybe: "independently of any simulations running there"

inductiva/resources/machines_base.py
```diff
@@ -142,10 +143,61 @@ def from_api_response(cls, resp: dict):
...
+        """Helper function to update the termination timers of a machine group.
+
+        At the moment, this serves as an helper function, but it could also be
+        provided to the users on demand.
+
+        Args:
+            max_idle_time (float): Time in minutes that the machine can remain
+                idle.
+            auto_terminate (float, str): Time to automatically terminate the
+                machines independently of any simulations being running there.
```
AugustoPeres, line 158: same

inductiva/resources/machines_base.py
```diff
@@ -142,10 +143,61 @@ def from_api_response(cls, resp: dict):
 
         return machine_group
 
-    def start(self, **kwargs):
+    def _update_termination_timers(self,
```
AugustoPeres, line 146: Agree that this should be exposed to the user. It can be helpful if, for example the user, underestimated the time some simulation would take

AugustoPeres (approved):

inductiva/resources/machine_cluster.py
```diff
@@ -73,12 +74,26 @@ def __str__(self):
         return f"MPI Cluster {self.name} with {self.machine_type} " \
                f"x{self.num_machines} machines"
 
-    def start(self):
-        """Start the MPI Cluster."""
+    def start(self,
+              max_idle_time: float = None,
+              auto_terminate: Union[str, float] = None):
```
sergiomsantos, line 79: A bit of inconsistency here: `max_idle_time` is in minutes but `auto_terminate` can be in hours.
In either case, I don't like any of them because they're hard to understand and give no indication of the units.

I suggest using existing python primitives :

```suggestion
              max_idle_time: timedelta = None,
              auto_terminate: datetime = None):
```

inductiva/resources/machines.py
```diff
@@ -95,11 +95,26 @@ def __repr__(self):
     def __str__(self):
         return f"Machine Group {self.name} with {self.machine_type} machines"
 
-    def start(self):
-        """Starts all machines of the machine group."""
+    def start(self,
+              max_idle_time: float = None,
+              auto_terminate: Union[str, float] = None):
```
sergiomsantos, line 100: same here

inductiva/resources/machines.py
```diff
@@ -221,13 +236,28 @@ def __str__(self):
         return f"Elastic Machine Group {self.name} with {self.machine_type} " \
              "machines"
 
-    def start(self):
-        """Starts minimum number of machines."""
+    def start(self,
+              max_idle_time: float = None,
+              auto_terminate: Union[float, str] = None):
```
sergiomsantos, line 241: and here

inductiva/resources/machines.py
```diff
@@ -221,13 +236,28 @@ def __str__(self):
...
+        """Starts minimum number of machines.
+
+        Args:
+            max_idle_time (float): Time in minutes that the machine can remain
+                idle.
+            auto_terminate (float, str): Time to automatically terminate the
+                machines independently of any simulations running there.
+                The time can be a float, indicating the number of hours the
+                machine up until the machine can be up, or an actual timestamp
+                with the format '2024-12-31T00:00:00+00'.
```
sergiomsantos, line 251: it is weird to define a str manually in this format.
```suggestion
                with the format '2024-12-31T00:00:00+00'.
```

inductiva/resources/machines_base.py
```diff
@@ -55,15 +56,15 @@ def __init__(self,
         """
 
         provider = ProviderType(provider.upper())
+        self.provider = provider.value
 
-        if machine_type not in list_available_machines(provider.value.lower()):
-            raise ValueError(f"Machine type not supported in {provider}")
+        if machine_type not in list_available_machines(self.provider.lower()):
```
sergiomsantos, line 61: `list_available_machines` should receive either a `str` or a `ProviderType`.

```python
def list_available_machines(provider: str|ProviderType):
    provider = ProviderType(provider)   # if ProviderType is case insensitive, this will work
    ...
```

```suggestion
        if machine_type not in list_available_machines(self.provider):
```

inductiva/resources/machines_base.py
```diff
@@ -132,20 +133,72 @@ def from_api_response(cls, resp: dict):
...
+            auto_terminate (float, str): Time to automatically terminate the
+                machines independently of any simulations running there.
+                The time can be a float, indicating the number of hours the
+                machine up until the machine can be up, or an actual timestamp
+                with the format '2024-12-31T00:00:00+00'.
+        """
+
+        # Convert max_idle_time from minutes to seconds
+        if max_idle_time is not None:
+            max_idle_time = max_idle_time * 60
```
sergiomsantos, line 166:
```suggestion
        max_idle_secs = max_idle_time.total_seconds() if max_idle_time else None
```

inductiva/resources/machines_base.py
```diff
@@ -132,20 +133,72 @@ def from_api_response(cls, resp: dict):
...
+
+        # Convert float auto_terminate_ts to ISO format
+        if isinstance(auto_terminate, float):
+            auto_terminate = datetime.datetime.now(datetime.timezone.utc) + \
+                datetime.timedelta(hours=auto_terminate)
+            auto_terminate = auto_terminate.isoformat()
```
sergiomsantos, line 172: having `auto_terminate` be a `datetime` object will largely simplify this logic.
You'll still have to check that `auto_terminate > now`!

inductiva/resources/machines_base.py
```diff
@@ -132,20 +133,72 @@ def from_api_response(cls, resp: dict):
...
         machine_group._id = resp["id"]
+        machine_group.provider = resp["provider_id"]
         machine_group._name = resp["name"]
         machine_group.create_time = resp["creation_timestamp"]
         machine_group._started = True
 
         return machine_group
 
-    def start(self, **kwargs):
+    def update_termination_timers(self,
```
sergiomsantos, line 146: this is a helper method and can be hidden.

```suggestion
    def _update_termination_timers(self,
```

sergiomsantos (changes requested):
Units are a bit weird.

IvanPombo:
## QA

@AugustoPeres after this is merged, you can work on yours to clean up!

Launch any machine group type, set `max_idle_time` with `timedelta` and `auto_terminate_ts` with `datetime` and verify that the machine actually terminates at that time.

inductiva/resources/machines_base.py
```diff
@@ -136,16 +133,65 @@ def from_api_response(cls, resp: dict):
...
+              auto_terminate_ts: Union[str, float] = None,
+              **kwargs):
         """Starts a machine group.
 
         Args:
+            max_idle_time: The maximum amount of time a machine group can
+              remain idle before it is automatically terminated (in minutes).
+            auto_terminate_ts: The timestamp at which the machine group will
+                be automatically terminated. The timestamp should be in the
+                format "YYYY-MM-DDTHH:MM:SS+00".
```
sergiomsantos, line 194: see my other comment

sergiomsantos (changes requested):
an implemetation of the `start` method still needs to be made consistent.

sergiomsantos (approved):

Pbarbosa92:
**QA**

Started a machine group with a 30 seconds max_idle_time. After a minute or 2 the machine was gone (which is the expected result) ✅ 
Starting a machine group with auto_terminate fails with an error.
Code used
```
import inductiva
import datetime
import pytz

mg=inductiva.resources.MachineGroup(machine_type="c2d-standard-4",num_machines=1)

now=datetime.datetime(2024,3,25,14,50,tzinfo=pytz.timezone('Europe/Lisbon')) 
mg.start(auto_terminate_ts=now)
```
Captura de ecrã 2024-03-25, às 14 48 29

Adding `if auto_terminate_ts < datetime.datetime.now(auto_terminate_ts.tzinfo):` this to the line of the error seemed to fix it.

Pbarbosa92:
Even after the fix my machine is not getting terminated when using auto_terminate_ts

AugustoPeres:
## QA

* `max_idle_time` seems to work fine. Started a machine with five minutes iddle time. After about 5/6 minutes the machine was gone ✅ ;
* `auto_terminate_ts`. The machine persists after the time_stamp given. Code used:

```python
 mg.start(auto_terminate_ts=datetime.datetime(2024,3,25,15,7,tzinfo=pytz.timezone('Europe/Lisbon'))
```

## Other behaviour to think about

If, for example in python interpreter I use:

```python
mg.start(max_idle_time=datetime.timedelta(minutes=5))
```

After 5 minutes the machine is gone. Therefore erros will be raised if I try to useit. For example:


```python
mg.start()
```

Will say that I am trying to start a machine that is already started.
If I try:
```python
mg.terminate()
```

It will raise an error because the machine no longer exists.

I guess this is expected because we are not constantly updating the info from the backend. But maybe we should have some time of check in `mg.start()` that will check if, in fact, the machine exists in the backend

IvanPombo:
> ## QA
> * `max_idle_time` seems to work fine. Started a machine with five minutes iddle time. After about 5/6 minutes the machine was gone ✅ ;
> * `auto_terminate_ts`. The machine persists after the time_stamp given. Code used:
> 
> ```python
>  mg.start(auto_terminate_ts=datetime.datetime(2024,3,25,15,7,tzinfo=pytz.timezone('Europe/Lisbon'))
> ```
> 
> ## Other behaviour to think about
> If, for example in python interpreter I use:
> 
> ```python
> mg.start(max_idle_time=datetime.timedelta(minutes=5))
> ```
> 
> After 5 minutes the machine is gone. Therefore erros will be raised if I try to useit. For example:
> 
> ```python
> mg.start()
> ```
> 
> Will say that I am trying to start a machine that is already started. If I try:
> 
> ```python
> mg.terminate()
> ```
> 
> It will raise an error because the machine no longer exists.
> 
> I guess this is expected because we are not constantly updating the info from the backend. But maybe we should have some time of check in `mg.start()` that will check if, in fact, the machine exists in the backend

Those are excellent points! For the `start()` method I think the key is the naming and the knowledge that the user can only start this configuration once. For the termination, I am not sure if any log of the machine remains in the backend. But we can discuss the interface further.

IvanPombo:
@AugustoPeres @Pbarbosa92 on my side it worked! Notice, that the machine won't be killed exactly when you set the date, that just provides a timestamp from when the janitor starts killing it.

AugustoPeres:
> @AugustoPeres @Pbarbosa92 on my side it worked! Notice, that the machine won't be killed exactly when you set the date, that just provides a timestamp from when the janitor starts killing it.

I will try again and give it some more time

AugustoPeres:
> > @AugustoPeres @Pbarbosa92 on my side it worked! Notice, that the machine won't be killed exactly when you set the date, that just provides a timestamp from when the janitor starts killing it.
> 
> I will try again and give it some more time

I cannot verify this on my side. The machines are not being shut down. Maybe I am doing something wrong. Here ist code I used to  start the machine:

mg.start(auto_terminate_ts=datetime.datetime(2024,3,25,15,50,tzinfo=pytz.timezone('Europe/Lisbon'))

At 15:50 the machine was not shutdown

AugustoPeres:
QA passed. Problem was with my time stamp

IvanPombo:
> QA passed. Problem was with my time stamp. I recommend using datetime.datetime(2024, 3, 25, 16, 10, tzinfo=datetime.timezone.utc) instead of datetime.datetime(2024,3,25,15,7,tzinfo=pytz.timezone('Europe/Lisbon')

@bgregoinductiva @Franreno do you know the timezone of reference for the janitor? Or what may be causing this?

IvanPombo:
Merging as it passed the QA, but further digging needs to be done for the timezone.

PR #1302, opened 2024-04-03
Get docs version from env var instead of git branch

tiagoafpereira:
Previous [method](9ccaeca81ffbd4def924e78901b38f658fcea956) of determining the docs version was a bit too clever for it's own good and did not produce the expected results ('unknown_version' was returned instead of the expected 'latest').

This PR simplifies the algorithm: it simply uses environment variables read from `docs/.env`.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -33,14 +33,14 @@ jobs:
           docs-folder: "source/docs/"
           # required by #1291
           pre-build-command: "apt-get update -y && apt-get install -y git"
-          build-command: "sphinx-build . ../../build/en/latest"
+          build-command: "sphinx-build . ../../build/en/latest -D version=latest"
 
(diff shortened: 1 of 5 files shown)


sergiomsantos (commented):
how do we differentiate between latest and development? do we need to change the env file all the time?

tiagoafpereira:
Now that you mention it I need to re-add .env to .gitignore otherwise there will be conflicts in the merge from main > development

tiagoafpereira:
@sergiomsantos found a better solution: leverage `sphinx-build` -D argument that enables overriding variables in `conf.py`.

My previous suggestion of setting a `.env` would not work because there would _always_ be conflicts when merging/rebasing like you've mentioned in your previous [comment](https://github.com/inductiva/inductiva/pull/1302#pullrequestreview-1979478399).

sergiomsantos (approved):
LGTM.
Thanks

PR #1338, opened 2024-04-19
Added ability to get user quotas

Pbarbosa92:
This PR adds the ability to get user quotas.

Using the cli
Captura de ecrã 2024-04-23, às 11 19 23

Using a python script
Captura de ecrã 2024-04-23, às 11 18 46


inductiva/users/methods.py
```diff
@@ -0,0 +1,39 @@
...
+    """Get information about a user's quotas.
+    """
+    api_config = api.validate_api_key(inductiva.api_key)
+
+    with ApiClient(api_config) as client:
+        api_instance = UsersApi(client)
+        try:
+            resp = api_instance.get_user_quotas().response
+
+            response_body = json.loads(resp.data.decode("utf-8"))
```
AugustoPeres, line 21: why not just `resp.body`?

inductiva/users/methods.py
```diff
@@ -0,0 +1,39 @@
...
+    """Get the user quotas.
+
+    This function gets a dict with the user quotas.
+
+    Returns:
+        Dict with the user quotas.
+    """
+    quotas = _fetch_quotas_from_api()
+
+    return quotas
```
AugustoPeres, line 39: This function is equivalent to:

```python
def get_quotas():
    return _fetch_quoatas_from_api()
```

Which is equivalent to `_fetch_quoatas_from_api`. Therefore why not just have all the code above here instead?

inductiva/users/methods.py
Pbarbosa92: the response has no `.body.
It has data that is just binary data.

inductiva/users/methods.py
Pbarbosa92: We get an array of dicts from the back-end `[{name:"num_machines"},{name:"vcpus"},...]`. Right now, we just return that.

But I figure that in the future, it would be better to convert the array of dicts into a dict of dicts (since it makes it easier for the end user to manipulate the data).
```
{
      "vcpu":{...}
      "num_machines":{...}
}
```
So, in the future, that conversion will be placed in the get_quotas function.

inductiva/_cli/cmd_user/__init__.py
sergiomsantos, line 1: I don't think this command should live inside the "users" supercommand.
Let's move it to `inductiva quotas list`

inductiva/_cli/cmd_user/quotas.py
```diff
@@ -0,0 +1,49 @@
...
+
+    header_formatters = [
+        lambda x: emph_formatter(x.upper(), format_utils.Emphasis.BOLD)
+    ]
+
+    print(format_utils.get_tabular_str(quotas,
+                                       headers=quotas.keys(),
+                                       header_formatters=header_formatters),
+          file=fout,
+          end="")
```
sergiomsantos, line 30:
```suggestion
    table = format_utils.get_tabular_str(quotas,
                                         headers=quotas.keys(),
                                         header_formatters=header_formatters)
    print(table,
          file=fout,
          end="")
```

inductiva/users/methods.py
sergiomsantos, line 21: if this fails, it will raise something other than an `ApiException` exception.

inductiva/users/methods.py
```diff
@@ -0,0 +1,39 @@
...
+
+            response_body = json.loads(resp.data.decode("utf-8"))
+
+            return response_body
+
+        except ApiException as e:
+            raise e
+
+
+def get_quotas() -> Dict[str, Dict[str, Any]]:
```
sergiomsantos, line 29: this method is basically a wrapper for `_fetch_quoatas_from_api`. Why not have a single `get_quotas` method?

inductiva/users/methods.py
```diff
@@ -0,0 +1,39 @@
...
+    api_config = api.validate_api_key(inductiva.api_key)
+
+    with ApiClient(api_config) as client:
+        api_instance = UsersApi(client)
+        try:
+            resp = api_instance.get_user_quotas().response
+
+            response_body = json.loads(resp.data.decode("utf-8"))
+
+            return response_body
```
sergiomsantos, line 23: let's convert this to a dict of dicts

sergiomsantos (changes requested):

inductiva/users/methods.py
Pbarbosa92: https://github.com/inductiva/inductiva/pull/1338#discussion_r1574556449
😄

inductiva/_cli/cmd_quotas/list.py
```diff
@@ -0,0 +1,56 @@
...
+    name_list = list(quotas)
+    max_allowed_list = [value["max_allowed"] for value in quotas.values()]
+    in_use_list = [value["in_use"] for value in quotas.values()]
+
+    # Construct the dictionary of lists
+    quotas_dict_of_lists = {
+        "name": name_list,
+        "max_allowed": max_allowed_list,
+        "in_use": in_use_list
+    }
```
sergiomsantos, line 26: whuy not use a defaultdict (`from collections import defaultdict`) to avoid multiple iterations over the dict?

```suggestion
    table = defaultdict(list)
    for name, quota in users.get_quotas().items():
        table["name"].append(name)
        table["in_use"].append(quota["in_use"])
        table["max_allowed"].append(quota["max_allowed"])
```

sergiomsantos (commented):
a minor suggestion

sergiomsantos (approved):
LGTM

Pbarbosa92:
@AugustoPeres Can you QA this? 

in the CLI it is inductiva quotas list
In a script we do inductiva.users.get_quotas()

AugustoPeres:
QA ✅ 

Screenshot 2024-04-26 at 10 40 51

PR #1344, opened 2024-04-23
adds the option to disable task logs

AugustoPeres:


inductiva/tasks/streams.py
```diff
@@ -109,6 +111,10 @@ def __on_message(self, ws: websocket.WebSocketApp, message: str):
                 return
             self._write_message(msg)
 
+    def _disable_logs(self):
+        tasks_api.TasksApi(api.get_client()).disable_task_logs(
+            {"task_id": self.task_id})
+
```
sergiomsantos, line 117:
```suggestion
    def _disable_logs(self):
        client = api.get_client()
        payload = {"task_id": self.task_id}
        tasks_api.TasksApi(client).disable_task_logs(payload)

```

sergiomsantos (changes requested):

inductiva/tasks/streams.py
```diff
@@ -268,6 +275,9 @@ def _run_websocket(self):
         and update the start query to get the logs from the last
         received timestamp."""
 
+        for sig in [signal.SIGTERM, signal.SIGINT, signal.SIGHUP]:
+            signal.signal(sig, self._handle_termination)
```
Pbarbosa92, line 279: Why don't we just do 
```
if self._end_of_stream or self._keyboard_interrupt:
      self._disable_logs()
      break
```
I don't think there is a way to capture a sigkill.
@sergiomsantos can give a better input on this.

inductiva/tasks/streams.py
```diff
@@ -109,6 +112,16 @@ def __on_message(self, ws: websocket.WebSocketApp, message: str):
                 return
             self._write_message(msg)
 
+    def _handle_termination(self, signum, _):
+        print(f"Received termination signal {signum}.")
+        self._disable_logs()
+        sys.exit()
```
sergiomsantos, line 118: let's remove this method.

inductiva/tasks/task.py
```diff
@@ -161,6 +161,10 @@ def get_info(self) -> Dict[str, Any]:
 
         return info
 
+    def disable_logs(self) -> None:
+        """Disables logs of the task."""
+        self._api.disable_task_logs({"task_id": self.id})
```
sergiomsantos, line 166: not used.

sergiomsantos (changes requested):
let's keep the initial version. If needed, we can iterate afterwards.

sergiomsantos (approved):

PR #1350, opened 2024-04-24
Fix remote tutorials path and create separate build and deploy actions

tiagoafpereira:
This PR:

* Fixes the rsync tutorials upload path
* Splits the previous build+deploy action separate build and deploy actions (same structure as we have for /docs)


sarmento (approved):
LGTM!

PR #1355, opened 2024-04-24
Simplifying deployment of tutorials to staging web server

hpenedones:
1. rsync directly the build/en/latest folder to the root of the destination folder, so that we don't need redirect files.
2. Because of 1, deleted the redirect.html file and the rules in the action
3. I am also not copying the robots.txt file, because in the stating environment we have a global robots.txt to avoid being indexed by search engines. But we can setup something for the prod deployment of tutorials, that will allow for crawling.
4. In the actions/checkout@v2 I removed the ref: main, so that the git checkout uses the git branch from which the action is being run from -- which is the more useful scenario for deploying something to the staging environment, that is not yet merged to main.


tiagoafpereira (approved):
LGTM!

PR #1365, opened 2024-04-29
Another round

sarmento:


tutorials/generating-synthetic-data/synthetic-data-generation-5.md
```diff
@@ -112,39 +85,33 @@ $ inductiva storage list -m 4
...
+| Particle Radius | Total N. of Particles | MB of data produced  |
 |-----------------|-----------------------|------------------------|
-| 0.01            | 15625                 | 166 MB                 |
-| 0.008           | 29791                 | 213 MB                 |
-| 0.006           | 73720                 | 525 MB                 |
-| 0.004           | 244584                | 1.76 GB                |
+| 0.01            | 15625                 | 166  |
+| 0.008           | 29791                 | 213  |
+| 0.006           | 73720                 | 525  |
+| 0.004           | 244584                | 1.76 |
```
hpenedones, line 101: Here, you moved the MB to the header of the table, but the last row was in GB, so we need to update the number to be 1000x bigger, so 1760. (Note: I think that when people use bases of 2 for storage units, i.e. 1024 factor instead of 1000, the notation is MiB instead of MB -- so here I am assuming the storage unit is using powers of 10 -- source: https://en.wikipedia.org/wiki/Byte#Multiple-byte_units)

tutorials/generating-synthetic-data/synthetic-data-generation-5.md
```diff
@@ -112,39 +85,33 @@ $ inductiva storage list -m 4
...
+
+For reference, the dataset produced by [Sanchez-Gonzalez et
+al.](https://arxiv.org/abs/2002.09405) was based on simulations with about 8k to 25k particles, so it seems that for creating a comparable dataset we don't need that particle radius to go below *0.008*. Also, if each simulation is producing hundreds of Mb of data, then we probably don't want to smaller particle radius anyway because we it would be challenging to store the data produced by *thousands* of simulation (let alone training the GNNs with all that data).
```
hpenedones, line 104: Mb -> MB (if referring to megabytes)

hpenedones, line 104: remove "**to**" from : we probably don't want **to** smaller 

Add comma before "because"

tutorials/generating-synthetic-data/synthetic-data-generation-3.md
```diff
@@ -134,56 +106,37 @@ task.download_outputs()
 my_machine_group.terminate()
 ```
 Notice the significant reduction in runtime for this simulation: down from 30m 
-to just **10m07s**, achieving an approximately 3-fold increase in speed!
+to just **10m07s**, achieving an approximately 3-fold increase in speed! Remember that, in practice, speed ups are not linear with the number of vCPUs, so this is alreasy quite an achievement with just a few lines of code.
```
hpenedones, line 109: *already

hpenedones (approved):

tutorials/generating-synthetic-data/synthetic-data-generation-5.md
sarmento: You are correct! Changing.

tutorials/generating-synthetic-data/synthetic-data-generation-5.md
sarmento: Done

PR #1366, opened 2024-04-29
Update function that uploads input files

luispcunha:
Previously, the backend accepted input files as `multipart/form-data`, but now interprets the request body as being the full input file. This PR updates the function that uses the generated client to upload input files. It also includes the latest generated client changes (#1364), which can't be merged due to conflicts.

### Note

The tests are failing because the backend endpoint hasn't been merged yet. This should be merged only after the merge in the backend happens. Follow the instructions in the next section to test the changes introduced in this PR.

### Testing it

Install the client package from this branch.
Create an input directory with a large file inside:
```bash
mkdir inputs
dd if=/dev/random of=inputs/large-file bs=1g count=5
```
Run a simulation using that directory as input dir, using the same credentials as for the dev environment but with the following url:
```bash
export INDUCTIVA_API_URL=https://inductiva-web-api-input-upload-endpoint-aodkovjrqa-ew.a.run.app
```

```python
import inductiva

commands = [
    "gmx solvate -cs tip4p -box 2.3 -o conf.gro -p topol.top",
]

gromacs = inductiva.simulators.GROMACS()

input_dir = "inputs"

task = gromacs.run(
    input_dir=input_dir,
    commands=commands,
)
task.wait()
 ```
The inputs upload should work correctly, even though the task fails due to the input files not including a valid gromacs input configuration.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -48,10 +48,16 @@ jobs:
       if: ${{ always() }}
       run: |
         pytest --suppress-no-test-exit-code
+
     - name: Test docs with pytest pytest-markdown-docs
       if: ${{ always() }}
       run: |
-        export INDUCTIVA_API_KEY=${{ secrets.API_TEST_KEY }}
+        if [ "${{ github.ref }}" = "refs/heads/main" ]; then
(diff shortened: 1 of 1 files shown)


sergiomsantos (approved):
LGTM

PR #1370, opened 2024-05-02
Tutorial typos and template fix

rcvalerio:
The tutorial was missing both `splishsplash-base-dir.zip` and `splishsplash-template-dir.zip`

I added `splishsplash-base-dir.zip` from the previous website and created `splishsplash-template-dir.zip` following the tutorial. However i had too change the file type from `config.json` to `config.json.jinja`. Is this correct? If so it should be mentioned in the tutorial.


hpenedones (approved):
This is great, thanks so much @rcvalerio for all these fixes!!!

rcvalerio:
added a minor fix to maintain consistency, now splishspash-template-dir.zip is inside a folder of the same name

hpenedones:
Cool, you can merge when you want :)

PR #1371, opened 2024-05-02
Add CaNS cpu simulator

Pbarbosa92:
Add CaNS Simulator.


.github/workflows/check_add_simulator.yaml
sergiomsantos, line 1: maybe we can keep this one outside this PR until a further decision is made

Pbarbosa92: Will do.

sergiomsantos (approved):

PR #1402, opened 2024-05-15
Refactor file downloads

bgregoinductiva:


inductiva/utils/data.py
```diff
@@ -229,6 +229,7 @@ def download_file(
     response: urllib3.response.HTTPResponse,
     output_path: pathlib.Path,
     chunk_size=1000,
+    download_size=None,
```
sergiomsantos, line 232: missing argument definition in docstring

inductiva/tasks/task.py
```diff
@@ -401,13 +410,66 @@ def download_outputs(
...
+                        for filename in filenames:
+                            try:
+                                remote_zip_file.extract(
+                                    filename,
+                                    path=output_dir,
+                                )
+                            except KeyError:
+                                logging.warning(
+                                    "File %s not found in the output archive.",
+                                    filename)
```
sergiomsantos, line 432: this has a lot of cyclomatic complexity and a bit har to read. Can the code be refactored (maybe move the try block to a separate method)?

inductiva/utils/data.py
bgregoinductiva: Done

inductiva/tasks/task.py
bgregoinductiva: Moved the block to `download_partial_outputs` function.

inductiva/tasks/task.py
```diff
@@ -385,9 +385,22 @@ def download_outputs(
         api_response = self._api.get_output_download_url(
             path_params=self._get_path_params(),)
 
-        download_url = api_response.body["url"]
+        download_url = api_response.body.get("url")
```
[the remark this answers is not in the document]
bgregoinductiva: Added a check to raise an exception if the download is not provided by the API.

inductiva/tasks/task.py
```diff
@@ -422,27 +434,42 @@ def download_outputs(
...
-            "GET",
-            download_url,
-            preload_content=False,
-        )
-        data.download_file(response, zip_path)
+        else:
+            api_response = self._api.download_task_output(
+                path_params=self._get_path_params(),
+                query_params={
+                    "filename": filenames or [],
```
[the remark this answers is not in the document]
bgregoinductiva: I removed the query parameters but maintained the conditional logic. We need to know if the download URL corresponds to a file server or to the Web API in order to use the API Key when making the request to the latter.

inductiva/utils/data.py
```diff
@@ -245,7 +246,8 @@ def download_file(
     # "x-content-length" is a custom header that is set by the API instead
     # of the standard "content-length" header, because the API needs to use
     # "transfer-encoding: chunked" to stream the response.
-    download_size = response.headers.get("x-content-length", 0)
+    if download_size is None:
+        download_size = response.headers.get("x-content-length", 0)
```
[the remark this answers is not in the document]
bgregoinductiva: We could, but the download_size is already obtained with `remote_filesystem.info()`.

inductiva/tasks/task.py
```diff
@@ -379,18 +382,24 @@ def download_outputs(
...
+            download_size = remote_filesystem.info(download_url).get("size")
+            logging.debug("Download size: %s", download_size)
+            file_server_available = True
+        except FileNotFoundError:
+            # If the Web API returns a fallback URL instead of a different
+            # file server (GCP, ICE, etc.) the remote_filesystem.info() will
+            # raise an exception. In this case, the output donwload will be
+            # provided by the Web API itself.
+            file_server_available = False
+            download_size = None
```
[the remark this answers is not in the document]
bgregoinductiva: I would rather maintain this logic for the reason mentioned above, determine if a file server is available to use the API Key when it's not.

inductiva/tasks/task.py
```diff
@@ -401,16 +414,56 @@ def download_outputs(
             warnings.warn("Path already exists, files may be overwritten.")
         output_dir.mkdir(parents=True, exist_ok=True)
 
+        download_message = "Downloading simulation outputs to %s"
+
+        if filenames:
```
hpenedones, line 419: Thanks Bruno, this function is becoming a bit too complex with lots of conditional logic, do you think you can simplify it a bit?

inductiva/tasks/task.py
```diff
@@ -379,18 +381,29 @@ def download_outputs(
...
-                "filename": filenames or [],
-            },
-            stream=True,
-            skip_deserialization=True,
-        )
-        # use raw urllib3 response instead of the generated client response, to
-        # implement our own download logic (with progress bar, first checking
-        # the size of the file, etc.)
-        response = api_response.response
+        api_response = self._api.get_output_download_url(
```
hpenedones, line 384: Thanks Bruno, this download_outputs function is becoming a bit too complex with lots of conditional logic, do you think you can simplify it a bit? I will leave a couple of suggestions...

inductiva/tasks/task.py
```diff
@@ -379,18 +381,29 @@ def download_outputs(
...
+
+        download_url = api_response.body.get("url")
+
+        if download_url is None:
+            raise RuntimeError(
+                "The API did not return a download URL for the task outputs.")
+
+        logging.debug("\nDownload URL: %s\n", download_url)
+
+        try:
```
hpenedones, line 395: suggestion: create small function to detect if file_server_is available

inductiva/tasks/task.py
```diff
@@ -401,16 +414,56 @@ def download_outputs(
             warnings.warn("Path already exists, files may be overwritten.")
         output_dir.mkdir(parents=True, exist_ok=True)
 
+        download_message = "Downloading simulation outputs to %s"
```
hpenedones, line 417: on line 408, there is no need to make the same call on each branch of the if statement. You can do:

```python
if output_dir is None:
   output_dir = self.id

output_dir = files.resolve_output_path(output_dir)
```

It could even just be (but I don't necessarily recommend it as it might be harder to read):

```python
output_dir = files.resolve_output_path(output_dir or self.id)
```

inductiva/tasks/task.py
```diff
@@ -401,16 +414,56 @@ def download_outputs(
...
+                logging.error("Partial download is not available.")
+
+            # If the user requested a partial download, the full download
+            # will be skipped.
+            return output_dir
+
         zip_path = output_dir.joinpath("output.zip")
+        logging.info(download_message, zip_path)
+
+        if file_server_available:
```
hpenedones, line 438: can you maybe just have one if file_server_available condition, rather than here and above?

hpenedones (commented):
Thanks Bruno, left some suggestions to reduce Cyclomatic complexity ...

inductiva/tasks/task.py
bgregoinductiva: Sure, thanks for the suggestions.

inductiva/tasks/task.py
bgregoinductiva: We moved this logic to the backend since it is also needed for uploads.

https://github.com/inductiva/inductiva-web-api/pull/852

inductiva/tasks/task.py
hpenedones: Thanks!

inductiva/tasks/task.py
hpenedones: Ok, thanks for checking

sergiomsantos (approved):
LGTM. Thanks

PR #1420, opened 2024-05-26
Adding a more advanced example to Reed3D docs

sarmento:


docs/simulators/Reef3D.md
```diff
@@ -70,6 +76,254 @@ task.wait()
...
+task.download_outputs()
+
+## Make sure we terminate the machine.
+machine_group.terminate()
+```
+
+Running this script end-to-end should take about 25 minutes, distributed in
+the following way:
+
+* about 2 minutes in the preparation stage, including starting the executer VM.
```
sarmento, line 189: @sergiomsantos @lneves-ai @luispcunha : for now I am informally collecting these number. It would be great to have these number explicit in the stdout of my python script. These stages should be more explicit so that the user  (and ourselves) can understand more exactly where the time is being spent, and then make the necessary changes.

(cc @hpenedones )

hpenedones (approved):

PR #1426, opened 2024-05-27
Add Project documentation

sergiomsantos:
This PR adds documentation for the `Project` API.

### DIFF
--- a/tutorials/intro_to_api/projects.md
+++ b/tutorials/intro_to_api/projects.md
@@ -2,7 +2,7 @@
 
 Projects are the primary way to organize `Tasks` in the Inductiva API.
 A project is a collection of tasks that run under a common umbrella, identified
-by its name. Any submitted task will always belong to a project, either a
+by its name. Any submitted task will always belong to a project: either a
 user-specified one or the default project. This allows users to organize their
 tasks in a way that makes sense to them.
 The following sections will explain how to create and interact with projects.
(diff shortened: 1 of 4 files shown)


tutorials/intro_to_api/projects.md
```diff
@@ -0,0 +1,258 @@
...
+
+# task2 will be added to the default project
+task2 = simulator.run(input_dir=input_dir,
+                      sim_config_filename="params.txt")
+
+print(task1.get_info()['project']) # "my_xbeach_project"
+print(task2.get_info()['project']) # "userab1cdef2" (default project)
+```
+
+#### Explicit management
```
sarmento, line 129: Maybe we should start with the explicit option.

sarmento (commented):
This looks great!
Thank you!
I wonder if we really need the append=True. Should that be actually a permanent state? Why? Why not?
Can you pls clarify that somewhere?

sarmento (commented):
LGTM!

tutorials/intro_to_api/projects.md
sergiomsantos: Done in the development branch.

sergiomsantos:
> This looks great! Thank you! I wonder if we really need the append=True. Should that be actually a permanent state? Why? Why not? Can you pls clarify that somewhere?

The only reason to have this behaviour is so that the user is aware that tasks will be appended to a project if a project is open. It doesn't need to be that way, but at least it provides some sort of security check to ensure that the user is aware of what is going on. This is explained in the `Adding a Task to a Project` section, right at the top.
We can change this behaviour, though.

PR #1433, opened 2024-05-31
Improve get output info

luispcunha:
Closes inductiva/tasks#200

### DIFF
--- a/inductiva/tests/tasks/test_task.py
+++ b/inductiva/tests/tasks/test_task.py
@@ -180,5 +180,5 @@ def test__get_output_info():
     output_info = task.get_output_info()
 
     assert output_info.n_files == 2
-    assert output_info.total_size == 320
-    assert output_info.total_compressed_size == 150
+    assert output_info.total_size_bytes == 320
+    assert output_info.total_compressed_size_bytes == 150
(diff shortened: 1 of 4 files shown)


inductiva/utils/output_info.py
```diff
@@ -14,26 +14,33 @@ class FileInfo:
...
+
+    @property
+    def n_files(self) -> int:
+        return len(self.files)
 
     @property
-    def num_files(self) -> int:
-        return len(self.contents)
+    def total_compressed_size(self) -> int:
+        return sum(file.compressed_size for file in self.files)
```
sergiomsantos, line 28: these are bytes?
maybe change the name of the method to reflect the units:

```suggestion
        def total_compressed_size_bytes(self):
```

inductiva/tests/tasks/test_task.py
```diff
@@ -145,3 +147,38 @@ def test__check_if_pending_kill__wait_timeout_positive__success_status(
...
+    """
+    task = inductiva.tasks.Task("123")
+    # pylint: disable=W0212
+    mock_resp_body = output_list_get.SchemaFor200ResponseBodyApplicationJson(
+        size=320,
+        contents=[
+            {
+                "name": "file1.txt",
+                "size": 100,
+                "compressed_size": 50
```
sergiomsantos, line 164: what are the units for these fields?

sergiomsantos (commented):
I have some suggestions about the name/location of some classes and methods;
I think we should be informative about the units we're using.

sergiomsantos (approved):
LGTM, thanks!

PR #1450, opened 2024-06-07
Update inductiva version in README.md

rcvalerio:


hpenedones (commented):
Thanks Rodrigo, I think that now that we have the badge with the pypi version dynamically updated, probably we could even remove the version number from this paragraph. This way is always up to date.

hpenedones (approved):

PR #1459, opened 2024-06-12
State how many tasks are ahead in the queue

Franreno:
This PR aims to provide more information to the user in the way of stating how many tasks are ahead in the queue for the task's machine_group.

When submiting one task a new logging statment will show up:

```diff
Local input directory successfully uploaded.
Task buqz7ko70spkmgn6hmc2epahy submitted to the default queue.
+ There are 0 tasks ahead of this task in the queue.
```
Or 

```diff
Local input directory successfully uploaded.
Task p1bk70jr85uvqg1k7hbuhv8s9 submitted to the default queue.
+ There are 1 tasks ahead of this task in the queue.
```

Closes [#203](https://github.com/inductiva/tasks/issues/203)


inductiva/tasks/task.py
```diff
@@ -137,6 +137,19 @@ def get_status(self) -> models.TaskStatusCode:
...
+
+        This method issues a request to the API.
+        """
+        try:
+            resp = self._api.get_task_position_in_queue(self._get_path_params())
+            tasks_ahead = resp.body.get("tasks_ahead", None)
+            return tasks_ahead
+        except exceptions.ApiException as exc:
+            if exc.status == 404:
+                return None
```
Pbarbosa92, line 151: This should behave in the same manner as task.get_info() and task.info
task.get_position_in_queue() can do this and task.get_position_in_queue returns the value stored in memory.

Does this work in any kind of resource or only for tasks that run on the default queue ?

inductiva/tasks/run_simulation.py
```diff
@@ -65,6 +65,11 @@ def run_simulation(
         raise RuntimeError(
             f"Expected result to be a string with task_id, got {type(task_id)}")
 
+    position = task.get_position_in_queue()
+    if position is not None:
+        logging.info("There are %s tasks ahead of this task in the queue.",
```
hpenedones, line 70: Minor detail: the plural does not quite work for a single task... it should be "there is 1 task" and not "there are 1 tasks"... If you want to avoid an if-statement just for this, maybe say "Number of tasks ahead in the queue: %s"... (or something along those lines)

But this is great feedback for the user, thanks Francisco!

inductiva/tasks/task.py
Pbarbosa92: I've talked with @sergiomsantos this is not live yet but the comment might still be relevant.

Franreno: > returns the value stored in memory

Okay! I will update the logic to store it in memory.

> Does this work in any kind of resource or only for tasks that run on the default queue ?

This work for every task that have a machine group, so I think it works for everything :)

Pbarbosa92: > @Pbarbosa92, please let me know if [a8e96d0](https://github.com/inductiva/inductiva/pull/1459/commits/a8e96d036937a127fd095ddaf6f1920836970ce8) is sufficient.

Now you can create 
```python
@property
def position_in_queue
```
That just returns the position in queue without any call to the api

Franreno: Hmmmm gotcha!
But since this isn't live for the other properties, should we wait for them to land first?

Pbarbosa92: Yes. We can refactor the task later. And make all the changes at the same time.

Franreno:
> can we update the logs with newest position in queue in the task wait method?

That's a great ideia. Since the status endpoint already returns information from the queue we can ue that to update the queue on the wait method. 

The video below is an example of this behavior. There are 10 tasks on the queue, but I'm just using the task.wait() on the last last submitted. 

https://github.com/inductiva/inductiva/assets/67326251/7dedf03a-158e-4460-b51f-b2432594ac19


Note: Video is speedup

Pbarbosa92 (approved):

inductiva/tasks/run_simulation.py
```diff
@@ -65,6 +65,11 @@ def run_simulation(
         raise RuntimeError(
             f"Expected result to be a string with task_id, got {type(task_id)}")
 
+    position = task.get_position_in_queue()
+    if position is not None:
```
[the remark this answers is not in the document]
sergiomsantos: when would the task not have queue information?
when it is submitted, it is either picked immediately or needs to wait in the queue, correct?

inductiva/tasks/task.py
```diff
@@ -216,6 +236,11 @@ def wait(self, polling_period: int = 5) -> models.TaskStatusCode:
                         "while performing the task.", status)
             prev_status = status
 
+            if (self._tasks_ahead is not None and
+                    self._tasks_ahead != prev_tasks_ahead):
+                logging.info("Number of tasks ahead of task %s: %s", self.id,
+                             self._tasks_ahead)
```
sergiomsantos, line 242: this is a bit verbose and, in my opinion, will pollute stdout quite a lot, specially if it has to wait several minutes in the queue. In the worst case, and if the task only moves  by one slot every 5 seconds, then I foresee several of these lines being printed. If this was a dynamic print, then that problem would be solved because the line would be re-written every time. But this is using the logging mechanism!

However, we can print directly to stdout and have that feedback given to the user. Something like:

```python
while ...:
    printf("\33[2K\r", file=sys.stdout) <-- because numbers decrease and we want the buffer to be cleared out
    print("Number of tasks ahead of task %s: %s", self.id, self._tasks_ahead, file=...)
sys.stdout.flush() <-- make sure logs are flushed
printf("\33[2K\r") <-- a final cleanup
```

sergiomsantos (changes requested):
The feedback is very verbose and prone to polluting the terminal screen.
I suggest a slightly more user-friendly version.

sergiomsantos (approved):
Looks great. Thanks!

PR #1466, opened 2024-06-13
inductiva logs supports --stdout and --stderr

Pbarbosa92:
This PR introduces the ability to filter log outputs with the following commands:

- `inductiva logs :id: --stdout`: Displays only the standard output (stdout) from the logs.
- `inductiva logs :id: --stderr`: Displays only the standard error (stderr) from the logs.
- `inductiva logs :id:`: Displays both stdout and stderr from the logs, with stderr messages highlighted in red.

### How to Test

1. **OpenFOAM Simulation:**
   - Run the OpenFOAM simulation as described in the [documentation](https://docs.inductiva.ai/en/latest/simulators/OpenFOAM.html).
   - Use the `--stdout` option to see the logs normally.
   - Use the `--stderr` option to verify that no errors appear (assuming there are no errors in the simulation).

2. **FDS Simulation:**
   - Run the FDS simulation as outlined in the [documentation](https://docs.inductiva.ai/en/latest/simulators/FDS.html).
   - Note that this simulator writes all output to stderr.
     - Running `inductiva logs :id:` will display all text in red since it comes from stderr.
     - Using `--stdout` will result in no output.
     - Using `--stderr` will display the logs normally.


Pbarbosa92:
> For simulations such as the `FDS` one, in which everything goes to stderr, it might be too abrasive showing the whole outputs in red, no? My suggestion would be either to add a flag to disable colouring the logs, or instead of identifying stdout and stderr via the whole line being coloured, we could prepend a colored character to the beginning of each line?

It only shows red if you are consuming both streams. If you are using --stderr it does not show in red

inductiva/_cli/cmd_logs/logs.py
```diff
@@ -7,6 +7,15 @@
 
 def stream_task_logs(args):
     """Consume the stream logs of a certain task."""
+
+    if args.stdout and args.stderr:
+        print(
+            "inductiva logs: error: "
+            "argument --stdout not allowed with argument --stderr",
+            file=sys.stderr)
+        return 1
```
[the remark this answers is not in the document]
Pbarbosa92: Will do that

Pbarbosa92:
@luispcunha I've done what you asked. You can now use --no-color to disable colors and you can also use stdout and stderr to have the default behaviour.

Pbarbosa92:
Closes https://github.com/inductiva/tasks/issues/194

inductiva/tasks/streams.py
```diff
@@ -60,18 +60,23 @@ class TaskStreamConsumer:
...
     END_OF_STREAM = "<<end_of_stream>>"
+    ALLOWED_IO_STREAMS = ["std_out", "std_err"]
 
     def __init__(self,
                  task_id: str,
                  fout: IO = sys.stdout,
-                 ferr: IO = sys.stderr):
+                 ferr: IO = sys.stderr,
+                 io_stream: str = None,
+                 no_color: bool = False):
```
Franreno, line 70: Should add the `no_color` on the docstring as well

inductiva/tasks/streams.py
```diff
@@ -136,9 +151,13 @@ def _ansi_message_formatter(self, msg):
...
         # 7. add new blank line
-        return f"{UP}\r{CLEAR_LINE}{msg}{DOWN}\r{CLEAR_LINE}\n"
+        if is_stderr:
+            m = f"{UP}\r{CLEAR_LINE}{FAILURE}{msg}{RESET}{DOWN}\r{CLEAR_LINE}\n"
+        else:
+            m = f"{UP}\r{CLEAR_LINE}{msg}{DOWN}\r{CLEAR_LINE}\n"
+        return m
 
-    def _plain_message_formatter(self, msg):
+    def _plain_message_formatter(self, msg, **_):
```
Franreno, line 160: Could you add a simple comment to explain why the **_ ?

Franreno (commented):
Overall LGTM. 

We need to sync to decide when we should merge because of the PR #1471, because they make changes on the same file. Maybe yours can land first :)

inductiva/tasks/streams.py
Pbarbosa92: Done

inductiva/tasks/streams.py
Pbarbosa92: "# _ to make it have the same number or arguments as the other formatters
 # In this case, the second argument is not used."

Done. Is this clear ?

Franreno: Yes! Thanks :)

Franreno (approved):

Pbarbosa92:
@Franreno can you do a mini QA? There are instructions in the PR ** How to Test**

Franreno:
@Pbarbosa92 I did a QA here and found some problems. But first, for the openFoam it worked correctly! 

However, for the fds (which outputs to sterr) failed in the print parts:

On VSCode terminal:
Screenshot 2024-06-17 at 11 19 19

Screenshot 2024-06-17 at 11 19 42

On zsh terminal:
Screenshot 2024-06-17 at 11 20 54

The `Streaming output from task` seems to be repeating for me.

Franreno:
**QA**

Logs output seems to be working correctly now:
Screenshot 2024-06-17 at 17 31 14


Previous tests are still working!

PR #1471, opened 2024-06-14
[CLI] Defaulting logs behavior

Franreno:
## Summary

This PR aims to revamp the behavior of the `inductiva logs` cli command. Previously the command needed to have a `task_id` to work, instead we could use some default behavior such as `inductiva logs`, without any other arguments, default to the last submitted task. 

This is the new specification for the logs comand:

`inductiva logs SUBMITTED[-n] | STARTED[-n] | task_id`

If none is provided, then it defaults to the submitted behavior. 

Closes [#195](https://github.com/inductiva/tasks/issues/195)

## Examples

Without tasks running it tries to get the last task submmited:
➜  inductiva logs            
The current status of task caze0p7xn9ncp2tm5qpot7c6t is 'success'
and the simulation logs are not available for streaming.

Getting the second last task submitted:
➜  inductiva logs submitted-1
The current status of task bpnefgzsahdrynf2adowmchpf is 'success'
and the simulation logs are not available for streaming.

Without started task running it returns that there are no tasks with `started`
➜  inductiva logs started                        
No 'started' tasks found.

Getting logs for a specific task_id
➜  inductiva logs cv0p5hpbppb4herns44ods0s1                         
The current status of task cv0p5hpbppb4herns44ods0s1 is 'killed'
and the simulation logs are not available for streaming.

With a task running:
➜  inductiva logs started
 ● Opening connection to task ioa8vjbpa8i6339zda14ft2eu...

With an invalid task_id format:
➜  inductiva logs loremip 
Invalid task_id: loremip

With a valid task_id format but does not exist:
➜  inductiva logs aaaaaaaaaaaaaaaaaaaaaaaaa
Task ID not present in the list of tasks owned by the user.

### DIFF
--- a/inductiva/_cli/cmd_logs/logs.py
+++ b/inductiva/_cli/cmd_logs/logs.py
@@ -49,13 +49,14 @@ def _check_if_task_is_running(task: tasks.Task) -> Tuple[bool, str]:
 
     is_running = False
     try:
-        is_running = task.is_running()
+        info = task.get_info()
+        is_running = info.is_running
     except ApiException as e:
         return False, f"{ json.loads(e.body)['detail']}"
 
     if not is_running:
         return False, (
-            f"The current status of task {task.id} is '{task.get_status()}'\n"
+            f"The current status of task {task.id} is '{info.status}'\n"
             "and the simulation logs are not available for streaming.\n"
             "For more information about the task status, use:\n\n"
             f"  inductiva tasks list --task-id {task.id}\n")
--- a/inductiva/_cli/cmd_logs/logs.py
+++ b/inductiva/_cli/cmd_logs/logs.py
@@ -13,9 +13,16 @@ def _get_task_id_from_mode(mode: str) -> Tuple[bool, str]:
     if not match:
         return False, f"Invalid mode format: {mode}"
 
-    status = match.group(1).lower()
-    offset = int(match.group(3)) if match.group(3) else 0
-    offset += 1
+    status_match = match.group(1)
+    offset_match = match.group(3)
+
+    status = status_match
+    offset = (int(offset_match) + 1) if offset_match else 1
+
+    # If the status is submitted, send None to the API to get all the tasks.
+    # The endpoint already returns the tasks in the submitted order by default
+    if status == "submitted":
+        status = None
 
     task_list = tasks.get(last_n=offset, status=status)
 
--- a/inductiva/_cli/cmd_logs/logs.py
+++ b/inductiva/_cli/cmd_logs/logs.py
@@ -1,15 +1,16 @@
 """CLI for logs."""
+import json
 import re
 import sys
 from typing import Tuple
 
 from .. import utils as cli_utils
-from ... import tasks
+from ... import tasks, ApiException
 
 
 def _get_task_id_from_mode(mode: str) -> Tuple[bool, str]:
     """Extract the task_id from the mode."""
-    match = re.match(r'^(submitted|started)(-(\d+))?$', mode)
+    match = re.match(r"^(submitted|started)(-(\d+))?$", mode)
     if not match:
         return False, f"Invalid mode format: {mode}"
 
@@ -43,6 +44,29 @@ def validate_mode_or_task_id(mode: str) -> Tuple[bool, str]:
     return True, mode
 
 
+def _check_if_task_is_running(task: tasks.Task) -> Tuple[bool, str]:
+    """Check if the task is running.
+    
+    This is done inside a try-except block to catch any exceptions that may
+    occur when trying to get the task status. Mainly if the task does not exist.
+    """
+
+    is_running = False
+    try:
+        is_running = task.is_running()
+    except ApiException as e:
+        return False, f"{ json.loads(e.body)['detail']}"
+
+    if not is_running:
+        return False, (
+            f"The current status of task {task.id} is '{task.get_status()}'\n"
+            "and the simulation logs are not available for streaming.\n"
+            "For more information about the task status, use:\n\n"
+            f"  inductiva tasks list --task-id {task.id}\n")
+
+    return True, task.id
+
+
 def stream_task_logs(args):
     """Consume the stream logs of a certain task."""
     mode = args.mode.lower()
@@ -54,13 +78,9 @@ def stream_task_logs(args):
     task_id = data
     task = tasks.Task(data)
 
-    if not task.is_running():
-        print(
-            f"The current status of task {task_id} is '{task.get_status()}'\n"
-            "and the simulation logs are not available for streaming.\n"
-            "For more information about the task status, use:\n\n"
-            f"  inductiva tasks list --task-id {task_id}\n",
-            file=sys.stderr)
+    result, data = _check_if_task_is_running(task)
+    if not result:
+        print(data, file=sys.stderr)
         return 1
 
     consumer = tasks.streams.TaskStreamConsumer(task_id)
@@ -74,7 +94,7 @@ def register(parser):
     parser.add_argument(
         "mode",
         type=str,
-        nargs='?',
+        nargs="?",
         default="SUBMITTED",
         help=(
             "Mode of log retrieval. "
--- a/inductiva/_cli/cmd_logs/logs.py
+++ b/inductiva/_cli/cmd_logs/logs.py
@@ -14,10 +14,9 @@ def _get_task_id_from_mode(mode: str) -> Tuple[bool, str]:
     if not match:
         return False, f"Invalid mode format: {mode}"
 
-    status_match = match.group(1)
     offset_match = match.group(3)
 
-    status = status_match
+    status = match.group(1)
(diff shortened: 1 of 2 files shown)


inductiva/_cli/cmd_logs/logs.py
```diff
@@ -27,9 +91,16 @@ def stream_task_logs(args):
...
+    parser.add_argument(
+        "mode",
+        type=str,
+        nargs="?",
+        default="SUBMITTED",
+        help=(
+            "Mode of log retrieval. "
+            "Use 'SUBMITTED[-n]' for the last submitted tasks, "
+            "'STARTED[-n]' for the last started tasks, "
+            "or a specific 'task_id' to retrieve logs for a particular task."))
```
Pbarbosa92, line 103: I'm a bit confused here. Is submitted the same as submitted-1 ? Do they both represent the last submitted task ?

inductiva/_cli/cmd_logs/logs.py
```diff
@@ -1,22 +1,86 @@
...
+
+    if not task_list:
+        return False, f"No '{status}' tasks found."
+
+    return True, task_list[-1].id
+
+
+def validate_mode_or_task_id(mode: str) -> Tuple[bool, str]:
+    """Validate the mode or task_id."""
+    if mode.startswith("submitted") or mode.startswith("started"):
```
Pbarbosa92, line 38: Isn't this checked at `_get_task_id_from_mode` during the regular expression? If the str does not start with submitted or started the regex will fail.

inductiva/_cli/cmd_logs/logs.py
```diff
@@ -1,22 +1,86 @@
...
+        return _get_task_id_from_mode(mode)
+
+    if not cli_utils.is_task_id_valid(mode):
+        return False, f"Invalid task_id: {mode}"
```
Pbarbosa92, line 42: Why do we do this? If we use an invalid task_id wont the api fail to fetch the task?

inductiva/_cli/cmd_logs/logs.py
Franreno: No. 
Submitted is the last submitted task
Submitted-1 is the second last submitted task.

This is similar behavior to git's one:
git show HEAD shows the last log
git show HEAD~1 shows the second last log

inductiva/_cli/cmd_logs/logs.py
Franreno: Hmmm yes.. they are checked again in the `_get_task_id_from_mode`, the regex checks for the -n parameter as well. Therefore the -n parameter only exists if the `mode` starts with submitted or started. 

If it does not starts with these two strings it wont even enter the `_get_task_id_from_mode` function and it is assumed that the mode is a task_id.

inductiva/_cli/cmd_logs/logs.py
Franreno: To better represent the failure of the error. Since everything else then submitted and started is treated as a task_id, if the user writes `inductiva logs success` its better to show him that in this case `success` is an _invalid task_id_ instead of showing that the task_id `success` _does not exists_.

inductiva/_cli/cmd_logs/logs.py
Pbarbosa92: Then we need a better explanation here. Maybe add some examples in the docs.

inductiva/_cli/cmd_logs/logs.py
```diff
@@ -1,22 +1,86 @@
...
-    """Consume the stream logs of a certain task."""
-    task_id = args.task_id
-    task = tasks.Task(task_id)
+def _get_task_id_from_mode(mode: str) -> Tuple[bool, str]:
+    """Extract the task_id from the mode."""
+    match = re.match(r"^(submitted|started)(-(\d+))?$", mode)
+    if not match:
+        return False, f"Invalid mode format: {mode}"
+
+    status_match = match.group(1)
```
bgregoinductiva, line 17: I think this variable could be just `status`.

inductiva/_cli/cmd_logs/logs.py
```diff
@@ -1,22 +1,86 @@
...
+
+    is_running = False
+    try:
+        is_running = task.is_running()
+    except ApiException as e:
+        return False, f"{ json.loads(e.body)['detail']}"
+
+    if not is_running:
+        return False, (
+            f"The current status of task {task.id} is '{task.get_status()}'\n"
```
bgregoinductiva, line 62: Occasionally, there might be a mismatch between `task.is_running()` and `task.get_status()` because these methods make 2 independent requests to the backend. I suggest you use the information returned by `task.get_info()`, but in that case it would be better to wait for #1456 to be merged.

bgregoinductiva (commented):
LGTM overall. Left 2 improvement suggestions.

inductiva/_cli/cmd_logs/logs.py
Franreno: Yes I agree! What would you suggest to improve this explanation?
What do you think about this:
```python
help=(
        "Mode of log retrieval. "
        "Use 'SUBMITTED' for the last submitted task, "
        "'SUBMITTED-1' for the second last submitted task, and so on. "
        "'STARTED' and 'STARTED-n' follow the same pattern for started tasks. "
        "Or, use a specific 'task_id' to retrieve logs for a particular task."
    )
```

inductiva/_cli/cmd_logs/logs.py
Franreno: I will remove this validation here and open a ticket for the backend

inductiva/_cli/cmd_logs/logs.py
Franreno: Ops, yeah nice catch!
thanks

inductiva/_cli/cmd_logs/logs.py
Pbarbosa92: LGTM!

Pbarbosa92 (approved):
LGTM

inductiva/_cli/cmd_logs/logs.py
bgregoinductiva: Now that #1456 is merged, I strongly suggest the use of `task.get_info()`. It avoids the possible mismatch between `task.is_running()` and `task.get_status()`, while it also reduces the number of API requests to 1.

```
    info = task.get_info()

    if not task.info.is_running:
        return False, (
            f"The current status of task {task.id} is '{task.info.status}'\n"
```

bgregoinductiva (changes requested):

bgregoinductiva:
I did some QA and the functionalities are correct.

```
(env) bruno@Brunos-Laptop temp % inductiva logs
The current status of task r5mcww3zxjsf33e1jq97qer6i is 'submitted'
and the simulation logs are not available for streaming.
For more information about the task status, use:

  inductiva tasks list --task-id r5mcww3zxjsf33e1jq97qer6i

(env) bruno@Brunos-Laptop temp % inductiva logs submitted-1
The current status of task 2irryufmtz9vqtz295ugg53p5 is 'success'
and the simulation logs are not available for streaming.
For more information about the task status, use:

  inductiva tasks list --task-id 2irryufmtz9vqtz295ugg53p5

(env) bruno@Brunos-Laptop temp % inductiva logs started
No 'started' tasks found.

(env) bruno@Brunos-Laptop temp % inductiva logs            

 ● Opening connection to task r5mcww3zxjsf33e1jq97qer6i...

(env) bruno@Brunos-Laptop temp % inductiva logs 2irryufmtz9vqtz295ugg53p5
The current status of task 2irryufmtz9vqtz295ugg53p5 is 'success'
and the simulation logs are not available for streaming.
For more information about the task status, use:

  inductiva tasks list --task-id 2irryufmtz9vqtz295ugg53p5

(env) bruno@Brunos-Laptop temp % inductiva logs loremip
Invalid task ID format.

(env) bruno@Brunos-Laptop temp % inductiva logs aaaaaaaaaaaaaaaaaaaaaaaaa
Task ID not present in the list of tasks owned by the user.

(env) bruno@Brunos-Laptop temp % inductiva logs started                  
 Time Step:      50, Simulation Time:      1.07 s
 Time Step:      60, Simulation Time:      1.14 s
 Time Step:      70, Simulation Time:      1.20 s
 Time Step:      80, Simulation Time:      1.26 s
 Time Step:      90, Simulation Time:      1.32 s
 Time Step:     100, Simulation Time:      1.39 s
 Time Step:     200, Simulation Time:      1.99 s
Time Step:     300, Simulation Time:      2.57 s

 ● 0:00:41 Streaming output from task fdpi69ymebtgfik3otsdrrdmt
```

inductiva/_cli/cmd_logs/logs.py
Franreno: Thanks Bruno! 
I've updated the code to use the get_info method!

PR #1472, opened 2024-06-14
Change quota table headers

luispcunha:


Pbarbosa92 (approved):
I will miss New Total :(

Issue #215, opened 2023-04-14
Configuration issues to render videos on Google Colab


IvanPombo:
At this moment in time, to run a video on Google Colab we need to do two things:

- Do the following installation:  `apt-get install libgl1-mesa-glx xvfb`
- Start a virtual display: `pyvista.start_xvfb()`

While for the first step there is not at this point a way to circumvent it, unless we move rendering to the server side, the second one is circumvented by adding the command in the rendering function on the client side. 

For this to function both locally and in Colab, we need to pass a bool argument  denoted as `virtual_display` to create a virtual screen. For Colab we need to set it to True and for local dev we set it to False.

This is very bulky and needs to be addressed.

PR #1474, opened 2024-06-17
Improved resources terminate and task kill messages

rcvalerio:
### Terminate message
- The message for the --all command has been updated to display the machines that will be removed (or the number of machines if there are more than 5).

Current output:
```
$ inductiva resources terminate --all
You are about to terminate ALL resources.
Are you sure you want to proceed (y/[N])? y
...
```

Output with the new changes:
```
$ inductiva resources terminate --all                        
You are about to terminate the following resources:
  - api-cgxaxawfstran71xszgserpb2
  - api-cd0imcswv8f46mf8eqvuh9whk
  - api-8u4z3ypf09dpd7krm2893t271
  - api-cef4169p84mdoe0bm819rj1x2
  - api-4wo9mr6bef16ofeaqux51oqa5
Are you sure you want to proceed (y/[N])? y
```

The same thing also for tasks

### DIFF
--- a/inductiva/_cli/cmd_tasks/kill.py
+++ b/inductiva/_cli/cmd_tasks/kill.py
@@ -34,13 +34,17 @@ def kill_task(args):
                 "task_id": task.id
             } for task in get_all(status=status)])
 
+    if not tasks:
+        print("There are not tasks to kill.")
+        return 1
+
     ids = [task["task_id"] for task in tasks]
 
     confirm = args.yes or \
         user_confirmation_prompt(ids,
                                 __("task-prompt-kill-all"),
                                 __("task-prompt-kill-big", len(ids)),
-                                __("task-prompt-kill-small"), kill_all
+                                __("task-prompt-kill-small"), False
                                 )
 
     if not confirm:
--- a/inductiva/_cli/cmd_tasks/kill.py
+++ b/inductiva/_cli/cmd_tasks/kill.py
@@ -35,7 +35,7 @@ def kill_task(args):
             } for task in get_all(status=status)])
 
     if not tasks:
-        print("There are not tasks to kill.")
+        print("There are no tasks to kill.")
         return 1
 
     ids = [task["task_id"] for task in tasks]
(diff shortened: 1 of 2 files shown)


Pbarbosa92 (approved):
LGTM!

rcvalerio:
This pull request can be merged but ideally we would also change the message of the `storage remove` cmd in the client. Please see #215

Pbarbosa92:
> This pull request can be merged but ideally we would also change the message of the `storage remove` cmd in the client. Please see [#215](https://github.com/inductiva/tasks/issues/215)

I believe we should do this in this PR. Just to keep the code consistent.

inductiva/_cli/cmd_tasks/kill.py
```diff
@@ -34,13 +34,17 @@ def kill_task(args):
                 "task_id": task.id
             } for task in get_all(status=status)])
 
+    if not tasks:
+        print("There are not tasks to kill.")
+        return 1
```
hpenedones, line 39: Could we please document what is the meaning of each return code ? @Pbarbosa92 @rcvalerio

Pbarbosa92: > Could we please document what is the meaning of each return code ? @Pbarbosa92 @rcvalerio

There is no specific meaning. If it all goes well we return 0. Any kind of error we return non 0 (1 in this case) but there is no error codes at this time.

rcvalerio: I believe the return codes are from `argParse` and they are common in all our client commands.

`0`: Indicates that the program executed successfully without errors. This is the standard convention for success in Unix-like operating systems.
`1`: Indicates that the program encountered an error. This is a common convention for general errors, but specific programs might use other non-zero values to signify different types of errors.

hpenedones: Ok, I see, thanks -- this integer return codes are being used exclusively on the CLI (command line interface), which makes sense because it's returning to the shell:
https://github.com/search?q=repo%3Ainductiva%2Finductiva%20%22return%201%22&type=code

In the Python package logic, functions and methods are not using that pattern (correctly), because in Python you would typically return some richer type, or potentially raise an exception if an error occurs.

Cool, thanks for the clarification.

hpenedones (approved):

PR #1487, opened 2024-06-19
181 - Added error_detail to the get_info() method

Pbarbosa92:
This PR adds the error_detail to the get_info method. The error detail will only be added if it exists.

We also added the cli command `inductiva tasks info task_id` that will print the get_info() to the cli

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -225,6 +225,8 @@ class TaskInfo:
             "\t" + line for line in data_metrics_table.splitlines())
 
         table_str = f"\nTask status: {self.status}"
+        if self.executer and self.executer.error_detail:
+            table_str += f"\n\tStatus detail: {self.executer.error_detail}"
         table_str += f"\n{wall_time_table}"
         table_str += f"\nTime breakdown:\n{time_metrics_table}"
         table_str += f"\nData:\n{data_metrics_table}\n"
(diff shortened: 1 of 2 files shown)


inductiva/tasks/task.py
```diff
@@ -225,6 +225,8 @@ def __str__(self):
             "\t" + line for line in data_metrics_table.splitlines())
 
         table_str = f"\nTask status: {self.status}"
+        if self.executer and self.executer.error_detail:
```
bgregoinductiva, line 228: I think this conditional statement should also check if task status is different from success.

Pbarbosa92: > I think this conditional statement should also check if task status is different from success.

I assume if we have error_details present they are relevant, even if the task was successful. I rather show it every time its present.

bgregoinductiva:
I found a weird output. There is an error detail although the task terminated successfully. @luispcunha is this expected?

```
(env_tests) brunogrego@oxygen:~/inductiva/test_scripts$ inductiva tasks info bycv8e1asxvnz4ufsm265vnxg

Task status: success
	Status detail: The command must start with 'gmx'.
Wall clock time:  16.15 s
Time breakdown:
	Input upload:              0.40 s
	Time in queue:             0.00 s
	Container image download:  1.96 s
	Input download:            0.06 s
	Input decompression:       0.02 s
	Computation:               13.21 s
	Output compression:        0.11 s
	Output upload:             0.12 s
Data:
	Size of zipped output:    878.69 KB
	Size of unzipped output:  2.05 MB
	Number of output files:   33
```

sergiomsantos (approved):

Pbarbosa92:
Waiting for a back end fix to the error detail.

Pbarbosa92:
This was fixed on the back end.

PR #1493, opened 2024-06-21
Template manager simplification

hpenedones:
The goal of this change is to make the TemplateManager class much simpler and its behaviour more predictable.

Fundamentally now it will only do one thing: take a directory of files (some templates, some non-templates), the values for the keywords to replace, and generate a new directory with the same substructure with the values filled in into the placeholder. We maintain the "overwrite" flag work atomically.

```python
TemplateManager.render_dir(templates_dir,
                           target_dir,
                           overwrite=False,
                           key1="text",
                           key2="word")
```

Note that this is even a class method, there is no need to create a TemplateManager object with internal state.

Some notable consequences:
- No more side effects or clever guessing of what folders to create -- this is done explicitly from the outside, very simply.
- It was possible to delete many functions, and completely get rid of the FileManager base class (overall there are now less ~800 lines of code or text than before)
- Tutorials and documentation also greatly simplified (several sections could be deleted)
- However, the new interface is *not* back-compatible with existing code

TODO:
- update other repos of scenarios (WindTunnel, FluidTank, etc) that use the TemplateManager.


inductiva/templating/manager.py
```diff
@@ -1,153 +1,39 @@
...
-                             **render_args)
-
-        return target_file
-
-    def render_dir(self,
-                   source_dir: Optional[types.PathOrStr] = None,
-                   target_dir: Optional[types.PathOrStr] = None,
+    @classmethod
+    def render_dir(cls,
+                   source_dir: types.PathOrStr,
```
sarmento, line 18: What is this PathOrStr?
Can this just be a string?

inductiva/templating/renderers.py
```diff
@@ -1,6 +1,5 @@
 """Base and specific renderers packages to manage template files."""
```
sarmento, line 1: Is this comment correct?

inductiva/templating/renderers.py
```diff
@@ -9,43 +8,8 @@
...
-        if not cls.is_template(file):
-            return file
-        if isinstance(file, pathlib.Path):
-            return file.with_suffix("")
-        return os.path.splitext(file)[0]
-
-
-class JinjaRendererAdapter(RendererAdapter):
-    """Adapter for the Jinja2 template engine.
+class JinjaRenderer():
```
sarmento, line 11: Not really sure why we need this class.
Shouldn't all these be internal methods of the TemplateManager class?

sarmento (commented):
I just left a few comments.
Do you think we could go one step further and remove the pathorstr type and the JinjaRenderer class?

inductiva/templating/manager.py
hpenedones: Yes, this is a refactoring that I did not do here, but we can do it separately, because it happens in a few more places.

inductiva/templating/renderers.py
hpenedones: updated.

inductiva/templating/renderers.py
hpenedones: Yes, you are right, I think it would be possible to merge the two classes into one.

sarmento: Please consider doing that in a subsequent PR. Giving LGTM now.
Thank you!

sarmento (approved):
LGTM!

inductiva/templating/manager.py
hpenedones: thanks, I coordinated with @sergiomsantos and there will be a separate issue and PR.

inductiva/templating/manager.py
```diff
@@ -1,153 +1,39 @@
...
+                   target_dir: types.PathOrStr,
                    overwrite: bool = False,
-                   **render_args) -> pathlib.Path:
+                   **render_args):
```
Pbarbosa92, line 21: We are missing the return type in the docstring.

Pbarbosa92 (approved):
LGTM.

inductiva/templating/manager.py
hpenedones: Oh... thanks, good catch. But actually now this method does not need to return anything, as the target_dir is one of the inputs and it is where the data will appear (no internal logic to create a new directory with a different name)

sergiomsantos (approved):

hpenedones:
Thanks @Pbarbosa92  and @sergiomsantos! Will now merge to development.

PR #1497, opened 2024-06-25
Add tasks compare

Pbarbosa92:
## Waiting for cost to be added to the task info.

This PR adds the method `inductiva.tasks.compare` to compare tasks.
```
Args:
        list_of_tasks: An iterable of tasks.
        sort_by: The metric to sort by. Allowed values: "Computation",
            "Compression", "Upload", "Total Cost".
        reverse: Whether to sort in reverse order.
        verbose: 0: no output, 1: only prints the table, 2: prints table and
        task info
    returns:
        The ID of the first task in the sorted dict.
```
This functionality has also been integrated into the CLI, enabling users to compare tasks by their IDs or within a specific project (comparing all tasks within that project).
```
positional arguments:
  id                    ID(s) of the task(s) to compare.

options:
  -h, --help            show this help message and exit
  -p PROJECT, --project PROJECT
                        Project of the task(s) to compare.
  --sort-by SORT_BY     Sort by a column.
  -r, --reverse         Determines the sort order.
  --no-info             Do not print the information of the first task.
```


bgregoinductiva:
The cost is already available in the API response in dev after this PR: https://github.com/inductiva/inductiva-web-api/pull/967. It is necessary to update the client.

Issue #138, opened 2023-03-24
Express generation of tank mesh file in interface that represents shapes in meshes


fabiocruz:
> [...] Couldn't we provide a higher level gmsh utils interface that received the shape object and the position and then performed all this logic?

_Originally posted by @luispcunha in https://github.com/inductiva/inductiva/pull/119#discussion_r1146503299_

PR #1501, opened 2024-06-26
Renames types.Path -> str

Franreno:
Closes [#138](https://github.com/inductiva/tasks/issues/138)

### DIFF
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -154,7 +154,8 @@ def upload_input(api_instance: TasksApi, task_id, original_params,
 def download_output(
         api_instance: TasksApi,
         task_id,
-        output_dir: Optional[types.Path] = None) -> Tuple[List, pathlib.Path]:
+        output_dir: Optional[types.PathOrStr] = None
+) -> Tuple[List, pathlib.Path]:
     """Downloads the output of a given task from the API.
 
     Args:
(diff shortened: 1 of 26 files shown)


hpenedones:
Thanks @Franreno , I've been discussing with @sergiomsantos that we should actually try to get rid of these custom types for paths -- we shouldn't need to define a new type for things that are very standard. We will try to consolidate everything, potentially just using strings, which is one of the types we accept anyways -- and that's enough, no need to support multiple variants.

Franreno:
> Thanks @Franreno , I've been discussing with @sergiomsantos that we should actually try to get rid of these custom types for paths -- we shouldn't need to define a new type for things that are very standard. We will try to consolidate everything, potentially just using strings, which is one of the types we accept anyways -- and that's enough, no need to support multiple variants.

Thanks for the feedback @hpenedones. I've started a refactor to change from `PathOrStr` to only `str`.  I've also added some tests to some old code that didn't had tests.

Pbarbosa92 (approved):
LGTM, but we need to test this thoroughly.

Pbarbosa92:
**QA**

I have tested this PR and everything works perfectly for me.

I ran multiple simulations using `str` as the `input_dir` and successfully downloaded the outputs. Additionally, I downloaded the outputs using the CLI without any issues. Using both filenames and `output_dir` as `str` also worked flawlessly.

sergiomsantos (approved):

PR #1504, opened 2024-06-27
[Refactor] Remove JinjaRenderer class

sergiomsantos:
This PR follows on the refactoring initiated with #1493 to further simplify the template manager by removing the `JinjaRenderer` class and incorporating its logic directly into the manager module.

The behaviour introduced in #1493, with respect to the `TemplateManager.render_dir` method, is kept.


Pbarbosa92 (approved):
LGTM

inductiva/templating/manager.py
```diff
@@ -11,12 +13,53 @@ class TemplateManager:
...
+        TEMPLATE_EXTENSION (str): The file extension used by Jinja2 templates.
     """
 
-    @classmethod
-    def render_dir(cls,
-                   source_dir: types.PathOrStr,
-                   target_dir: types.PathOrStr,
+    TEMPLATE_EXTENSION = ".jinja"
+
+    def __init__(self, template_dir: str):
```
hpenedones, line 23: If the TemplateManager is meant to be used only by accessing a class method (render_dir), let's try to avoid having a constructor that allows to create instances of it (what are the instances supposed to do?). That is quite confusing, I think.

hpenedones (changes requested):

hpenedones:
Ideally, the breaking change that would require str would go in a separate PR (because I think there are more places where we can consolidate that, and we could totally remove the type definition). And this PR would be non-breaking.

inductiva/templating/manager.py
sergiomsantos: addressed in [e7fc2ce](https://github.com/inductiva/inductiva/pull/1504/commits/e7fc2ce0e82d140b76eac5b8ad93edc305a75da8)

sergiomsantos:
> Ideally, the breaking change that would require str would go in a separate PR (because I think there are more places where we can consolidate that, and we could totally remove the type definition). And this PR would be non-breaking.

addressed in [e7fc2ce](https://github.com/inductiva/inductiva/pull/1504/commits/e7fc2ce0e82d140b76eac5b8ad93edc305a75da8)

Pbarbosa92 (approved):
LGTM!

PR #1522, opened 2024-07-10
257 Use spot machines by default

sergiomsantos:
[#257](https://github.com/inductiva/tasks/issues/257)

# Use Spot machines by default

This PR ensures that all machines from the `MachineGroup` and `ElasticMachineGroup` are instantiated as **spot** by default.


Pbarbosa92 (approved):
LGTM!

Pbarbosa92:
QA

LGTM.
If the task status is `SPOTINSTANCEPREEMPTED` when waiting I get the message:
```
The task was preempted by the cloud provider.
Consider using non-spot machines by setting `spot=False` when instantiating the machine group.
```
And when i create a resource spot is `True` by default.

PR #1534, opened 2024-07-18
Add `resubmit_on_preemption`option to the `Simulator.run` method

sergiomsantos:
Task [#267](https://github.com/inductiva/tasks/issues/267)

This PR addresses the need to add the parameter `resubmit_on_preemption` to the Simulator.run method that specifies if the task is to be resubmitted when a spot is preempted and the task is not completed.


Pbarbosa92 (approved):
LGTM!

PR #1535, opened 2024-07-18
Improved exception handling

Pbarbosa92:
This PR improves on the exception handling of the client.
We now print the error message when we get an APIException, followed by the first and last 2 lines of the stack.
After that we point that the user can get more information on the logs file.

### DIFF
--- a/inductiva/logs/log.py
+++ b/inductiva/logs/log.py
@@ -1,4 +1,5 @@
 """Custom logging functions"""
+import traceback
 import json
 import os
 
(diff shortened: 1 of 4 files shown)


inductiva/logs/log.py
```diff
@@ -37,16 +38,59 @@ def format(self, record):
...
+    ]
+
+    result_string = ("\n".join(first_formatted_tb) + "\n" + \
+                     "    ...\n" +
+                     "\n".join(last_formatted_tb) + "\n")
+
+    return result_string
+
+
+def _handle_api_exception(exc_type, exc_value, exc_traceback, notebook: bool):
```
sergiomsantos, line 78:
```suggestion
def _handle_api_exception(exc_type, exc_value, exc_traceback, is_notebook: bool):
```

sergiomsantos (approved):

Pbarbosa92:
@sergiomsantos can you QA this PR ? 
You can use a container image that is not from kutu on prod, that should generate an APIException.

PR #1544, opened 2024-07-24
Added mpiconfig to the command class

Pbarbosa92:
Issue https://github.com/inductiva/tasks/issues/273

This PR adds mpiconfigs to the commands send to the backend.

```python
import inductiva
import inductiva.commands
from inductiva.commands.mpiconfig import MPIConfig

mpi_config = MPIConfig("1.10.7",np=2,use_hwthread_cpus=True)

command = inductiva.commands.Command("ls",mpi_config=mpi_config)
command2 = inductiva.commands.Command("cd")

print(command.to_dict())
print(command2.to_dict())
```
Will give 
```
{'cmd': 'ls', 'prompts': [], 'mpi_config': {'version': '1.10.7', 'options': {'np': 2, 'use-hwthread-cpus': True}}}
{'cmd': 'cd', 'prompts': [], 'mpi_config': None}
```

### DIFF
--- a/inductiva/commands/mpiconfig.py
+++ b/inductiva/commands/mpiconfig.py
@@ -16,23 +16,13 @@ class MPIConfig:
         self.version = version
         self.options = kwargs
 
-    def _replace_underscores_in_keys(self, d):
-        """Replace underscores with hyphens in dictionary
-        Since we are using kwargs we cant use flags like use-hwthread-cpus
-        so we need to use use_hwthread_cpus as the parameter and then replace
-        the underscores with hyphens before sending the request to the API.
-        """
-        new_dict = {}
-        for key, value in d.items():
-            new_key = key.replace('_', '-')
-            if isinstance(value, dict):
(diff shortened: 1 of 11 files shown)


inductiva/commands/mpiconfig.py
```diff
@@ -0,0 +1,38 @@
...
+                new_dict[new_key] = self._replace_underscores_in_keys(value)
+            else:
+                new_dict[new_key] = value
+        return new_dict
+
+    def to_dict(self):
+        """Convert this MPIConfig to a dictionary representation.
+        This method will replace the "_" with "-".
+        """
+        return self._replace_underscores_in_keys(self.__dict__)
```
sergiomsantos, line 38: we can even inline the replacement here (no need for an additional method):

```suggestion
        return {
            version: self.version,
            options: {k.replace('_', '-'): v for k, v in self.options.items()}
        }
```

sergiomsantos (approved):

PR #1553, opened 2024-08-01
184 update inductiva rm all

Pbarbosa92:
https://github.com/inductiva/tasks/issues/184
We can no longer send "*" to inductiva storage remove. The user should use remove --all if he wants to delete everything.


bgregoinductiva:
This PR can be closed as obsolete. The command `inductiva storage rm` does not work and needs a refactor as explained here: https://github.com/inductiva/tasks/issues/184#issuecomment-2298485148

Pbarbosa92:
Closing this PR because the storage remove is now task based and it does not longer make sense to do rm all.

PR #1576, opened 2024-09-09
Make `on` argument on `simulator.run` method mandatory

luispcunha:
It is now mandatory to specify a computational resource when submitting a task, as a result of no longer providing a shared queue. The `on` argument doesn't have a default, but remains a keyword argument.

Tests and examples that rely on default machine group need to be updated.


Pbarbosa92:
Why was this closed @luispcunha ?

PR #1585, opened 2024-09-11
Fix user-specified local machines

rodrigotuna:
This PR solves problems related to running simulations on user-specified local machines. 
The following changes were added:
- Bypassing unnecessary hardware checks
- Enforcing consistency in providers when running on user-defined machines


inductiva/resources/machine_groups.py
```diff
@@ -50,24 +50,28 @@ def get_by_name(machine_name: str):
...
-def _get_machine_group_class(machine_type: str, is_elastic: bool):
+def _get_machine_group_class(machine_type: str, is_elastic: bool,
+                             provider: str):
     """Returns the class of the machine group"""
+    kwargs = {}
     if is_elastic:
         mg_class = resources.ElasticMachineGroup
     elif machine_type == "standard":
         mg_class = resources.MachineGroup
+        kwargs = {"provider": provider}
```
Pbarbosa92, line 69: Why do we only use `kwargs` when `machine_type == "standard"`? Shouldn't the back end also send the `provider_id` for other machine types?

inductiva/resources/machines_base.py
```diff
@@ -65,7 +65,7 @@ def __init__(
         provider = machine_types.ProviderType(provider)
         self.provider = provider.value
 
-        if data_disk_gb <= 0:
+        if data_disk_gb <= 0 and provider != machine_types.ProviderType.LOCAL:
```
Pbarbosa92, line 68: So, if our machine is LOCAL I can send data_disk_gb=-1 ?

inductiva/tasks/run_simulation.py
```diff
@@ -40,6 +40,9 @@ def run_simulation(
         "sim_dir": pathlib.Path,
     }
 
+    if computational_resources:
+        provider_id = computational_resources.provider
```
Pbarbosa92, line 44: I am a bit confused now. Provider id is linked to the resource (machine) but we can send a provider id on the simulator.run, that can be overwritten.

Another thing (correct me if I'm wrong) some machines might have no provider id right?  (as seen on the machine_groups.py)
```python
kwargs = {}
if is_elastic:
    mg_class = resources.ElasticMachineGroup
```

I think we need to clean this up, we should remove the provider_id from the simulator.run and make sure all machines have a provider_id.

Pbarbosa92 (changes requested):
I have some questions

inductiva/resources/machine_groups.py
rodrigotuna: Not part of the constructor.

Pbarbosa92: Maybe we can add the provider to the BaseMachineGroup. This would make the code more uniform and it would make sense to me that every resource has a provider. If BaseMachineGroup had a provider we wouldn't need to be using kwargs (kwargs can make the code a bit harder to understand and to follow).

inductiva/resources/machines_base.py
rodrigotuna: The server has no knowledge of the disk size of a local machine, that variable has no meaning if the machine is local.

Pbarbosa92: So, if we are changing the behavior of a class based on the type, shouldn't that be a separate class? In this case a class LocalMachine with no data_disk_gb.

As long as a class has this variable, we need to check if data_disk_gb is 0 or less.

So, we can create a new class or keep this one but with this check. Even if it does not make sense to a local machine asking for a disk with 0 or negative GB should give an exception.

hpenedones: Yes, I think we will need to refactor these classes -- at the moment they are quite messy...
It's called "BaseMachineGroup" which would imply is almost an Abstract class without implementations, but it's full of logic. It's docstring says class  """Base class to manage Google Cloud resources."""  and then takes an argument called "provider" which can be something different than GCP... It's full of if statements regarding the provider, which potentially could be dealt with using inheritance, like Paulo suggested, or leaving some logic to the backend...

@rodrigotuna : you could probably be unblocked by just passing any dummy value for this variable for now --- but I will open an issue for us to refactor these classes.

PR #1587, opened 2024-09-11
#352 Allow resize vm disk

Pbarbosa92:
Issue https://github.com/inductiva/tasks/issues/352
This PR introduces the ability to configure VM disk resizing using a new `DiskConfig` class. This enhancement allows for more flexibility in managing disk sizes and supports auto-resizing behavior based on defined thresholds.

### Key Changes

A new class, `DiskConfig`, has been added to handle disk configuration. This class provides several parameters to control disk size and automatic resizing:

- `max_size_gb`: **int** (required) - Specifies the maximum disk size.
- `is_resizable`: **bool** (optional default=True) - Defines if the disk is resizable (True by default).

### Usage Examples

The following examples demonstrate how to configure disk size and enable or disable disk resizing:

1. **Fixed Disk Size (Resizing OFF)**
    ```python
    # Creates a machine group with a fixed disk size of 11 GB
    disk = inductiva.resources.DiskConfig(11,is_resizable=False)
    inductiva.resources.MachineGroup(machine_type='c2d-standard-4', num_machines=1, disk_config=disk)
    ```

2. **Auto-Resizing Disk**
    ```python
    # Creates a machine group with an initial disk size of 11 GB, and auto-resizing enabled.
    # Resizing starts when the disk reaches 5 GB, increasing by 10 GB, up to a maximum of 11 GB.
    disk = inductiva.resources.DiskConfig(11)
    inductiva.resources.MachineGroup(machine_type='c2d-standard-4', num_machines=1, disk_config=disk)
    ```

3. **Legacy Disk Size Specification (Resizing OFF)**
    ```python
    # Uses the existing `data_disk_gb` parameter to set the disk size to 12 GB
    inductiva.resources.MachineGroup(machine_type='c2d-standard-4', num_machines=1, data_disk_gb=12)
    ```

4. **Default Configuration (Resizing OFF)**
    ```python
    # Creates a machine group with the default disk size of 10 GB
    inductiva.resources.MachineGroup(machine_type='c2d-standard-4', num_machines=1)
    ```

### Notes

- If `max_size_gb` is provided in the `DiskConfig`, it takes precedence over the legacy `data_disk_gb` parameter.
- To maintain backward compatibility, the existing `data_disk_gb` parameter is still supported, but its usage is optional when using `DiskConfig`.
- `resize_trigger_gb` and `resize_increment_gb` are internal variables that are always 5 and 10 gb. They will only be used if is_resizable == True


Pbarbosa92:
I keep the data_disk_gb to avoid a breaking change with old code and our docs....Making all 3 into a class would for you to create 2 new objects for a disk, that would become too verbose.Enviado do meu Galaxy
-------- Mensagem original --------De : Luís Cunha ***@***.***> Data: 11/09/24  17:54  (GMT+00:00) Para: inductiva/inductiva ***@***.***> Cc: Paulo Barbosa ***@***.***>, Author ***@***.***> Assunto: Re: [inductiva/inductiva] #352 Allow resize vm disk (PR #1587) 


If size_gb is provided in the DiskConfig, it takes precedence over the legacy data_disk_gb parameter.


Why not keep the data_disk_gb in addition to the DiskConfig? Note that this introduces a breaking change.


If any of the optional parameters (resize_trigger_gb, resize_increment_gb, max_size_gb) are specified, all must be provided to enable resizing behavior.


If all or none must be provided, I'd rather have them be encapsulated in a class, that way it is more clear.

inductiva/resources/disk_config.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # resize_trigger_gb (int): Free space threshold (in GB) to trigger
+    #     resizing. Example: If set to 5GB, the disk will resize when 5GB of
+    #     free space remains.
+    resize_trigger_gb: int = field(default=5, init=False)
+
+    # resize_increment_gb (int): Amount (in GB) to increase the disk size
+    #     during each resize.
+    resize_increment_gb: int = field(default=10, init=False)
+    is_resizable: bool = True
```
sergiomsantos, line 36: Put both mandatory arguments together:

```suggestion
    max_size_gb: int
    is_resizable: bool = True

    # resize_trigger_gb (int): Free space threshold (in GB) to trigger
    #     resizing. Example: If set to 5GB, the disk will resize when 5GB of
    #     free space remains.
    resize_trigger_gb: int = field(default=5, init=False)

    # resize_increment_gb (int): Amount (in GB) to increase the disk size
    #     during each resize.
    resize_increment_gb: int = field(default=10, init=False)
```

inductiva/resources/disk_config.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @property
+    def resize_config(self):
```
sergiomsantos, line 39:
```suggestion
    def to_dict(self):
```

inductiva/resources/disk_config.py
```diff
@@ -0,0 +1,48 @@
...
+        """Return the disk resize configuration as a dictionary.
+        If the disk is not resizable, it will return None."""
```
sergiomsantos, line 41: why return None? It simply return a dict representation of the object.

inductiva/resources/disk_config.py
```diff
@@ -0,0 +1,48 @@
...
+        if not self.is_resizable:
+            return None
+        return {
+            "free_space_threshold_gb": self.resize_trigger_gb,
+            "size_increment_gb": self.resize_increment_gb,
+            "max_disk_size_gb": self.max_size_gb,
```
sergiomsantos, line 47: the names do not match, which is a bit weird. If this method only exists for the sake of serializing the object for the BE request, then I suggest moving it to where it is needed.

inductiva/resources/machines.py
```diff
@@ -52,6 +54,9 @@ def __init__(
             threads_per_core: The number of threads per core (1 or 2).
             spot: Whether to use spot machines.
             data_disk_gb: The size of the disk for user data (in GB).
+            disk_config: Disk configuration for the machine group. This config
+                includes disk size_gb and all the parameters needed for resizing
+                the disk. If provided will take precedence over data_disk_gb.
```
sergiomsantos, line 59: I think they should be disjoint. Either one or the other, not both.

sergiomsantos (changes requested):

inductiva/resources/disk_config.py
Pbarbosa92: I did this to avoid logic when creating the request to send to the back end. I will change the method to be called resizing_disk_config, if the disk is not resizable it makes sense that resizing_disk_config is None.
Like this i just need to have 
```
instance_group_config = inductiva.client.models.VMGroupConfig(
            ...
            dynamic_disk_resize_config=self.disk_config.resizing_disk_config,
            **kwargs,
        )
```
Like this there is no need for if's here

inductiva/resources/machines.py
Pbarbosa92: I've added a check for that

inductiva/resources/machines_base.py
```diff
@@ -160,6 +173,16 @@ def _convert_auto_terminate_ts(
...
         return None
 
+    def _disk_config_to_dict(self):
+        if self.disk_config is None or not self.disk_config.is_resizable:
+            return None
+
+        return {
+            "free_space_threshold_gb": self.disk_config.resize_trigger_gb,
+            "size_increment_gb": self.disk_config.resize_increment_gb,
+            "max_disk_size_gb": self.disk_config.max_size_gb,
```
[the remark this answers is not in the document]
Pbarbosa92: We changed the dataclass, and not the user only needs to configure the max_size_gb. I can change the old names back. I just removed disk from the names because this is a DiskConfig the disk in the name would be redundant.

hpenedones:
Thanks @Pbarbosa92 : I think we might want to simplify this a bit... I don't have a final solution on top of my mind either -- but let's brainstorm about one! :)

My concerns are that, essentially, this new class only responsibility is to convert its fields into a dictionary. And then we have several tests, but none of them tests behaviour (of the disk resizing itself), and they are only testing things related to this class to dict conversion. It might be ok to not test the disk resizing here (could be too slow, and it's a backend feature), but fundamentally we would still be introducing a significant amount of code for almost no reason.

In addition, this new argument is not quite compatible with the existing one (user can not pass both)...

inductiva/resources/machines.py
```diff
@@ -29,6 +30,7 @@ def __init__(
         threads_per_core: int = 2,
         spot: bool = True,
         data_disk_gb: int = 10,
+        disk_config: DiskConfig = None,
```
hpenedones, line 33: auto_resize_disk_max_gb

inductiva/resources/machines_base.py
```diff
@@ -47,8 +48,12 @@ def __init__(
               Check https://cloud.google.com/compute/docs/machine-resource for
               more information about machine types.
             provider: The cloud provider of the machine group.
-            data_disk_gb: The size of the disk for user data (in GB).
             threads_per_core: The number of threads per core (1 or 2).
+            data_disk_gb: The size of the disk for user data (in GB).
+            auto_resize_disk_max_gb: The maximum size to resize the disk. If set
```
hpenedones, line 53: Consider changing to: 
"The maximum size in GB that the hard disk of the cloud VM can reach. If set, the disk will be automatically resized, during the execution of a task, when the free space falls below a certain threshold. This mechanism helps prevent "out of space" errors, that can occur when a task generates a quantity of output files that exceeds the size of the local storage. Increasing disk size during task execution increases the cost of local storage associated with the VM  -- therefore the user must set an upper limit to the disk size, to prevent uncontrolled costs. Once that limit is reached, the disk is no longer automatically resized, and if the task continues to output files, it will fail."

inductiva/resources/machines.py
```diff
@@ -192,6 +199,10 @@ def __init__(
             spot: Whether to use spot machines.
             threads_per_core: The number of threads per core (1 or 2).
             data_disk_gb: The size of the disk for user data (in GB).
+            auto_resize_disk_max_gb: The maximum size to resize the disk. If set
```
hpenedones, line 202: left a suggestion below for this docstring...

inductiva/resources/machines_base.py
```diff
@@ -68,13 +73,26 @@ def __init__(
         if data_disk_gb <= 0:
             raise ValueError("`data_disk_gb` must be positive.")
 
+        if auto_resize_disk_max_gb is not None:
+            if not isinstance(auto_resize_disk_max_gb,
+                              int) or auto_resize_disk_max_gb <= 0:
+                raise ValueError(
+                    "`auto_resize_disk_max_gb` must be a positive integer.")
+
+            if auto_resize_disk_max_gb < data_disk_gb:
```
hpenedones, line 82: is this strict enough? potentially we could ask the auto_resize_disk_max_gb to be at least size_increment_gb bigger than data_disk_gb... Not sure what happens if you set the value to auto_resize_disk_max_gb = data_disk_gb -- @luispcunha ?

Also, maybe we should suggest that it should be at least 2x bigger than the data_disk_gb, otherwise it's not worth using this mechanism?

hpenedones (approved):
Thanks @Pbarbosa92 this is much simpler! left a couple of small suggestions.

inductiva/resources/machines_base.py
Pbarbosa92: @luispcunha does the max size need to be a factor of the increment? 
does this work?
data_disk_gb - 10
max_disk_size_gb - 20
increment - 3

Pbarbosa92: I've changed this check to
```
if auto_resize_disk_max_gb < data_disk_gb + self._size_increment_gb:
                raise ValueError("`auto_resize_disk_max_gb` must be greater "
                                 "than or equal to `data_disk_gb + "
                                 f"{self._size_increment_gb}GB`.")
```

sergiomsantos:
Having the need to have `auto_resize_disk_max_gb` greater than `data_disk_gb` plus a a term that is unknown to the user is akward. Also, I don't see why `auto_resize_disk_max_gb` cannot be equal to `data_disk_gb`, which would simply mean that no resize would take place.

The introduction of the `auto_resize_disk_max_gb` argument, in addition to the already existing `data_disk_gb`, will only add to the backward compatibility issues that are piling up. This is a perfect example for the need of using a parameter object (dto) to configure the runtime behaviour of a function, just as clean code suggests.

hpenedones:
The back compatibility aspect is a very good point -- thanks @sergiomsantos for pointing that out. I think there is some normal "tension" between having a clean, simple interface, where the input arguments you need are explicit and are of primitive types -- and the fact that if you need to regularly add more arguments this starts to get a bit messy and you incur more risk of non back compatibility (if you later decide to clean up and change the interface).
The deeper problem is that the MachineGroup and related classes are in need of some care, and we might need to do a breaking change before our 1.0 release -- so maybe we could plan that and keep in mind that groups of arguments that are related and likely to change could potentially be encapsulated. It's non-trivial where to put the boundaries, though, because the extreme of having just a big bag of key-value pairs, or having many custom classes to create arguments, impacts negatively the code the user has to write. So we have to strike a balance there.

Pbarbosa92:
@hpenedones it's not clear to me what changes you want to do.

sergiomsantos (approved):

PR #1591, opened 2024-09-12
Run snippets inside the simulators docs

Pbarbosa92:
Run the snippets inside the simulators docs.


sergiomsantos (approved):
We're always just using a single machine in the examples. I think we can make it a bit less verbose by simply doing:

```python
machine_group = inductiva.resources.MachineGroup('c2-standard-4')
```

rather than

```python
machine_group = inductiva.resources.MachineGroup(
    machine_type="c2-standard-4",
    num_machines=1,
    data_disk_gb=10)
```

PR #1598, opened 2024-09-13
resource.start now takes wait_on_pending_quota argument

Pbarbosa92:
This PR adds the flag wait_on_pending_quota to the resource.start().

By default is False, so all old code should work fine.

Issue https://github.com/inductiva/tasks/issues/351


inductiva/resources/machines_base.py
```diff
@@ -31,6 +31,8 @@ class ResourceType(enum.Enum):
 class BaseMachineGroup(ABC):
     """Base class to manage Google Cloud resources."""
 
+    QUOTAS_EXCEEDED_SLEEP_TIME = 60
```
sergiomsantos, line 34:
```suggestion
    QUOTAS_EXCEEDED_SLEEP_SECONDS = 60
```

sergiomsantos (approved):

PR #1605, opened 2024-09-19
Improve openfoam Docs

Pbarbosa92:
This PR aims to improve the docs of openfoam, adding an `Advanced Example`.

Edit: I've changed this PR to merge to main in order to publish the new docs page. After the merge to main i will merge main to dev.


docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,216 @@ task.download_outputs()
...
+machine_group = inductiva.resources.MachineGroup(machine_type="c3d-standard-180")
+machine_group.start()
+
+import inductiva
+
+# Set simulation input directory
+input_dir = "/path/to/highLiftConfiguration"
+
+# Set the simulation commands
+commands = [
```
sarmento, line 206: Can't all this be in a file?
I think this is one of UI the issues we need to solve :)
We can read all the lines to a list, right?
Or we can have a parameter called commands_file.
@hpenedones wdyt?

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,216 @@ task.download_outputs()
...
+  0 drwxr-xr-x    4      128 Sep 19 16:06 processor127
+ 32 -rw-r--r--    1    14437 Sep 19 16:06 stderr.txt
+536 -rw-r--r--    1   272331 Sep 19 16:06 stdout.txt
+  0 drwxr-xr-x   30      960 Sep 19 16:06 system
+ 24 -rw-r--r--    1    11399 Sep 19 16:06 thumbnail.png
+```
+
+You can now perform post-processing on the results, just as you would if you had
+run the simulation locally.
+
```
sarmento, line 328: Can you also print the execution times?
And maybe try with two different machines. 
Try a bigger one....

docs/simulators/OpenFOAM.md
hpenedones: Yes, sounds like a good idea!

Pbarbosa92: We don't need a parameter for that. We can read a txt file with one command per line. It's pretty simple to read it into a list of strings.
```python
# Read the simulation commands
with open(input_dir + '/input.txt', 'r') as file:
    commands = [line.strip() for line in file]
```

I will update with this.

docs/simulators/OpenFOAM.md
Pbarbosa92: I will run the following executions (adding something about the times here after that)

c3d-standard-90 - nCores/n_vcpus = 90
c3d-standard-180 - nCores/n_vcpus = 180
c3d-standard-360 - nCores/n_vcpus = 360

c3d-standard-90 - nCores/n_vcpus = 45 (Disable hyperthreading)
c3d-standard-180 - nCores/n_vcpus = 90 (Disable hyperthreading)
c3d-standard-360 - nCores/n_vcpus = 180 (Disable hyperthreading)

docs/simulators/OpenFOAM.md
sarmento: But I wouldn't mind a parameter and do that on the client side.
That way we can hide all this from the user. We want to make thing simpler.
Plus, you code in not correct. You should be using os.path.join :)

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,278 @@ task.download_outputs()
...
+After the simulation completes, the output files will be downloaded locally by
+calling `task.download_outputs()`. The outputs will include simulation results
+like mesh quality reports, final results, and log files for each step in the process.
+
+To inspect the output, look under the `inductiva_output` folder where the
+results are downloaded:
+
+```bash
+ls -lasgo inductiva_output/task_id
```
sergiomsantos, line 333:
```suggestion
ls -lasgo inductiva_output/<task_id>
```

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,278 @@ task.download_outputs()
...
+runApplication sed -i s/"ff_zMin"/"ff_SAVE"/g constant/polyMesh/boundary
+runApplication sed -i s/"ff_zMax"/"ff_zMin"/g constant/polyMesh/boundary
+runApplication sed -i s/"ff_SAVE"/"ff_zMax"/g constant/polyMesh/boundary
+runApplication checkMesh
+runApplication surfaceFeatureExtract
+runApplication decomposePar -decomposeParDict system/decomposeParDict.SHM
+runParallel snappyHexMesh -decomposeParDict system/decomposeParDict.SHM -dict system/snappyHexMeshDict -overwrite
+runParallel checkMesh -decomposeParDict system/decomposeParDict.SHM -latestTime -meshQuality
+runApplication reconstructParMesh -mergeTol 1e-08 -constant -latestTime
+runApplication cp -r constant/polyMesh constant/polyMesh.origHalf
```
sergiomsantos, line 268: if you can't run arbitrary commands, just prefix them with `runApplication` and use openfoam 😅 

```
runApplication rm -rf /
```

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,278 @@ task.download_outputs()
...
+goal was to assess how these hardware differences influence simulation time.
+
+| Machine Configuration            | nCores/n_vCPUs           | Hyperthreading Status        | Time (min:sec) |
+|-----------------------------------|--------------------------|------------------------------|----------------|
+| c3d-standard-90                   | 90                       | Enabled                      | 7:48           |
+| c3d-standard-180                  | 180                      | Enabled                      | 7:39           |
+| c3d-standard-360                  | 360                      | Enabled                      | 8:32           |
+| c3d-standard-90                   | 45                       | Disabled (Hyperthreading Off) | 7:20           |
+| c3d-standard-180                  | 90                       | Disabled (Hyperthreading Off) | 6:46           |
+| c3d-standard-360                  | 180                      | Disabled (Hyperthreading Off) | 6:57           |
```
sergiomsantos, line 376: are these results with hyperthreding disabled (`threads_per_core` = 1) or just using using half of the total number of available vcpus?

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,278 @@ task.download_outputs()
...
+```
+
+The commands used in this simulation were taken from the `Allrun` script found
+in the downloaded files. Some adjustments were made because we execute all
+commands directly from the `highLiftConfiguration` directory, meaning we can't
+use `cd` to navigate into subdirectories and run commands from there.
+Additionally, every command must start with `runApplication` or `runParallel`,
+which is why even basic commands like `rm` and `mv` are preceded by `runApplication`.
+
+We have consolidated all the commands into a file named `input.txt`, with each
```
sergiomsantos, line 237: I don't like having commands in a separate file. It just makes everything more verbose.
Apart from that, I think `input.txt` is not a good name for the commands file. Why not `commands.txt`?

Pbarbosa92: Changed it to `commands.txt`

sarmento:
Please activate the docs pipeline for staging, so I can read the whole thing there.
Thank you!

Pbarbosa92:
> Please activate the docs pipeline for staging, so I can read the whole thing there. Thank you!

Done. You can read it there now.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
 machine_group.terminate()
 ````
 
+## Advanced Example: OpenFOAM Simulation (ExaFOAM Benchmark)
```
sarmento, line 129: ## Advanced Example: Running MB9 micro-benchmark by ExaFOAM

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+
+Here's a guide for running an OpenFOAM simulation using the Inductiva platform,
```
sarmento, line 131: Try to remove this paragraph.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+focusing on a high-lift configuration case from the `rhoPimpleFoam` solver.
+This specific example is based on one of the benchmark cases provided by the
+[ExaFOAM project](https://exafoam.eu), which aims to develop highly
+scalable CFD solvers using OpenFOAM for exascale computing. We'll use a powerful
+cloud setup to handle the computationally intensive mesh generation and simulation.
+
+We will run a complex OpenFOAM simulation based on a high-lift configuration
```
sarmento, line 138: Try to start here.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+The MB9 micro-benchmark captures the essential characteristics of the Grand
+Challenge (such as flow physics and simulation approach) while requiring
+significantly fewer computational resources. It features a two-dimensional,
+three-element high-lift wing configuration, simulated with the IDDES model,
+which provides WMLES functionality in regions of resolved near-wall turbulence.
+This case is based on the well-known 30P30N test case, extensively studied in
+the 4th AIAA CFD High Lift Prediction Workshop (HLPW-4) and supported by
+available experimental data.
+
+Additionally, this simulation utilizes dynamic code compilation. This can be
```
sarmento, line 154: You don't need to talk about this and give all this explanation. Users will assume that this is a natural feature of the Inductiva.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+the simulation. However, in the stdout file, the creation of a new library during
+runtime is confirmed:
+
+```
+Creating new library in "dynamicCode/_6a0e8d329f43dbbd4e13ef75c75be041b26176fc/platforms/linux64GccDPInt32Opt/lib/libcodeStream_6a0e8d329f43dbbd4e13ef75c75be041b26176fc.so"
+```
+
+This dynamic compilation feature allows for runtime adjustments and optimizations,
+contributing to the flexibility and efficiency of OpenFOAM simulations in
+high-performance computing environments.
```
sarmento, line 174: Yes. All this text about dynamic code can be removed.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+
+To make the simulation run faster, we made some adjustments to the original
+parameters. Specifically, we modified the time step (`dt`) to 0.00002, the
+initial time to 0.10, and the final time to 0.30. These changes help reduce the
+overall simulation time while maintaining reasonable accuracy for this high-lift
+configuration case.
+
+### Configuring and Running the Simulation
+
+We'll run the simulation on a virtual machine with 180 CPUs using the
```
sarmento, line 214: vCPUs

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+
+The MB9 micro-benchmark captures the essential characteristics of the Grand
+Challenge (such as flow physics and simulation approach) while requiring
+significantly fewer computational resources. It features a two-dimensional,
+three-element high-lift wing configuration, simulated with the IDDES model,
+which provides WMLES functionality in regions of resolved near-wall turbulence.
+This case is based on the well-known 30P30N test case, extensively studied in
+the 4th AIAA CFD High Lift Prediction Workshop (HLPW-4) and supported by
+available experimental data.
+
```
sarmento, line 153: In this example, we will run a modified version of this benchmark using two hardware configurations powered by Inductiva. First, we will run the simulation workflow on a large 360 vCPU machine. This is one of the largest single node configuration one can use to run simulations at Inductiva. Then, to see if we can halve the execution time of the simulation, we will run the simulation on a MPI cluster with 2 of these machines large machines.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
 machine_group.terminate()
 ````
```
sarmento, line 128: Isn't the previous example the same as the foundation? Do the openfoam commands change? If not, then I would remove this example and just talk about the different parameter in the constructor.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+    commands = [line.strip() for line in file]
+
+# Initialize the Simulator
+openfoam = inductiva.simulators.OpenFOAM(distribution="esi")
+
+# Run simulation
+task = openfoam.run(
+    input_dir=input_dir,
+    commands=commands,
+    n_vcpus=128,
```
sarmento, line 244: Shouldn't it be 180?

docs/simulators/OpenFOAM.md
sarmento: Let's just use the 360 machines on this example. We use both the single node and the 2 node configuration. Just those 2. Also, we will be using the 2 thread per core, right? Since that is the default, don't even talk about that. We will be writing about all those things in detail on another document. For this one let's reduce the complexity of the discussion. Will just show how to run this on a really large machine and then on an MPI Cluster

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+machine_group = inductiva.resources.MachineGroup(machine_type="c3d-standard-180")
+machine_group.start()
+
+import inductiva
+
+# Set simulation input directory
+input_dir = "/path/to/highLiftConfiguration"
+
+# Read the simulation commands
+with open(os.path.join(input_dir,'commands.txt'), 'r') as file:
```
sarmento, line 234: Before presenting this script, show the commands file. Remember that you are writing for OpenFOAM experts, so 1st give them something that they are confortable with.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+
+# Initialize the Simulator
+openfoam = inductiva.simulators.OpenFOAM(distribution="esi")
+
+# Run simulation
+task = openfoam.run(
+    input_dir=input_dir,
+    commands=commands,
+    n_vcpus=128,
+    use_hwthread=True,
```
sarmento, line 245: Do you need this parameter here?

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+steps, including mesh generation (`blockMesh`, `snappyHexMesh`), and finally,
+the simulation run itself. 
+
+Below is the Python code to configure and run the OpenFOAM simulation:
+
+```python notest
+import inductiva
+import os
+
+machine_group = inductiva.resources.MachineGroup(machine_type="c3d-standard-180")
```
sarmento, line 225: Are we not running on spots?
What is the default?

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,323 @@ task.download_outputs()
...
+ 32 -rw-r--r--    1    14437 Sep 19 16:06 stderr.txt
+536 -rw-r--r--    1   272331 Sep 19 16:06 stdout.txt
+  0 drwxr-xr-x   30      960 Sep 19 16:06 system
+ 24 -rw-r--r--    1    11399 Sep 19 16:06 thumbnail.png
+```
+
+You can now perform post-processing on the results, just as you would if you had
+run the simulation locally.
+
+### Small benchmark
```
sarmento, line 390: Remove this section and just have a section on running with MPI instead. Show the difference.

docs/simulators/OpenFOAM.md
Pbarbosa92: The commands change. One uses `surfaceFeatures` and the other `surfaceFeatureExtract`

docs/simulators/OpenFOAM.md
Pbarbosa92: The original simulation ran on 128. I've changed it to 180 in the new iteration.

docs/simulators/OpenFOAM.md
Pbarbosa92: In this case no. But i wanted to make it explicit.

docs/simulators/OpenFOAM.md
Pbarbosa92: The default is spot.

docs/simulators/OpenFOAM.md
Pbarbosa92: | Machine Configuration             | n_vCPUs/total           | Threads per core       | Time (h:min:sec)|
|-----------------------------------|-------------------------|------------------------|----------------|
| c3d-highcpu-360 x2 (Mpi Cluster)  | 360/720                 | 2                      | 19:10:09       |
| c3d-highcpu-360                   | 250/360                 | 2                      | 39:58:37       |
| c3d-highcpu-360                   | 180/360                 | 2                      | 40:20:00       |
| c3d-highcpu-360                   | 180/180                 | 1                      | 40:47:14       |
| c3d-highcpu-60 x6 (Mpi Cluster)   | 180/360                 | 2                      | 41:23:46       |
| c3d-highcpu-360                   | 360/360                 | 2                      | 42:17:59       |

Just to save the info here too

docs/simulators/OpenFOAM.md
sarmento: Is this the only difference? Then I would refer that, and show which other line changes. Otherwise there is a lot of repetition.

docs/simulators/OpenFOAM.md
sarmento: @Pbarbosa92 please own this issue of running arbitrary commands and check with @hpenedones what is the best options. Ideally, users would be able to run their own .sh files with minor or no changes, instead of having to issue the commands like this. Please talk with Hugo about this. Let's try to reduce the barrier to adoption of OpenFoam users.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,286 @@ task.download_outputs()
...
+You can adjust this depending on the available resources or your project's needs.
+   
+3. **Commands**: The list of `commands` involves several key OpenFOAM tasks
+like copying necessary configuration files, running mesh generation, and
+parallelizing the process using `decomposePar` and `snappyHexMesh`.
+
+4. **Parallelization**: We use `n_vcpus=180` in the simulation, which will
+allow OpenFOAM to run in parallel, greatly speeding up the computation.
+
+### Running the Simulation
```
sarmento, line 319: @Pbarbosa92 for lists of commands, we would love to know which command is being executed at each time. It could show up on the console or on the CLI. Please own this UX issue and chat with @ClaudiaRSilva about this.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,286 @@ task.download_outputs()
...
+2. **Machine Configuration**: We are using a powerful `c3d-highcpu-360`
+machine with 360 vCPUs to ensure that the computation proceeds efficiently.
+You can adjust this depending on the available resources or your project's needs.
+   
+3. **Commands**: The list of `commands` involves several key OpenFOAM tasks
+like copying necessary configuration files, running mesh generation, and
+parallelizing the process using `decomposePar` and `snappyHexMesh`.
+
+4. **Parallelization**: We use `n_vcpus=180` in the simulation, which will
+allow OpenFOAM to run in parallel, greatly speeding up the computation.
```
sarmento, line 317: But why not 360? We need to explain this...

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,286 @@ task.download_outputs()
...
+
+### Running the Simulation
+
+Running the script above will take some time, as OpenFOAM tasks like
+`snappyHexMesh` and `checkMesh` can be computationally intensive, especially
+with large meshes. The `task.wait()` command will block until the simulation
+is complete, after which the output files will be available.
+
```
sarmento, line 325: This is a good place to say how long this took and how much data was generated.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,286 @@ task.download_outputs()
...
+You can now perform post-processing on the results, just as you would if you had
+run the simulation locally.
+
+### Utilizing an MPI Cluster for Enhanced Simulation Performance
+
+The initial simulation conducted on a single `c3d-highcpu-360` machine took 40
+hours and 20 minutes to complete. To expedite this process, we opted to deploy
+the simulation across two `c3d-highcpu-360` machines in an MPI cluster configuration.
+This strategic change reduced the simulation time significantly to just 19 hours
```
sarmento, line 369: Tell this only in the end. Don't be a spoiler... :)

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,286 @@ task.download_outputs()
...
+and 10 minutes.
+
+Leveraging Inductiva's Python library, scaling the simulation resources is
+straightforward. You simply need to modify the instantiation of the machine group,
+replacing it with an MPI cluster setup, and adjust the `n_vcpus` parameter accordingly.
+Below is an example of the code implementation:
+
+```python notest
+auto_terminate = (datetime.datetime.now() + datetime.timedelta(days=6)).replace(tzinfo=pytz.utc)
```
sarmento, line 378: Wait.. Why do we need this? This is super confusing....

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,286 @@ task.download_outputs()
...
+mpi_cluster = inductiva.resources.MPICluster(
+    machine_type="c3d-highcpu-360",
+    data_disk_gb=300,
+    num_machines=2,
+    auto_terminate_ts=auto_terminate
```
sarmento, line 383: This should not be needed. Please submit and issue to get the desired behaviour. Please don't let these things accumulate (plus, this ends in less than 24hours so it shouldn't even be an issue)

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,286 @@ task.download_outputs()
...
+)
+
+mpi_cluster.start()
+
+...
+# Execute the simulation
+task = openfoam.run(
+    input_dir=input_dir,
+    commands=commands,
+    n_vcpus=360,
```
sarmento, line 393: comment that now we are running on twice as many threads.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,286 @@ task.download_outputs()
...
+    use_hwthread=True,
+    on=mpi_cluster
+)
+```
+
```
sarmento, line 398: talk about the times and the number of files/amount of data produced only here.

docs/simulators/OpenFOAM.md
```diff
@@ -126,6 +126,286 @@ task.download_outputs()
...
+### Conclusion
+
+This setup demonstrates how to leverage cloud resources to run computationally
+heavy OpenFOAM simulations, specifically using a case from the ExaFOAM benchmarks.
+The cloud-based environment enables parallel execution on high-performance
+machines, drastically reducing computation time. Once the simulation completes,
```
sarmento, line 404: You can be more specific. It's not just drastically reducing. It's halving. In fact is actually even better.

docs/simulators/OpenFOAM.md
Pbarbosa92: We have that on the stdout and stderr.
We have the command and the respective output.

Were do you want this information to show ?

docs/simulators/OpenFOAM.md
Pbarbosa92: The issue already exists https://github.com/inductiva/tasks/issues/389
I can remove this from here, in the expectation that in the near future this will be fixed (and you are correct, for this case we would not need it but i didn't know it at the time)

docs/simulators/OpenFOAM.md
sarmento: It should show as output of the method wait(). It should be interactive.

"Executing command XXXXX".

Ideally we would have that showing on the console as the task updates.

docs/simulators/OpenFOAM.md
sarmento: Please remove it here. Let's assume it will be fixed somewhere else.

docs/simulators/OpenFOAM.md
Pbarbosa92: I believe this would be doable. A task could have a field called `command` that saves the current command being executed at each time (the executer tracker would need to update this field).

I will go further into this ( i will check with the back end the effort needed to implement this and will create and maybe implement the changes needed)

docs/simulators/OpenFOAM.md
```diff
@@ -230,7 +230,10 @@ Below is the Python code to configure and run the OpenFOAM simulation:
 import inductiva
 import os
 
-machine_group = inductiva.resources.MachineGroup(machine_type="c3d-highcpu-360")
+machine_group = inductiva.resources.MachineGroup(
+                    machine_type="c3d-highcpu-360",
+                    spot=True,
+                    )
```
sarmento, line 236: This can be pushed up:

```
spot=True)
```

sarmento (approved):

PR #1611, opened 2024-09-30
Delete task metadata json

HenriquePreto:
Removes the generation of the task metadata JSON file created after running a simulation. Updates the documentation, tests, and benchmark module.

This PR also removes the 'extra_metadata' argument from simulations, as it is not used.

Closes https://github.com/inductiva/tasks/issues/339


inductiva/_benchmarks/methods.py
```diff
@@ -58,21 +59,25 @@ def _analyze_task(task: Task, report: Dict[str, Any], skipped_tasks: List[Task],
             "machine_count": resource.num_machines,
             "tasks": []
         }
+    metadata_filename = "input.json"
+    inputs_path = task.download_inputs(filenames=[metadata_filename])
+    metadata_path = inputs_path.joinpath(metadata_filename)
+    with open(metadata_path, mode="r", encoding="utf-8") as metadata_file:
+        metadata_content = json.load(metadata_file)
```
Pbarbosa92, line 66: This works but I don't like this approach. I'm fine with having the method download_inputs, but we should not be using it for this. We should call an api endpoint to get this metadata for that task.

Pbarbosa92 (approved):
I would prefer if we got the metadata information from a endpoint instead of downloading files.

inductiva/_benchmarks/methods.py
HenriquePreto: Sure, we could implement a method in the class 'Task'. but I think it's better to leave that for a separate PR.

PR #1612, opened 2024-09-30
Added cost to task info

Pbarbosa92:
https://github.com/inductiva/tasks/issues/396

This PR adds the task cost to the info.

Running the cli command `inductiva task info <task_id>` should now also print the task cost.
This also works for python scripts with the task.info or task.get_info().

The task cost will only be displayed if there is a cost to be shown.

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -96,6 +96,7 @@ class TaskInfo:
         self.computation_start_time = None
         self.computation_end_time = None
         self.end_time = None
+        self.cost = None
         self.time_metrics = self.TimeMetrics()
         self.data_metrics = self.DataMetrics()
         self._kwargs = kwargs
@@ -238,6 +239,7 @@ class TaskInfo:
         table_str += f"\n{wall_time_table}"
         table_str += f"\nTime breakdown:\n{time_metrics_table}"
         table_str += f"\nData:\n{data_metrics_table}\n"
+        table_str += f"\nCost ($US): {self.cost}\n"
 
         return table_str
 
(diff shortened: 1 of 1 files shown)


inductiva/tasks/task.py
```diff
@@ -238,6 +239,8 @@ def __str__(self):
         table_str += f"\n{wall_time_table}"
         table_str += f"\nTime breakdown:\n{time_metrics_table}"
         table_str += f"\nData:\n{data_metrics_table}\n"
+        if self.cost:
```
HenriquePreto, line 242: Should we include the same null check for the 'status' attribute?

Pbarbosa92: A task will always have a status.

HenriquePreto: In theory, yes. But, the constructor of the 'TaskInfo' class sets the 'status' attribute to None. It's just a detail for consistency.

Pbarbosa92: I only added this check because a task will not have a cost until its finished. Without the "if" we would have cost: None until the task was finished.

For status we don't need to do the if because it will never be None, and we want to print the status every time. Adding an if would just make the code harder to read without any advantage.

HenriquePreto (approved):

HenriquePreto:
Using the cli i was the following commands:
```
inductiva tasks info 6hrps61hov6fhczsiivtgq14f
inductiva tasks info 6hrps61hov6fhczsiivtgq140
```
It worked as expected, printing:
````
...
Cost ($US): 0.00056
```

PR #1620, opened 2024-10-07
Remove function 'can_start_resource' from benchmarks

HenriquePreto:
This PR contributes to the refactoring of the benchmarks module by removing the 'can_start_resource' function from the module. Additionally, it updates the tests that used this function.


inductiva/resources/machines_base.py
```diff
@@ -285,7 +285,7 @@ def from_api_response(cls, resp: dict):
 
         return machine_group
 
-    def _can_start_resource(self) -> bool:
```
Pbarbosa92, line 288: why change the method to public?

Pbarbosa92: This function is only invoked during the start process and is not intended for user access. Therefore, it makes sense to keep it private. It was previously set as private and successfully passed the linter tests. I recommend reverting it to private.

HenriquePreto: I don't see why this function shouldn't be accessible to users. Also, the method was not being tested previously, at all -- it was another one imported from the _benchmarks module. I can try again, but I believe reverting it to private won't pass the linter tests.

Pbarbosa92: > I don't see why this function shouldn't be accessible to users. Also, the method was not being tested previously, at all -- it was another one imported from the _benchmarks module. I can try again, but I believe reverting it to private won't pass the linter tests.

This behaviour will be controlled by a flag `wait_on_pending_quota` in the start method. The user just need to do mg.start(wait_on_pending_quota=True) instead of doing 
```
if mg.can_start_resource:
    mg.start()
```

As for testing. That will be hard. We need to mock the object creation to prevent the machinegroup being created in the back end. We then need to mock the get_quotas. After that we can test the method.

Pbarbosa92: I can add the tests if you want.

HenriquePreto: Please check the final version of this PR, which includes tests using those mock objects.

HenriquePreto: > > I don't see why this function shouldn't be accessible to users. Also, the method was not being tested previously, at all -- it was another one imported from the _benchmarks module. I can try again, but I believe reverting it to private won't pass the linter tests.
> 
> This behaviour will be controlled by a flag `wait_on_pending_quota` in the start method. The user just need to do mg.start(wait_on_pending_quota=True) instead of doing
> 
> ```
> if mg.can_start_resource:
>     mg.start()
> ```
> 
> As for testing. That will be hard. We need to mock the object creation to prevent the machinegroup being created in the back end. We then need to mock the get_quotas. After that we can test the method.

Please check the final version of this PR, which includes tests using those mock objects.

HenriquePreto: > I can add the tests if you want.

It has already been done, thanks.

inductiva/tests/benchmarks/test_benchmarks.py
```diff
@@ -3,9 +3,9 @@
```
Pbarbosa92, line 3: Move the tests to `test_machines.py`

Pbarbosa92 (approved):
Just need to move the tests to the correct place now. Apart from that LGTM!

inductiva/tests/benchmarks/test_benchmarks.py
HenriquePreto: Done!

Pbarbosa92: I believe we can merge since the code was already tested on the original PR. This PR was just a refactor (and to delete a method that I forgot to remove)

PR #1626, opened 2024-10-09
Fix simulator name

Pbarbosa92:
This PR fixes amrwind simulator name for the BE. They expect all lower case.


HenriquePreto (approved):
LGTM!

PR #1627, opened 2024-10-09
Add ram information when registering Machine Group

rodrigotuna:
This PR adds the size of the RAM of the Machine Group being registered to the information printed to the user.

### DIFF
--- a/inductiva/resources/machines_base.py
+++ b/inductiva/resources/machines_base.py
@@ -250,6 +250,7 @@ class BaseMachineGroup(ABC):
             body.get("idle_seconds"))
         self._auto_terminate_ts = self._iso_to_datetime(
             body.get("auto_terminate_ts"))
+        self.ram_gb = body.get("ram_gb")
         self._log_machine_group_info()
 
     @abstractmethod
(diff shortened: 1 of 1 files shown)


inductiva/resources/machines_base.py
```diff
@@ -250,6 +250,7 @@ def _register_machine_group(self, **kwargs):
             body.get("idle_seconds"))
         self._auto_terminate_ts = self._iso_to_datetime(
             body.get("auto_terminate_ts"))
+        self.ram_gb = body.get("ram_gb")
```
Pbarbosa92, line 253: Is this total ram or ram per core ?

Pbarbosa92 (changes requested):

inductiva/resources/machines_base.py
rodrigotuna: Total ram

Pbarbosa92: Then i would recommend changing it to self.total_ram_gb. Since gcp has ram per vCpus, we need to make a distinction.

Pbarbosa92 (approved):

PR #1628, opened 2024-10-10
Implement 'Benchmark' Class

HenriquePreto:
This pull request adds the Benchmark class, which aims to improve the usability of our benchmarking functionality and eventually replace the current benchmarking library.

## Examples

- **Example 1)**: This benchmark case runs the OpenFAST simulator with configurations on two machine types: ```c2-standard-4``` with 2 vCPUs and ```c2-standard-8``` with 4 vCPUs, repeating twice each run for better reliability.

```python
import datetime
import json
import inductiva
from inductiva._benchmarks.benchmark import Benchmark
from inductiva.resources.machines import MachineGroup

input_dir = inductiva.utils.download_from_url(
            url="https://storage.googleapis.com/inductiva-api-demo-files/"
                "openfast-input-example.zip",
            unzip=True)
openfast = inductiva.simulators.OpenFAST()
max_idle_time = datetime.timedelta(minutes=1)

Benchmark(name='a-benchmark-demo') \
    .add_run(simulator=openfast,
             input_dir=input_dir,
             on=MachineGroup("c2-standard-4", max_idle_time=max_idle_time),
             commands=["openfast IEA-15-240-RWT-Monopile.fst"],
             n_vcpus=2) \
    .add_run(simulator=openfast,
             input_dir=input_dir,
             on=MachineGroup("c2-standard-8", max_idle_time=max_idle_time),
             commands=["openfast IEA-15-240-RWT-Monopile.fst"],
             n_vcpus=4) \
    .run(num_repeats=2) \
    .wait()
```

- **Example 2)**: This script benchmarks the OpenFAST simulator using the ```c2``` machine family, excluding spot instances.

```python
# filter machine groups by family
max_idle_time = datetime.timedelta(minutes=1)
available_machine_types = get_available_machine_types(machine_family='c2')
machine_groups = [
    MachineGroup(m.machine_type, max_idle_time=max_idle_time)
    for m in available_machine_types
    if not m.spot
]


# benchmark execution
benchmark = Benchmark(name='benchmark-by-family-c2')
for machine_group in machine_groups:
    benchmark.add_run(simulator=openfast,
                      input_dir=input_dir,
                      commands=["openfast IEA-15-240-RWT-Monopile.fst"],
                      on=machine_group)
benchmark.run().wait()
```

- **Example 3)**: This script benchmarks the ArmWind simulator using the ```c2``` machine family, exporting the performance metrics in the end.

```python
from inductiva._benchmarks.benchmark import Benchmark, ExportFormat
from inductiva.simulators import AmrWind
from inductiva.utils import download_from_url
from inductiva.resources import MachineGroup

input_dir = download_from_url(
    "amr-wind-input-example.zip",
    unzip=True)

Benchmark(name="Tutorial-Benchmark-AmrWind-on-c2-standard-Machines") \
    .set_default(simulator=AmrWind(),
                 input_dir=input_dir,
                 sim_config_filename="abl_amd_wenoz.inp") \
    .add_run(on=MachineGroup("c2-standard-4"), n_vcpus=4) \
    .add_run(on=MachineGroup("c2-standard-8"), n_vcpus=8) \
    .add_run(on=MachineGroup("c2-standard-16"), n_vcpus=16) \
    .add_run(on=MachineGroup("c2-standard-30"), n_vcpus=30) \
    .add_run(on=MachineGroup("c2-standard-60"), n_vcpus=60) \
    .run(num_repeats=2) \
    .wait() \
    .export(fmt="csv")
```


hpenedones:
I think we could already get rid of the _benchmark hidden package and just have:

```python
from inductiva.benchmarks import Benchmark
```
instead of 

```python
from inductiva._benchmarks.benchmark import Benchmark
```

Right, @Pbarbosa92 ?

inductiva/_benchmarks/benchmark.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Returns:
+            dict: A dictionary containing the results of each task, organized 
+            by virtual machine type.
+        """
+        results = {}
+        tasks = self.get_tasks()
+        for task in tasks:
+            task_info = task.info
+            tmp1, tmp2, *_ = task_info.executer.vm_name.split("-")
```
hpenedones, line 97: I think that here there is some hardcoded assumption that will only work for GCP vm names -- can we encapsulate this into some function? And see if what we need to extract also works for other machine groups? We will want to use this same Benchmark class to also include machine groups of local machines, for example...

inductiva/_benchmarks/benchmark.py
```diff
@@ -0,0 +1,119 @@
...
+from inductiva import types
+from inductiva.simulators.simulator import Simulator
+from inductiva.projects.project import Project
+from inductiva.resources import machine_groups
+
+
+class Benchmark(Project):
+    """Represents the benchmark runner."""
+
+    def __init__(self, name: str, append: bool = True):
```
Pbarbosa92, line 13: Should the default value for `append` be set to `True`?
For projects we are using append False by default to make the action of adding things to a project be explicit so the user does not add anything by mistake.

inductiva/_benchmarks/benchmark.py
```diff
@@ -0,0 +1,119 @@
...
+        Args:
+            num_repeats (int): The number of times to repeat each simulation
+            run (default is 2).
+
+        Returns:
+            Self: The current instance for method chaining.
+        """
+        with self:
+            for simulator, input_dir, machine_group, kwargs in self.runs:
+                machine_group.start(wait_on_pending_quota=True)
```
Pbarbosa92, line 65: Allow the configuration of wait_on_pending_quota on the Benchmark constructor. You can make it True by default.

inductiva/_benchmarks/benchmark.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Returns:
+            Self: The current instance for method chaining.
+        """
+        tasks = self.get_tasks()
+        for task in tasks:
+            task.wait(download_std_on_completion=False)
+        return self
+
+    def to_dict(self):
```
Pbarbosa92, line 85: What would happen if I called to_dict before a benchmark finishes running?
Maybe we should only collect info about tasks that are successful?  What about tasks that are still running?

We need to make a decision here. To_dict can only be called if all tasks are finished (in terminal state).
What if some tasks fail?

inductiva/_benchmarks/benchmark.py
HenriquePreto: This was removed in the most recent PR related to benchmarks, which focused on exporting and gathering benchmarking metrics and data.

HenriquePreto:
> I think we could already get rid of the _benchmark hidden package and just have:
> 
> ```python
> from inductiva.benchmarks import Benchmark
> ```
> 
> instead of
> 
> ```python
> from inductiva._benchmarks.benchmark import Benchmark
> ```
> 
> Right, @Pbarbosa92 ?



> I think we could already get rid of the _benchmark hidden package and just have:
> 
> ```python
> from inductiva.benchmarks import Benchmark
> ```
> 
> instead of
> 
> ```python
> from inductiva._benchmarks.benchmark import Benchmark
> ```
> 
> Right, @Pbarbosa92 ?

I'm not so sure. The package is named "_benchmarks" (folder) and the module is named "benchmark" (file). But we could improve the naming.

inductiva/_benchmarks/benchmark.py
HenriquePreto: In this case, I think it should. What should happen if a user runs the same benchmark file twice? Currently, it includes additional runs to the benchmark (e.g., makes the results / performance metrics more reliable). This behavior is due to ```append``` being set to ```True``` by default.

inductiva/_benchmarks/benchmark.py
HenriquePreto: I think we should keep this version of the Benchmark API as simple as possible to highlight only the most important parameters needed to create, configure, run, and export a benchmark.

We can consider including additional parameters in future versions, for example those that could also be passed to the ```wait``` method.

What do you think?

inductiva/_benchmarks/benchmark.py
HenriquePreto: Currently, it should fail -- tasks that are not in a terminal state don't have ```info```, right? Even so, the ```wait``` method was implemented to prevent this issue from happening.

I think we should collect information about tasks that are in a terminal state, even if there's a failure.

By the way, this method has been replaced by ```gather_metrics``` and ```gather_detailed_data``` in the most recent PR.

inductiva/_benchmarks/benchmark.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        Waits for all running tasks to complete.
+
+        Returns:
+            Self: The current instance for method chaining.
+        """
+        tasks = self.get_tasks()
+        for task in tasks:
+            task.wait(download_std_on_completion=False)
+        return self
```
bgregoinductiva, line 141: I'll just reinforce my suggestion that all machine groups associated with the finished tasks should be terminated in this method.

bgregoinductiva: Yes, if a machine is terminated twice the API returns an error. But could that happen in this scenario?

Pbarbosa92: Why don't we just start all machine groups with a max idle time of 1 minute?

HenriquePreto: Because that's user's decision -- the argument ```max_idle_time``` is passed in the ctor of ```MachineGroup```

HenriquePreto: > Yes, if a machine is terminated twice the API returns an error. But could that happen in this scenario?

I'm thinking that users might be surprised by an unexpected exception raised when calling ```wait``` due to its side effect of invoking ```terminate```. Because, in contrast, calling ```run``` on the same machine only warns the user that the machine has already started. 

Maybe, it might be more appropriate to also pass a boolean parameter named ```terminate``` to the ```wait``` method (as well as to the```run``` method passing a ```start``` parameter) to provide better control over the method's behavior.

bgregoinductiva: > > Yes, if a machine is terminated twice the API returns an error. But could that happen in this scenario?
> 
> I'm thinking that users might be surprised by an unexpected exception raised when calling `wait` due to its side effect of invoking `terminate`. Because, in contrast, calling `run` on the same machine only warns the user that the machine has already started.
> 
> Maybe, it might be more appropriate to also pass a boolean parameter named `terminate` to the `wait` method (as well as to the`run` method passing a `start` parameter) to provide better control over the method's behavior.

I don't agree. The exception shouldn't happen unless the user tries to terminate the machine groups manually while the benchmark is running. The goal of introducing the termination is to avoid wasting resources. Most users won't be aware the machine groups will still be running after terminating the benchmark.

bgregoinductiva: Given the current interface, it would be weird to overwrite attributes of the machine group. But I understand the point. Alternatively we could provide the `terminate` method and print a message after starting the machine groups to warn the users that the machines will continue to run after the tasks are completed.

HenriquePreto: > > > Yes, if a machine is terminated twice the API returns an error. But could that happen in this scenario?
> > 
> > 
> > I'm thinking that users might be surprised by an unexpected exception raised when calling `wait` due to its side effect of invoking `terminate`. Because, in contrast, calling `run` on the same machine only warns the user that the machine has already started.
> > Maybe, it might be more appropriate to also pass a boolean parameter named `terminate` to the `wait` method (as well as to the`run` method passing a `start` parameter) to provide better control over the method's behavior.
> 
> I don't agree. The exception shouldn't happen unless the user tries to terminate the machine groups manually while the benchmark is running. The goal of introducing the automatic termination is to avoid wasting resources. Most users won't be aware that the machine groups will still be running after terminating the benchmark.

I think the exception also happens when the user attempts to call the ```wait``` method twice, like this: ```benchmark.wait().wait()``` -- this is not a problem when calling ```task.wait().wait()```.

But, I agree that most users may not be aware of this and the machine groups should be automatically shut down.

bgregoinductiva (approved):

PR #1634, opened 2024-10-16
Fix quantum espresso location

Pbarbosa92:
This PR fixes the location of quantum espresso to match the current paths of the other simulators.


Pbarbosa92:
@maiyahershey I've added you here so you can add that intro like you did for the other simulators

maiyahershey (approved):

PR #1636, opened 2024-10-18
Fix websocket client

bgregoinductiva:


hpenedones (approved):
Uau, that was a truly surgical intervention! Nice.

Pbarbosa92 (approved):
So, what did change to break this ?

Pbarbosa92:
I just tested this. We manage to connect now. But we still get no logs.

Maybe it has something to do with the computation started status.

bgregoinductiva:
> So, what did change to break this ?

Unfortunately we are not sure. Somehow the URL encode was not being done correctly.

PR #1647, opened 2024-10-23
Reference existing files from bucket

rcvalerio:
closes https://github.com/inductiva/tasks/issues/403

Usage:
```python
import inductiva

inductiva.storage.upload_from_url(
    url=
    "https://storage.googleapis.com/inductiva-api-demo-files/test_assets/files.zip",
    remote_dir="example_bucket",
    unzip=False,
    overwrite=True,
)

inductiva.storage.upload(
    local_path="gromacs-input-example/",
    remote_dir="gromacs_bucket",
    overwrite=True,
)

task = gromacs.run(input_dir="empty_folder",
                   commands=commands,
                   on=machine,
                   remote_assets=["gromacs_bucket"])


task = gromacs.run(input_dir="empty_folder",
                   commands=commands,
                   on=machine,
                   remote_assets=["gromacs_bucket/file1.txt", "gromacs_bucket/file2.txt"])


inductiva.storage.remove_workspace(remote_dir="gromacs_bucket")

# delete single file
inductiva.storage.remove_workspace(remote_dir="gromacs_bucket", file_name="file1.txt")

```

@sarmento


hpenedones:
@sarmento : would you mind commenting here on the method or argument names, such that @rcvalerio can make any rename before merging?

inductiva/storage/storage.py
```diff
@@ -118,3 +129,154 @@ def _print_contents_table(contents):
...
+
+
+def upload_from_url(
+    url: str,
+    remote_dir: str,
+    file_name: str = None,
+    unzip: bool = False,
+    overwrite: bool = True,
+):
+    """Upload a file from a URL to the user workspace."""
```
Pbarbosa92, line 141: Add arguments to this docstring (improve description).

inductiva/storage/storage.py
```diff
@@ -118,3 +129,154 @@ def _print_contents_table(contents):
...
+    else:
+        logging.info("File upload failed.")
+
+
+def upload(
+    local_path: str,
+    remote_dir: str = "default",
+    overwrite: bool = False,
+):
+    """Upload local files to the user workspace."""
```
Pbarbosa92, line 169: Add arguments to this docstring (improve description).

inductiva/storage/storage.py
```diff
@@ -118,3 +129,154 @@ def _print_contents_table(contents):
...
+            "folder_name": remote_dir,
+        },
+    )
+    if contents.response.status == 200:
+        logging.info("File is being uploaded...")
+    else:
+        logging.info("File upload failed.")
+
+
+def upload(
```
Pbarbosa92, line 164: This method has a lot in common with the method to upload the input files to a task `upload_input`.
Lets refactor the code a bit to avoid all the code repetition.

inductiva/storage/storage.py
```diff
@@ -118,3 +129,154 @@ def _print_contents_table(contents):
...
+    input_zip_path = _zip_file_or_folder(local_path)
+
+    zip_file_size = os.path.getsize(input_zip_path)
+    logging.info("Input archive size: %s",
+                 format_utils.bytes_formatter(zip_file_size))
+
+    logging.info("Uploading input...")
+    try:
+
+        api_response = api_instance.get_upload_url(
```
Pbarbosa92, line 181: If this is the only thing that changes from `upload_input` we can create a method that gets the endpoint as an argument and reuse just passing a new endpoint.

Pbarbosa92 (changes requested):
There a small refactor we need to do to avoid duplication code.

sarmento:
A few quick comments / questions (I don't have the definite answer):



``

inductiva.storage.upload(
    local_path="gromacs-input-example/",
    remote_dir="gromacs_bucket",
    overwrite=True,
)

````

Not sure why we are using "local_path"? Is it because we can also upload single files? If we only allow directories it should be "local_dir". In fact, should we only allow uploading directories? If not (which is valid), can we merge "upload" and "upload_from_url" so that they behave the exact same way, and the only thing we change is the fact that we can have the source also being an url.

```
inductiva.storage.upload(
    source=
    "https://storage.googleapis.com/inductiva-api-demo-files/test_assets/files.zip",
    dest="example_bucket",
    unzip=False,
    overwrite=True,
)
```

I understand that this may be difficult, since remote url need to be a file, right? We can't download the contents of a bucket passed as input, or can we?
```
task = gromacs.run(input_dir="empty_folder",
                   commands=commands,
                   on=machine,
                   remote_assets=["workspace/gromacs_bucket"])
```

Maybe we could be more consistent with naming. These are all directories (or not?):

* "remote_assets" ---> "remote_dir" (and allow a list) 

Unless these could also be an URL....

Maybe we want to start moving in the direction where we still allow input_dir for a moment but we want to have local_input_dir as the parameter moving fwd. This would be easier to understand.

Also, we may want to allow the users to only have to fill in one of these parameters.
This should be possible:

```
task = gromacs.run(
                   commands=commands,
                   on=machine,
                   remote_dir=["workspace/gromacs_bucket"])
```

Pbarbosa92:
> A few quick comments / questions (I don't have the definite answer):
> 
> ```
> 
> inductiva.storage.upload(
>     local_path="gromacs-input-example/",
>     remote_dir="gromacs_bucket",
>     overwrite=True,
> )
> ```
> 
> Not sure why we are using "local_path"? Is it because we can also upload single files? If we only allow directories it should be "local_dir". In fact, should we only allow uploading directories? If not (which is valid), can we merge "upload" and "upload_from_url" so that they behave the exact same way, and the only thing we change is the fact that we can have the source also being an url.
> 
> ```
> inductiva.storage.upload(
>     source=
>     "https://storage.googleapis.com/inductiva-api-demo-files/test_assets/files.zip",
>     dest="example_bucket",
>     unzip=False,
>     overwrite=True,
> )
> ```
> 
> I understand that this may be difficult, since remote url need to be a file, right? We can't download the contents of a bucket passed as input, or can we?
> 
> ```
> task = gromacs.run(input_dir="empty_folder",
>                    commands=commands,
>                    on=machine,
>                    remote_assets=["workspace/gromacs_bucket"])
> ```
> 
> Maybe we could be more consistent with naming. These are all directories (or not?):
> 
> * "remote_assets" ---> "remote_dir" (and allow a list)
> 
> Unless these could also be an URL....
> 
> Maybe we want to start moving in the direction where we still allow input_dir for a moment but we want to have local_input_dir as the parameter moving fwd. This would be easier to understand.
> 
> Also, we may want to allow the users to only have to fill in one of these parameters. This should be possible:
> 
> ```
> task = gromacs.run(
>                    commands=commands,
>                    on=machine,
>                    remote_dir=["workspace/gromacs_bucket"])
> ```

Merging both methods into a single once can be a bit tricky. How do identify if the source is from a local path or a url? Do we look for http? some urls might not have http and things like that.

I would say we move forward with both methods.

Even the fact that we only allow for the upload of folders should be enough for now and we can add the feature to select single files at a later date. Since this is a requested feature we should just release it in working order and iterate over time.

rcvalerio:
> > ```python
> > import inductiva
> > 
> > 
> > inductiva.storage.upload(
> >     local_path="gromacs-input-example/",
> >     remote_dir="gromacs_bucket",
> >     overwrite=True,
> > )
> > ```
> 
> For me, this isn't very clear if we're copying `gromacs-input-example` to inside `gromacs_bucket`, or if we're copying `gromacs-input-example`'s files to inside `gromacs_bucket`?

We are copying `gromacs-input-example` files to `gromacs_bucket`

inductiva/storage/storage.py
rcvalerio: I refactored with a new method that accepts the endpoints and is reused by tasks and workspaces

Pbarbosa92 (approved):

PR #1651, opened 2024-10-24
fix-get-costs-for-external-users

Pbarbosa92:
get_costs is not called inside try catch to prevent problems with external users


HenriquePreto (approved):
LGTM

Pbarbosa92:
External users should see 
```
■ Estimated Costs (current month):
	Not available for external users.
```
when using user info.
If they call the get_costs directly they will get an ApiException with an error message

PR #1659, opened 2024-10-28
Export benchmarking metrics

HenriquePreto:
This PR introduces the ```gather_metrics``` and ```export``` methods for the benchmarking API. Users can collect benchmarking performance metrics with the ```gather_metrics``` method. For detailed information — including additional performance metrics and task metadata — users can use the ```gather_detailed_data```. Additionally, benchmarking performance metrics can be easily exported to CSV or JSON formats using the ```export``` method.

```python
from inductiva._benchmarks.benchmark import Benchmark, ExportFormat

# Initialize a previously executed benchmark.
benchmark = Benchmark(name="benchmark-family-c2")

# ---

# Collect benchmarking metrics (i.e., time, cost)
performance_metrics = benchmark.gather_metrics()

# Collect additional benchmarking data (e.g., additional metrics and input metadata)
detailed_data = benchmark.gather_detailed_data()

# ---

# Export benchmarking metrics (i.e., time, cost) to a JSON file
benchmark.export(format=ExportFormat.JSON)

# Export benchmarking metrics (i.e., time, cost) to a CSV file named "my_benchmark.csv"
benchmark.export(format=ExportFormat.CSV, filename="my_benchmark.csv")
```


inductiva/_benchmarks/benchmark.py
```diff
@@ -133,38 +141,78 @@ def wait(self) -> Self:
...
+                    writer.writerows(csv_content)
+            case _:
+                raise ValueError(f"Unsupported export format: {format}")
+
+    def gather_data(self, verbose: bool = False) -> dict:
+        """
+        Gathers information, including metrics and metadata, from all tasks.
+
+        Args:
+            verbose (bool): A boolean indicating whether to include additional
```
Pbarbosa92, line 184: Usually verbose is used to enable/disable prints. In this case verbose is controlling the information returned. We should pick another name for this. Maybe `simulation_arguments` or `gather_simulation_arguments`

inductiva/_benchmarks/benchmark.py
```diff
@@ -133,38 +141,78 @@ def wait(self) -> Self:
...
-            results[vm_type]["tasks"].append({
-                "info": task_info.to_dict(),
-                "metadata": metadata_content,
-            })
-        return results
+            info = task.get_info()
+            vm_type = info.executer.vm_type
+            data.setdefault(vm_type, [])
+            gather_function = _gather_verbose if verbose else _gather_summary
+            gather_function(data, info, vm_type)
```
Pbarbosa92, line 217: I'm not a big fan of using parameters as input/output.
Since the _gather functions are internal and dont have docstrings is not that clear what is going to happen.

Maybe make the gather_funtion return one record and append that in the main function.

This is just personal preference.

Pbarbosa92 (changes requested):
Looks good, just one relevant comment when it comes to the verbose argument.

inductiva/_benchmarks/benchmark.py
HenriquePreto: I'm not sure I fully understand the part about 'using parameters as input/output'. Could you clarify a bit more?

Pbarbosa92: Your data is an input of the method gather_function but it also works as an output since you modify data inside the method.
You can do that
```python
#Data gets modified inside gather_function
gather_function(data, info, vm_type)
```
Or
```python
#you have your data
new_data = gather_function(info)

data[vm_type].append(new_data)
```

I just prefer the second one since inputs and outputs are clearly defined. But its like I said before. It's just personal preference.

HenriquePreto: I get what you’re saying. This function is being refactored.

inductiva/_benchmarks/benchmark.py
```diff
@@ -133,38 +141,85 @@ def wait(self) -> Self:
...
+
+        Args:
+            format (str): The format to export the results in.
+            filename (Optional[str]): The name of the output file to save the
+                exported results. Defaults to the benchmark's name if not
+                provided.
+        """
+        metrics = self.gather_metrics()
+        filename = filename or f"{self.name}.{format.value}"
+        match format:
```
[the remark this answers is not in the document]
HenriquePreto: I believe that supporting Python >= 3.8 doesn't prevent us from using features from Python 3.10 or newer versions. I suggested to @Pbarbosa92 that we could upgrade the Python version to 3.12 in our GitHub workflows (https://github.com/inductiva/inductiva/blob/e8ca4b6d2a9e50e25e58e6028843429ba1d0592a/.github/workflows/python-package.yml#L19). Would it work?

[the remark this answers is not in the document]
HenriquePreto: Ah, got it! Thanks

HenriquePreto:
> > ```python
> > from inductiva._benchmarks.benchmark import Benchmark, ExportFormat
> > ```
> 
> This import should be easier. Could it be just `from inductiva.benchmarks import Benchmark, ExportFormat`?

I agree. But '_benchmarks' is a package and 'benchmark' is a module -- I would need to convert '_benchmarks' into a module.

inductiva/_benchmarks/benchmark.py
```diff
@@ -133,38 +141,85 @@ def wait(self) -> Self:
...
+    def export(
+        self,
+        format: ExportFormat = ExportFormat.JSON,
+        filename: Optional[str] = None,
+    ):
+        """
+        Exports the benchmark performance metrics in the specified format.
+
+        Args:
+            format (str): The format to export the results in.
```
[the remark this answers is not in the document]
HenriquePreto: Thanks, the docstring was fixed [here](https://github.com/inductiva/inductiva/pull/1664/commits/c669455c123af738a93fd9b29508c707f2522fc1).

The ```ExportFormat``` enum was implemented to enable autocompletion for users. But, I agree that passing a string is more user-friendly.

PR #1663, opened 2024-10-29
Sphinx redirects

Pbarbosa92:
This PR adds lots of redirects to old urls that are giving 404 on the google search console.

Some redirects are temporary and need to be changed in the future like the ones too https://tutorials.inductiva.ai/intro_to_api/shared_dedicated_resources.html (that will be moved to docs).


maiyahershey:
This is great Paulo! Redirects are great because this will not interrupt our users browsing experience and it's also a great way to tell search engines about the new location of the page. Google should deindex these moved pages when it tries to crawl again soon and should be fixed! 
Thanks so much for this. 

We could add this flow as best practice for the future when we move content around or remove it entirely!

After you merge, I will test them with the URL Inspection tool on google search console and verify the fixes.

maiyahershey (approved):

PR #1668, opened 2024-10-30
Bugfixes from bugbash

rcvalerio:


inductiva/storage/storage.py
```diff
@@ -181,40 +181,47 @@ def upload(
...
-    zip_file_size = os.path.getsize(input_zip_path)
-    logging.info("Input archive size: %s",
-                 format_utils.bytes_formatter(zip_file_size))
+    if is_dir:
+        input_path = _zip_folder(local_path)
+        if remote_path:
+            remote_path = os.path.dirname(remote_path.lstrip("/"))
+            remote_path = os.path.join(remote_path, constants.TMP_ZIP_FILENAME)
+        else:
+            remote_path = constants.TMP_ZIP_FILENAME
```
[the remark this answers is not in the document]
rcvalerio: I believe you reviewed old code

inductiva/storage/storage.py
```diff
@@ -181,40 +181,47 @@ def upload(
...
+    else:
+        input_path = local_path
+        remote_path = remote_path or input_path
```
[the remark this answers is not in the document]
rcvalerio: I believe you reviewed old code

rcvalerio:
@Pbarbosa92 can you help with improving the user experience? For example, improving the logging messages and error handling (should we print the exception? should we give a general error message?)

inductiva/storage/storage.py
```diff
@@ -133,34 +131,27 @@ def _print_contents_table(contents):
 def upload_from_url(
     url: str,
     remote_dir: str,
-    file_name: str = None,
-    unzip: bool = False,
-    overwrite: bool = True,
+    remote_path: str = None,
```
Pbarbosa92, line 134: What is the difference from remote_dir and remote_path?

Pbarbosa92:
> @Pbarbosa92 can you help with improving the user experience? For example, improving the logging messages and error handling (should we print the exception? should we give a general error message?)

What kind of exceptions can we get? Any APIException is already handled

inductiva/storage/storage.py
rcvalerio: remote_dir is the the name of the workspace, remote_path is the path inside the workspace. Accepting better argument names

Pbarbosa92: So, do you mean if `remote_dir=hello` and `remote_path=world` this would create the following path `paulo/hello/world` or `paulo/workspace/hello/world`?

Pbarbosa92:
If I try the upload `inductiva.storage.upload(local_path="/Users/paulobarbosa/Downloads/swan-jana",remote_dir="abc",remote_path="hello")` I get the following error:

`ApiTypeError: RequestQueryParams missing 1 required arguments: ['file_name']`

inductiva/storage/storage.py
rcvalerio: the prefix "workspace" is always included in GCP bucket but the user doesn't need to know. If u were uploading a file test.txt it would be saved in .../workspace/hello/test.txt

Pbarbosa92: So what is the remote path for?

rcvalerio:
> If I try the upload `inductiva.storage.upload(local_path="/Users/paulobarbosa/Downloads/swan-jana",remote_dir="abc",remote_path="hello")` I get the following error:
> 
> `ApiTypeError: RequestQueryParams missing 1 required arguments: ['file_name']`

You need to run the web-api with the bugfixes, it has not been merged to dev yet: https://github.com/inductiva/inductiva-web-api/pull/1337

inductiva/storage/storage.py
rcvalerio: sorry, i forgot to add the remote_path, it will be  .../workspace/hello/world/test.txt

remote_path is basically used to specify the path inside the workspace (in this case, workspace named hello)

Pbarbosa92: I'm a bit confused why would we need both and not just use remote_dir="hello/world"

rcvalerio: that makes sense, I will implement that change

rcvalerio: Upon further investigation I think combining `remote_dir` and `remote_path` might be a bit confusing, especially since we don’t want files stored at the root of the workspace. I believe renaming `remote_dir` to `workspace_name` would make it clearer that this argument is defining a workspace folder, while `remote_path` can still specify a location within that workspace.

This way, it’s clear that files go into a workspace folder, and it feels a bit more intuitive. What do you think?

rcvalerio: Since we now allow storing files in arbitrary paths, rather than just within the workspace folder with specified filenames, I believe it might make sense to use a single `remote_path` argument. However, if we make that change, we need to decide whether to keep the restriction of only allowing files to be stored inside subfolders (e.g., providing `remote_path`="" would not be allowed).

If we lift this restriction, we might need to adjust the list command and modify the storage database table.

rcvalerio: I implemented those changes, now there is only a single remote_dir. Let me know what you think

PR #1681, opened 2024-11-06
Update reef3d and xbeach

Pbarbosa92:
This PR updates the xbeach tutorial.


tutorials/simulators/CustomImage.md
```diff
@@ -0,0 +1,378 @@
...
+```sh
+apptainer run ... mpirun -np 4 -use_hwthread_cpus fds mccaffrey.fds
+```
+
+Ensure the MPI version in your container matches the one specified in `MPIConfig`.
+For instance, if your container has OpenMPI 1.10.6, choose a compatible OpenMPI
+version (like 1.10.7) in `MPIConfig`.
+
+Currently, we support OpenMPI versions 1.10.7 and 4.1.6. If you need a different
+version, please contact us.
```
maiyahershey, line 121: can you add a contact here @Pbarbosa92

tutorials/simulators/CustomImage.md
```diff
@@ -0,0 +1,378 @@
...
+        input_dir=input_dir,
+        commands=commands,
+        on=machine_group,
+        n_vcpus=56,
+        use_hwthread=False,
+        storage_dir="3D_dam_break_with_obstacle")
+	```
+
+	As you can see, this is very similar to the Reef3D example. The only difference
+    is that here we need to specify the commands to run the simulation.
```
maiyahershey, line 337: The only difference here is that we need to specify the specific commands required to run the simulation. @Pbarbosa92

maiyahershey (approved):

PR #1700, opened 2024-11-14
Add method to terminate machine groups for benchmarks

HenriquePreto:
This PR adds a method to terminate the machine groups created for a benchmark, including unit tests.

**QA**: Running the following script will terminate the machine group created for this benchmark after waiting for the task to complete.

```python3
from inductiva.benchmarks import Benchmark
from inductiva.utils import download_from_url
from inductiva.simulators import AmrWind
from inductiva.resources import MachineGroup

input_dir = download_from_url(
    "amr-wind-input-example.zip",
    unzip=True)

Benchmark(name="test-terminate-benchmark") \
    .add_run(simulator=AmrWind(),
             input_dir=input_dir,
             sim_config_filename="abl_amd_wenoz.inp",
             on=MachineGroup("c2-standard-4"),
             n_vcpus=4) \
    .run(num_repeats=1)
    .wait() \
    .terminate()
```

Closes inductiva/tasks#494


inductiva/benchmarks/benchmark.py
```diff
@@ -243,3 +245,32 @@ def filter_distinct_columns(info):
         return filter_distinct_columns(info) \
             if columns == ColumnExportMode.DISTINCT \
             else info
+
+    def terminate(self) -> Self:
```
bgregoinductiva, line 249: I would prefer to have this method implemented in the `Project` class.

inductiva/benchmarks/benchmark.py
```diff
@@ -243,3 +245,32 @@ def filter_distinct_columns(info):
...
+        """
+        Terminates all active machine groups associated with the
+        benchmark.
+
+        Returns:
+            Self: The current instance for method chaining.
+        """
+        def _handle_suffix(executer):
+            if executer.host_type == "GCP":
```
bgregoinductiva, line 259: Please use `ProviderType.GCP` instead of the raw string.

inductiva/benchmarks/benchmark.py
HenriquePreto: It could be implemented there, but I think it should be implemented in the ```Benchmark``` class for now, as it is more directly related to machine groups, whereas the ```Project``` class is more related to tasks. Wdyt?

I also think there are other methods / features in the ```Benchmark``` class that could be moved to the base class, but this could be done at a later stage during refactoring.

bgregoinductiva (approved):

Pbarbosa92 (approved):
LGTM, you could just print a better message when exception happens.

inductiva/benchmarks/benchmark.py
```diff
@@ -243,3 +246,33 @@ def filter_distinct_columns(info):
...
+            if task.info.executer
+        }
+
+        for machine in resources.get():
+            if machine.name not in machine_names:
+                continue
+            try:
+                machine.terminate(verbose=False)
+            except ApiException as api_exception:
+                logging.warning(api_exception)
```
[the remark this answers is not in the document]
HenriquePreto: I tested the exception message. For example, if the ```terminate``` method raises an exception because it attempts to terminate an already terminated machine, it prints this message:

```
Attempting to terminate an unstarted machine group.
```

I think this is already quite nice -- the ```APIException``` class implements a custom ```__str__``` method.

PR #1712, opened 2024-11-21
Add ```path``` option to ```download_from_url``` for specifying unzip location

HenriquePreto:
This PR adds a ```path``` argument to the ```download_from_url``` function, providing users the option to specify where to unzip downloaded files.

If no ```path``` is provided, the files will be extracted into a subfolder named after the zip file. For example, if the zip file is named ```files.zip```, the extracted files will be placed in a ```files/``` directory. If the user specifies ```path="."```, the files will be extracted directly into the current directory, without being placed in a subfolder.

**QA**:

```python
import inductiva

local_path = inductiva.utils.download_from_url(
    url="https://zenodo.org/records/4091612/files/model_input.zip",
    unzip=True)

# ■ Uncompressing the downloaded archive to model_input

local_path = inductiva.utils.download_from_url(
    url="https://zenodo.org/records/4091612/files/model_input.zip",
    unzip=True,
    path="other_name")

# ■ Uncompressing the downloaded archive to other_name

local_path = inductiva.utils.download_from_url(
    url="https://zenodo.org/records/4091612/files/model_input.zip",
    unzip=True,
    path=".")

# ■ Uncompressing the downloaded archive to .
```

### DIFF
--- a/inductiva/utils/files.py
+++ b/inductiva/utils/files.py
@@ -190,8 +190,8 @@ def download_from_url(
 
     # Unzip all files as they were zipped.
     if unzip and zipfile.is_zipfile(downloaded_to):
-        resulting_path = pathlib.Path(path) if path \
-            else downloaded_to.with_suffix()
+        resulting_path = pathlib.Path(path) \
+            if path else downloaded_to.with_suffix("")
         logging.info("■ Uncompressing the downloaded archive to %s",
                      resulting_path)
         _unzip(zip_path=downloaded_to, unzip_path=resulting_path)
--- a/inductiva/utils/files.py
+++ b/inductiva/utils/files.py
@@ -117,7 +117,7 @@ def get_sorted_files(data_dir: str,
     return files
 
 
-def _unzip(zip_path: pathlib.Path, unzip_path: Optional[str] = None):
+def _unzip(zip_path: pathlib.Path, unzip_path: Optional[pathlib.Path] = None):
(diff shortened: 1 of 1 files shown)


inductiva/utils/files.py
```diff
@@ -183,10 +190,11 @@ def download_from_url(url: str, unzip: bool = False) -> str:
 
     # Unzip all files as they were zipped.
     if unzip and zipfile.is_zipfile(downloaded_to):
-        resulting_path = pathlib.Path(downloaded_to).with_suffix("")
+        resulting_path = (pathlib.Path(path)
+                          if path else downloaded_to.with_suffix(""))
```
bgregoinductiva, line 194: Should it be `if path else pathlib.Path(downloaded_to).with_suffix("")`?

HenriquePreto: Actually, no, because the variable ```downloaded_to``` is already of type ```<class 'pathlib.PosixPath'>```. I printed the types to verify, with ```path=None```:

```python
# ...
    resulting_path = downloaded_to
    print(type(resulting_path))
    # Unzip all files as they were zipped.
    if unzip and zipfile.is_zipfile(downloaded_to):
        resulting_path = (pathlib.Path(path)
                          if path else downloaded_to.with_suffix(""))
        print(type(resulting_path))
# ...
```

bgregoinductiva (approved):

PR #1721, opened 2024-11-26
Add machine group info command

rodrigotuna:
This PR adds a new command 
`inductiva resources info <name>` that presents detailed information about machines in a machine group. 

Example result:
image


hpenedones:
Thanks @rodrigotuna !!

PR #1724, opened 2024-11-28
Fix empty api key error message missing

rcvalerio:
closes https://github.com/inductiva/tasks/issues/585


hpenedones (approved):
Thanks for the timely fix!

PR #1728, opened 2024-11-28
feat add seissol simulator

ayomidetobi:
@Pbarbosa92

### DIFF
--- a/inductiva/simulators/seissol.py
+++ b/inductiva/simulators/seissol.py
@@ -1,0 +1,58 @@
+"""Class to run SeisSol simulations on a custom container image."""
+
+from typing import List, Optional
+from inductiva import types, tasks, simulators
+
+
+@simulators.simulator.mpi_enabled
+class SeisSol(simulators.Simulator):
+    """Class to run SeisSol simulations on a custom container image."""
+
+    def __init__(self, container_image: str):
+        """
+        Initialize the SeisSol simulator class.
+
+        Args:
+            container_image: The container image to use for the simulation.
+                Example: container_image="docker://inductiva/seissol:latest"
+        """
+        self.container_image = container_image
+        super().__init__()
+        self.simulator = "seissol"
+        
+    def _get_image_uri(self):
+        """Return the URI of the container image."""
+        return self.container_image
+
+    def run(self,
+            input_dir: Optional[str],
+            commands: List[str],
+            *,
+            on: types.ComputationalResources,
+            storage_dir: Optional[str] = "",
+            resubmit_on_preemption: bool = False,
+            remote_assets: Optional[List[str]] = None,
+            **kwargs) -> tasks.Task:
+        """
+        Run the SeisSol simulation.
+
+        Args:
+            input_dir: Path to the directory containing the input files.
+            commands: List of commands to run (e.g., SeisSol binaries with parameters).
+            on: The computational resource to launch the simulation on.
+            storage_dir: Parent directory for storing simulation results.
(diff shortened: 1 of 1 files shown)


inductiva/simulators/seissol.py
```diff
@@ -0,0 +1,58 @@
+"""Class to run SeisSol simulations on a custom container image."""
+
+from typing import List, Optional
+from inductiva import types, tasks, simulators
+
+
+@simulators.simulator.mpi_enabled
+class SeisSol(simulators.Simulator):
+    """Class to run SeisSol simulations on a custom container image."""
```
Pbarbosa92, line 9: If we are implementing a class for SeisSol on the python client we intend that class to use our docker images and not a custom docker image.

inductiva/simulators/seissol.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self, container_image: str):
+        """
+        Initialize the SeisSol simulator class.
+
+        Args:
+            container_image: The container image to use for the simulation.
+                Example: container_image="docker://inductiva/seissol:latest"
+        """
+        self.container_image = container_image
```
Pbarbosa92, line 19: No need to change the container_image here. Lets use the docker image we compiled in kutu.

The simulator init will get the docker image based on the class name. In this case, it will look into our kutu repo and search for seissoll docker images and pick the one with the latest version (or if you provide the version it will pick that version if it exists).

Pbarbosa92: Take a look into indutiva/simulators/cans.py

inductiva/simulators/seissol.py
```diff
@@ -0,0 +1,58 @@
...
+            resubmit_on_preemption (bool): Resubmit task for execution when
+                previous execution attempts were preempted. Only applicable when
+                using a preemptible resource, i.e., resource instantiated with
+                `spot=True`.
+            remote_assets: Additional remote files that will be copied to
+                the simulation directory.
+        """
+        return super().run(input_dir,
+                           on=on,
+                           commands=commands,
```
Pbarbosa92, line 53: This part is very simulator dependent.

If seissol is a simulator that only has one binary that you call every time we don't need to accept a list of commands, since the user will call the same binary every time.

If seissol is one of those simulators that has multiple binaries and a simulation consists in multiple commands this approach is the best option.

Pbarbosa92 (changes requested):

inductiva/simulators/seissol.py
ayomidetobi: okay noted

ayomidetobi:
seissol can generate multiple binary base on the various configuration (e.g the equation model)
I have fix the code to use the Inductiva Kutu image

ayomidetobi:
I have updated the seissol simulator class use Inductiva Kutu repository image

inductiva/simulators/seissol.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        super().__init__(version=version, use_dev=use_dev)
+        self.simulator = "seissol"
+
+    def run(self,
+            input_dir: Optional[str],
+            *,
+            on: types.ComputationalResources,
+            storage_dir: Optional[str] = "",
+            n_vcpus: Optional[int] = None,
```
Pbarbosa92, line 29: Don't forget to pass the argument `use_hwthread` here.

Pbarbosa92 (changes requested):

PR #1735, opened 2024-12-03
Update simulators to run on the Custom Docker Image

Pbarbosa92:
This PR makes the amrwind use the custom docker image.


bgregoinductiva (commented):
What is the motivation for this change? Does the docs example need to be updated? Is it necessary to update the end to end test of this simulator?

Pbarbosa92:
> What is the motivation for this change? Does the docs example need to be updated? Is it necessary to update the end to end test of this simulator?

The code works in the exact same way.

The point is to move all simulators to the "custom image". After that we can remove all executers from the task-runner, making the task runner simpler and making the process of adding new simulators easier.

bgregoinductiva:
> > What is the motivation for this change? Does the docs example need to be updated? Is it necessary to update the end to end test of this simulator?
> 
> The code works in the exact same way.
> 
> The point is to move all simulators to the "custom image". After that we can remove all executers from the task-runner, making the task runner simpler and making the process of adding new simulators easier.

Ok... so why not change all the simulators now?

Pbarbosa92:
> > > What is the motivation for this change? Does the docs example need to be updated? Is it necessary to update the end to end test of this simulator?
> > 
> > 
> > The code works in the exact same way.
> > The point is to move all simulators to the "custom image". After that we can remove all executers from the task-runner, making the task runner simpler and making the process of adding new simulators easier.
> 
> Ok... so why not change all the simulators already?

hmmm, I was going to do one PR per simulator but i can do them all in the PR.

bgregoinductiva:
> > > > What is the motivation for this change? Does the docs example need to be updated? Is it necessary to update the end to end test of this simulator?
> > > 
> > > 
> > > The code works in the exact same way.
> > > The point is to move all simulators to the "custom image". After that we can remove all executers from the task-runner, making the task runner simpler and making the process of adding new simulators easier.
> > 
> > 
> > Ok... so why not change all the simulators already?
> 
> hmmm, I was going to do one PR per simulator but i can do them all in the PR.

If it's already tested I suggest to change all in the same PR.

hpenedones:
Please update the PR title if now it will include multiple simulators in one go. Thanks!

examples/fvcom/fvcom.py
```diff
@@ -17,8 +17,8 @@
 # Run simulation with config files in the input directory
 task = fvcom.run(input_dir=input_dir,
                  working_dir="run/",
-                 create_namelist="tst",
                  n_vcpus=1,
+                 debug=7,
```
bgregoinductiva, line 21: Was this change intended?

bgregoinductiva (approved):

Pbarbosa92:
@hpenedones This is a big PR.

After the reviews and everything can we do a small bugbash to test this ?

hpenedones (approved):

PR #1740, opened 2024-12-05
Removed autocomplete part from action. Added script to run all examples

Pbarbosa92:
This PR removes some autocomplete things from the github actions.

I also added a script to run all example codes locally. All examples will run on merges to main but if you want to run them locally you can use the script, and check the logs in the logs folder.


examples/run_parallel_and_cleanup.py
```diff
@@ -0,0 +1,86 @@
...
+    log_file = os.path.join(log_dir, f"{script_name}.log")
+    print(f"Running {file_path}...")
+
+    try:
+        # Run the script and capture stdout/stderr
+        result = subprocess.run(["python3", file_path],
+                                capture_output=True,
+                                text=True,
+                                check=True)
+        # Write logs to the log file
```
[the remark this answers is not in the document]
hpenedones: https://inductiva.ai/blog/article/programming-at-inductiva#dry-dont-repeat-yourself

examples/run_parallel_and_cleanup.py
```diff
@@ -0,0 +1,86 @@
...
+        with open(log_file, "w", encoding="utf-8") as log:
+            log.write(f"--- Error running {script_name} ---\n")
+            log.write("STDOUT:\n")
+            log.write(e.stdout + "\n")
+            log.write("STDERR:\n")
+            log.write(e.stderr + "\n")
+        print(f"Error logs for {script_name} saved to {log_file}")
+
+
+# Step 1: Find all Python files, excluding this script
```
hpenedones, line 59: I know that this is really just a simple "script" to run the examples, but please try to avoid this anti-pattern:
https://inductiva.ai/blog/article/programming-at-inductiva#dont-write-scripts

And at a minimum, define small functions for each of the sections you are separating with comment headers.

Pbarbosa92: Done

examples/run_parallel_and_cleanup.py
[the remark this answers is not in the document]
hpenedones: this is the part that can be quite consolidated, because the block below is almost the same

        with open(log_file, "w", encoding="utf-8") as log:
            log.write(f"--- Output for {script_name} ---\n")
            log.write("STDOUT:\n")
            log.write(result.stdout + "\n")
            log.write("STDERR:\n")
            log.write(result.stderr + "\n")
        print(f"Logs for {script_name} saved to {log_file}")

examples/run_parallel_and_cleanup.py
hpenedones: Thanks Paulo!

examples/run_parallel_and_cleanup.py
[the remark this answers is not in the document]
Pbarbosa92: I changed things a bit. We still have the try except but now we only have one time this part that writes to the file

examples/run_parallel_and_cleanup.py
```diff
@@ -0,0 +1,114 @@
...
+import subprocess
+import shutil
+import sys
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+#checks if we are in the examples directory
+if not os.path.basename(os.getcwd()) == "examples":
+    print("Run this script from the examples directory.")
+    sys.exit(1)
+# Parse command-line arguments
```
hpenedones, line 12: consider using argparse library like we do for the CLI...

examples/run_parallel_and_cleanup.py
```diff
@@ -0,0 +1,114 @@
...
+    """Delete all directories named 'input-example' from the given path."""
+    for root, dirs, _ in os.walk(path, topdown=False):
+        for dir_name in dirs:
+            if dir_name.endswith("input-example"):
+                dir_path = os.path.join(root, dir_name)
+                print(f"Deleting directory {dir_path}...")
+                shutil.rmtree(dir_path)
+
+
+all_python_files = gather_python_files(".")
```
hpenedones, line 98: this could be inside a main() function

examples/run_parallel_and_cleanup.py
```diff
@@ -0,0 +1,114 @@
...
+if len(sys.argv) > 2:
+    print("Usage: python run_parallel_and_cleanup.py [max_threads]")
+    sys.exit(1)
+
+# Default to 2 threads if no argument is provided
+max_threads = int(sys.argv[1]) if len(sys.argv) == 2 else 4
+if max_threads <= 0:
+    print("Maximum number of threads must be a positive integer.")
+    sys.exit(1)
+
```
hpenedones, line 22: avoid having code that is not part of any function...

hpenedones (approved):
Thanks @Pbarbosa92, it is starting to look much better, I left a few more suggestions (sorry for being a bit picky) -- but we want our code to look good to show that we really care about engineering excellence, even in the "small things"! :)

examples/run_parallel_and_cleanup.py
```diff
@@ -0,0 +1,114 @@
...
+        for file in files
+        if file.endswith(".py") and file != current_script
+    ]
+    return python_files
+
+
+def run_python_files(python_files, max_working_threads):
+    """Run all Python files in parallel with a limit on active threads."""
+    with ThreadPoolExecutor(max_workers=max_working_threads) as executor:
+        futures = [executor.submit(run_script, file) for file in python_files]
```
HenriquePreto, line 83: Suggestion: you could instead use ```_= list(executor.map(run_script, python_files))```. The ```list``` ctor processes the results and ensures that any exceptions raised by the threads are raised in the main thread.

examples/run_parallel_and_cleanup.py
```diff
@@ -0,0 +1,114 @@
...
+if not os.path.basename(os.getcwd()) == "examples":
+    print("Run this script from the examples directory.")
+    sys.exit(1)
+# Parse command-line arguments
+if len(sys.argv) > 2:
+    print("Usage: python run_parallel_and_cleanup.py [max_threads]")
+    sys.exit(1)
+
+# Default to 2 threads if no argument is provided
+max_threads = int(sys.argv[1]) if len(sys.argv) == 2 else 4
```
HenriquePreto, line 18: Suggestion: Instead of using the hardcoded value of ```4``` threads, you could let the ```ThreadPoolExecutor``` determine the number of ```max_workers``` automatically by passing ```None```. By default, it uses the formula: ```min(32, (os.cpu_count() or 1) + 4)```.

examples/run_parallel_and_cleanup.py
Pbarbosa92: But what would be the difference?
Is it just readability ?

Pbarbosa92:
> Thanks @Pbarbosa92, it is starting to look much better, I left a few more suggestions (sorry for being a bit picky) -- but we want our code to look good to show that we really care about engineering excellence, even in the "small things"! :)

Done.
It's good to be picky. Standards should be high.

examples/run_parallel_and_cleanup.py
HenriquePreto: Yes, for readability reasons. And because you are reusing code already provided by the library -- that list comprehension with the ```executor.submit``` method call is used inside the ```executor.map``` method.

hpenedones (approved):

PR #1746, opened 2024-12-06
Move swan to custom docker

Pbarbosa92:
This PR makes swan use the custom docker image. By doing this I managed to fix the issue swan had with running a simulation with the outputs of another simulation.

### DIFF
--- a/inductiva/resources/machines_base.py
+++ b/inductiva/resources/machines_base.py
@@ -140,7 +140,23 @@ class BaseMachineGroup(ABC):
         max_vcpus = int(self.quota_usage["max_vcpus"])
         max_instances = int(self.quota_usage["max_instances"])
 
-        return VCPUCount(max_vcpus, max_vcpus // max_instances)
+        if self.threads_per_core == 2:
+            cores_per_machine = (max_vcpus // max_instances)
+        else:
+            cores_per_machine = (max_vcpus // max_instances) // 2
+
+        return VCPUCount(max_vcpus, cores_per_machine)
+    
+    @property
+    def available_vcpus(self):
+        """Returns the maximum number of vCPUs that can be used on a task.
+        
+        On a machine group with 2 machines, each with 4 vCPUs, this will return 4.
+        On an elastic machine group, this will also return 4.
+        On an MPI cluster this will return the total number of vcpus because
+        we can run on the total number of vcpus.
+        """
+        return self.n_vcpus.per_machine
 
     @property
     def name(self):
--- a/inductiva/resources/machines_base.py
+++ b/inductiva/resources/machines_base.py
@@ -140,12 +140,13 @@ class BaseMachineGroup(ABC):
         max_vcpus = int(self.quota_usage["max_vcpus"])
         max_instances = int(self.quota_usage["max_instances"])
 
-        if self.threads_per_core == 2:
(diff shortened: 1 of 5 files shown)


inductiva/resources/machines_base.py
```diff
@@ -140,7 +140,24 @@ def n_vcpus(self):
         max_vcpus = int(self.quota_usage["max_vcpus"])
         max_instances = int(self.quota_usage["max_instances"])
 
-        return VCPUCount(max_vcpus, max_vcpus // max_instances)
+        if self.threads_per_core == 2:
```
hpenedones, line 143: Isn't this the opposite?
Can you maybe write it down in two steps?
1. Compute num vcpus per machine
2. Compute the number of cores per machine 

Also, is the docstring consistent with this new logic?

hpenedones (commented):
Not sure about the calculations because I don't know exactly the values of the variables in each case, but seemed a bit confusing... Sorry

inductiva/resources/machines_base.py
Pbarbosa92: I agree that this is a bit confusing.

I have to do this because the max_vcpus that we get from the back end does not take into consideration the threads per core.

I've tried to simplify things a bit.

hpenedones: But then the fix needs to happen in the backend, not here... right?

Pbarbosa92: I'm not sure about that. We are getting the values from the quotas. Quotas are related to costs and the cost of a machine is not changed by threads per core. @luispcunha is this correct? This is the reason the max_vcpus is not affected by threads per core right?

hpenedones: That would be even more "worrying" -- as this is a property of a VM or physical machine, independently of things like costs or quotas.
We need clear methods that return the number of physical cores and the number of vcpus, ideally for each machine in a machine group. We can start by having that in the MachineGroup (assuming that the machines inside a MachineGroup are "uniform" -- i.e. they are all the same) -- but it ideally it should even be at the level of the machine, as potentially, with the task-runner being launched manually, a machine group can (already) be heterogeneous...

Pbarbosa92: But we should do this refactor in another PR no? I want to fix the swan feature asap so that the user can start using it.

hpenedones: Ok, then move forward, and open an issue please (if there isn't one yet) for that refactoring (with label technical debt). Thanks!

hpenedones (approved):

Pbarbosa92:
Can i have some QA on this?

```python
"""SWAN example."""
import inductiva
# Instantiate machine group
machine_group = inductiva.resources.MachineGroup("c2-standard-4")
machine_group.start()

   "swan-input-example.zip", True)

# Initialize the Simulator
swan = inductiva.simulators.SWAN()

task = swan.run(input_dir=input_dir,
               sim_config_filename="a11refr.swn",
               on=machine_group)
task.wait()
task2 = swan.run(input_dir=None,
               sim_config_filename=f"{task.id}/a11refr.swn",
               remote_assets=[f"{task.id}"],
               on=machine_group)

task2.wait()
task2.download_outputs()
machine_group.terminate()
```

This is one example

PR #1756, opened 2024-12-13
Use status on fail

rodrigotuna:
This PR uses the status sent by the file-tracker and prints the fail message to the user.

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -1136,6 +1136,7 @@ class Task:
 
     async def tail_file(self, filename: str) -> str:
         """Get the last 10 lines of a file in the task's working directory."""
+        # pylint: disable=C3001
         formatter = lambda message: self._format_list_of_lines(
             message, filename, sep="\n", endl="")
         return await self._file_operation(Operations.TAIL,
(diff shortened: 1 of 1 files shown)


inductiva/tasks/task.py
```diff
@@ -1122,17 +1123,25 @@ async def _file_operation(self, operation: Operations, **kwargs) -> str:
...
-        message = await self._file_operation(Operations.LIST)
-        return self._format_directory_listing(message)
+        return await self._file_operation(
+            Operations.LIST, formatter=self._format_directory_listing)
 
     async def tail_file(self, filename: str) -> str:
         """Get the last 10 lines of a file in the task's working directory."""
-        message = await self._file_operation(Operations.TAIL, filename=filename)
-        return self._format_list_of_lines(message, filename, sep="\n", endl="")
+        # pylint: disable=C3001
```
Pbarbosa92, line 1139: Why don't we just define a function instead of disabling the pylint errors?

Pbarbosa92 (changes requested):

Pbarbosa92 (approved):

PR #1757, opened 2024-12-16
Export output files of a task to a remote URL

luispcunha:
Adds a method to export the output files of a task to a remote storage location.

### Usage:

To export an arbitrary file from inductiva storage:
```python
operation = inductiva.storage.export("path/in/inductiva/storage", "https://www.signedurl.com")
operation.wait()
```

To export the output ZIP of a task:
```python
task = inductiva.tasks.Task(task_id)
task.export_output("https://www.signedurl.com", wait=True)
```

### Current limitations:

- Only works with individual files and https URLs. 
- If the provided URL is an AWS S3 URL, there's a 5Gb limit (on AWS side) on the upload size.

### DIFF
--- a/inductiva/storage/__init__.py
+++ b/inductiva/storage/__init__.py
@@ -1,2 +1,2 @@
 """Storage related functions."""
-from .storage import get_space_used, listdir, upload, upload_from_url, remove_workspace
+from .storage import get_space_used, listdir, upload, upload_from_url, remove_workspace, export, StorageOperation
--- a/inductiva/storage/storage.py
+++ b/inductiva/storage/storage.py
@@ -1,4 +1,5 @@
 """Methods to interact with the user storage resources."""
+import time
 import logging
 import os
 import tqdm
@@ -8,14 +9,14 @@ import urllib
 import inductiva
 from inductiva import constants
 from inductiva.api import methods
-from inductiva.client import exceptions
+from inductiva.client import exceptions, models
 from inductiva.client.apis.tags import storage_api
 from inductiva.utils import format_utils
 
 
 def _print_storage_size_and_cost() -> int:
     """ Print the storage size and cost.
-    
+
     return the storage size in bytes.
     """
     api = storage_api.StorageApi(inductiva.api.get_client())
@@ -134,12 +135,12 @@ def upload_from_url(
     Upload a file from a given URL to a specified remote directory.
 
     If no file name is provided, the function extracts the name from the URL.
-    
+
     Args:
-        url (str): The URL of the file to be uploaded. 
+        url (str): The URL of the file to be uploaded.
         remote_dir (str): The path to the remote directory where the file will
             be stored.
-        file_name (str, optional): The name to save the uploaded file as. 
+        file_name (str, optional): The name to save the uploaded file as.
             If not specified, the name will be extracted from the URL.
     """
     api_instance = storage_api.StorageApi(inductiva.api.get_client())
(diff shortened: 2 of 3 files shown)


rcvalerio (commented):
I think the name `storage.export` might be too generic—if we add more options in the future (e.g., direct S3 exports), should we rename it to `storage.export_signed_url` to keep it consistent?

`inductiva.storage.export` works for workspace files too, right? If so, there might be the problem of the `workspace` prefix in the path.

HenriquePreto:
I'm not sure if this is supposed to work, but I tried generating a signed URL to upload the input files to a bucket and export the output files there, and it didn't work:

```python
import inductiva
from inductiva.client.apis.tags.tasks_api import TasksApi

task_id = "n23gw9z3h75xcww0m9pbjmcga"

api_config = inductiva.api.methods.get_api_config()

with inductiva.ApiClient(api_config) as client:
    api_instance = TasksApi(client)
    api_response = inductiva.api.methods.get_upload_url(
        api_instance.get_input_upload_url,
        path_params={"task_id": task_id},
    )
    signed_url = api_response["url"]

task = inductiva.tasks.Task(task_id)
task.export_output(dest_url=signed_url)
```

rcvalerio:
> If we support `s3` exports, the method could be the same, but accept an `s3` url. Wdyt?

That might work, yes.

> For now it doesn't work. For the function to be generic and accept both task paths and "workspace" paths, I can't make it use the "workspace" prefix. Do you have any suggestion?

This is a recurring problem and should be fixed in the storage layer itself, maybe during the storage refactor. What do you think @HenriquePreto

HenriquePreto:
> > I'm not sure if this is supposed to work, but I tried generating a signed URL to upload the input files to a bucket and export the output files there, and it didn't work:
> > ```python
> > import inductiva
> > from inductiva.client.apis.tags.tasks_api import TasksApi
> > 
> > task_id = "n23gw9z3h75xcww0m9pbjmcga"
> > 
> > api_config = inductiva.api.methods.get_api_config()
> > 
> > with inductiva.ApiClient(api_config) as client:
> >     api_instance = TasksApi(client)
> >     api_response = inductiva.api.methods.get_upload_url(
> >         api_instance.get_input_upload_url,
> >         path_params={"task_id": task_id},
> >     )
> >     signed_url = api_response["url"]
> > 
> > task = inductiva.tasks.Task(task_id)
> > task.export_output(dest_url=signed_url)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Error message:
> > ```shell
> > Started export operation ...
> > Operation failed: Failed to export file: 400 <?xml version='1.0' encoding='UTF-8'?><Error><Code>MalformedSecurityHeader</Code><Message>Invalid argument.</Message><Details>Your request has a malformed header. Header was included in signedheaders, but not in the request.</Details><ParameterName>content-type</ParameterName></Error>
> > ```
> 
> It should be working now. (there was a missing header in the upload request)

Yes, it's working now. It prints:

```
Started export operation ...
Operation completed successfully.
```

HenriquePreto:
> > If we support `s3` exports, the method could be the same, but accept an `s3` url. Wdyt?
> 
> That might work, yes.
> 
> > For now it doesn't work. For the function to be generic and accept both task paths and "workspace" paths, I can't make it use the "workspace" prefix. Do you have any suggestion?
> 
> This is a recurring problem and should be fixed in the storage layer itself, maybe during the storage refactor. What do you think @HenriquePreto

Yes, this will be fixed during the storage refactor

HenriquePreto (approved):

rcvalerio (approved):

PR #1758, opened 2024-12-16
Updated Reuse Files tutorial with the feature to reference output of other tasks 

rcvalerio:
This PR documents the ability to reference files from other tasks outputs to the existing `Reuse Input Files Across Multiple Simulations` tutorial.

I renamed the existing tutorial as `Reuse Files Across Multiple Simulations` since now it is also possible to reuse output files.

PR #1765, opened 2024-12-17
Updated Reuse Files tutorial with the feature to reference output of other tasks

maiyahershey:
This PR documents the ability to reference files from other tasks outputs to the existing Reuse Input Files Across Multiple Simulations tutorial.

I renamed the existing tutorial as Reuse Files Across Multiple Simulations since now it is also possible to reuse output files.


maiyahershey:
This PR replaces this because of the branch merger mistake: https://github.com/inductiva/inductiva/pull/1758

Pbarbosa92 (approved):

PR #1775, opened 2024-12-19
Remove aiortc dependency

rodrigotuna:
This PR removes the aiortc dependency for python versions greater or equal to 3.13. Related to inductiva/tasks#661. File-tracking feature is still "available" for versions below 3.13

### DIFF
--- a/inductiva/tasks/file_tracker.py
+++ b/inductiva/tasks/file_tracker.py
@@ -3,14 +3,21 @@ import asyncio
 import json
 import uuid
 import enum
-import aiortc
 import logging
 
+aiortc_imported = True
+
+try:
+    import aiortc
+except ImportError:
+    aiortc_imported = False
+
 # STUN/TURN server configuration
-ICE_SERVERS = [
-    aiortc.RTCIceServer("stun:webrtc.inductiva.ai:3478"),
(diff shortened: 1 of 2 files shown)


Pbarbosa92:
Hmmm, but don't we need this for the new feature added ?

Pbarbosa92 (approved):
Looks good to me.

But I'm not sure we should do this gymnastics for an unannounced feature. Maybe we could move this code to a PR and remove it from prod. And work on a decent fix for the next release (maybe use another library or maybe they will fix this soon).

What do you think @hpenedones ?

hpenedones:
If this PR solves the problem, I am ok with it -- because we can keep it in PROD without a broad announcement, and we can ask a few selected external users to test it and give us feedback. Thanks both!

rcvalerio:
Tested with python 3.13 and installation worked.

rcvalerio (approved):

hpenedones (approved):

rodrigotuna:
> Tested with python 3.13 and installation worked.

Thank you for the feedback !!

Pbarbosa92:
Let's merge to main and do a quick release. Agree @hpenedones ?

hpenedones:
Yes, please! Thanks!

PR #1781, opened 2025-01-06
Add follow tail

rodrigotuna:
This PR adds the argument `follow` to the tail operation allowing the file to be streamed instead of just a one-shot operation. 

Example:
`inductiva tasks tail <task_id> <file> -f`


inductiva/_cli/cmd_tasks/_tail.py
```diff
@@ -10,11 +10,22 @@
...
+async def consume(task: tasks.Task, args: argparse.Namespace, fout: TextIO):
+    first = True
+    try:
+        async for lines in task._tail_file(  # pylint: disable=protected-access
+                args.filename, args.lines, args.follow):
+            prefix = "\033[s" if first else "\033[u\033[s"
+            print(f"{prefix}{lines}", file=fout, end="", flush=True)
+            first = False
+    except asyncio.CancelledError:
+        await task.close_stream()
```
Pbarbosa92, line 26:
```suggestion
async def consume(task: tasks.Task, args: argparse.Namespace, fout: TextIO):
    prefix = "\033[s"
    try:
        async for lines in task._tail_file(  # pylint: disable=protected-access
                args.filename, args.lines, args.follow):
            print(f"{prefix}{lines}", file=fout, end="", flush=True)
            prefix = "\033[u\033[s"
    except asyncio.CancelledError:
        await task.close_stream()
```

inductiva/_cli/cmd_tasks/_tail.py
```diff
@@ -35,4 +46,10 @@ def register(parser):
                            type=int,
                            default=10,
                            help="Number of lines to show.")
+    subparser.add_argument(
+        "--follow",
```
Pbarbosa92, line 50: Can i combine --follow with --lines? If yes, what is the result?

Pbarbosa92 (changes requested):
On the consume method we can change the logic a bit to avoid having to use the `first` variable. 

We we should also check for the flag follow and lines, and give an exception if the user tries to use them both.

How does this work on the back end? Can it deal with files that increase by a lot or is it limited by an X amount of lines? Lets say i do a follow. On the first request i get 10 lines. But meanwhile my file grew by 1000 lines. Will i get the full 1000 lines in the following request or only the last X lines and the rest will be dropped ?

inductiva/_cli/cmd_tasks/_tail.py
rodrigotuna: Yes. Instead of 10 lines being presented and refreshed, `--lines` will be presented and refreshed.

rodrigotuna:
> On the consume method we can change the logic a bit to avoid having to use the `first` variable.
> 
> We we should also check for the flag follow and lines, and give an exception if the user tries to use them both.
> 
> How does this work on the back end? Can it deal with files that increase by a lot or is it limited by an X amount of lines? Lets say i do a follow. On the first request i get 10 lines. But meanwhile my file grew by 1000 lines. Will i get the full 1000 lines in the following request or only the last X lines and the rest will be dropped ?

The backend periodically sends the last n lines of the files, so each message only contains the number of lines defined by the user.

Pbarbosa92:
> > On the consume method we can change the logic a bit to avoid having to use the `first` variable.
> > We we should also check for the flag follow and lines, and give an exception if the user tries to use them both.
> > How does this work on the back end? Can it deal with files that increase by a lot or is it limited by an X amount of lines? Lets say i do a follow. On the first request i get 10 lines. But meanwhile my file grew by 1000 lines. Will i get the full 1000 lines in the following request or only the last X lines and the rest will be dropped ?
> 
> The backend periodically sends the last n lines of the files, so each message only contains the number of lines defined by the user.

Hmmm. So, basically the follow is a tail that is done every X seconds?

What if I do follow to a file with lines=5 and the file was not changed for a while? Will the back end send me the same last 5 lines every time or it only sends when new lines are detected ?

rodrigotuna:
> > > On the consume method we can change the logic a bit to avoid having to use the `first` variable.
> > > We we should also check for the flag follow and lines, and give an exception if the user tries to use them both.
> > > How does this work on the back end? Can it deal with files that increase by a lot or is it limited by an X amount of lines? Lets say i do a follow. On the first request i get 10 lines. But meanwhile my file grew by 1000 lines. Will i get the full 1000 lines in the following request or only the last X lines and the rest will be dropped ?
> > 
> > 
> > The backend periodically sends the last n lines of the files, so each message only contains the number of lines defined by the user.
> 
> Hmmm. So, basically the follow is a tail that is done every X seconds?
> 
> What if I do follow to a file with lines=5 and the file was not changed for a while? Will the back end send me the same last 5 lines every time or it only sends when new lines are detected ?

Same 5 lines every time

rodrigotuna:
> > > On the consume method we can change the logic a bit to avoid having to use the `first` variable.
> > > We we should also check for the flag follow and lines, and give an exception if the user tries to use them both.
> > > How does this work on the back end? Can it deal with files that increase by a lot or is it limited by an X amount of lines? Lets say i do a follow. On the first request i get 10 lines. But meanwhile my file grew by 1000 lines. Will i get the full 1000 lines in the following request or only the last X lines and the rest will be dropped ?
> > 
> > 
> > The backend periodically sends the last n lines of the files, so each message only contains the number of lines defined by the user.
> 
> Hmmm. So, basically the follow is a tail that is done every X seconds?
> 
> What if I do follow to a file with lines=5 and the file was not changed for a while? Will the back end send me the same last 5 lines every time or it only sends when new lines are detected ?

I implemented your suggestion of only sending messages if there were file updates. I added you as a reviewer in the Task Runner.

Pbarbosa92:
> > > > On the consume method we can change the logic a bit to avoid having to use the `first` variable.
> > > > We we should also check for the flag follow and lines, and give an exception if the user tries to use them both.
> > > > How does this work on the back end? Can it deal with files that increase by a lot or is it limited by an X amount of lines? Lets say i do a follow. On the first request i get 10 lines. But meanwhile my file grew by 1000 lines. Will i get the full 1000 lines in the following request or only the last X lines and the rest will be dropped ?
> > > 
> > > 
> > > The backend periodically sends the last n lines of the files, so each message only contains the number of lines defined by the user.
> > 
> > 
> > Hmmm. So, basically the follow is a tail that is done every X seconds?
> > What if I do follow to a file with lines=5 and the file was not changed for a while? Will the back end send me the same last 5 lines every time or it only sends when new lines are detected ?
> 
> I implemented your suggestion of only sending messages if there were file updates. I added you as a reviewer in the Task Runner.

Nice. But I think the print on the client's side will be tricky. We need to avoid repeated prints with the same information. We need to test this.

PR #1793, opened 2025-01-13
Update Resource Type value on tasks list

arasahcor:
Related to https://github.com/inductiva/tasks/issues/690#issuecomment-2586860349

Resource type was showing `LOCAL n/a` for example, on tasks list.
Now it shows `vm_name` instead `vm_type`, if `vm_type === n/a`


Pbarbosa92 (approved):
LGTM

PR #1798, opened 2025-01-15
Add task-runner tutorial

rodrigotuna:
This PR adds the task-runner tutorial. Explaining how to setup, monitor and launch simulations on the local task-runner.


maiyahershey (changes requested):
this is excellent work Rodrigo! well done :)) 

I just have two comments!

1- make sure to add this new .md file to the toc tree (index.md) in the tutorials folder under Quick recipes just paste how_to/use-local-task-runner  (let me know if you need my help with that
2- just change the title of the .md file to be how-to/use-local-task-runner so the URL makes more sense and the title of the document is present tense **Use the Local Task-Runner** (without -ing) so it follows the same format as the rest of the tutorials.

maiyahershey (approved):
LGTM

tutorials/how_to/use-local-task-runner.md
```diff
@@ -0,0 +1,137 @@
+# Use the Local Task-Runner
+
+The local task-runner allows you to execute simulations on your own computer. This feature is particularly useful for testing and running simulations at minimal cost, as you only pay for storage while utilizing your local computational resources for free.
```
hpenedones, line 3: I think we need to explain better that the reason the user pays for (cloud) storage is that the inputs and outputs of the simulation will still be transferred to and from the cloud buckets and preserved there for future access.

PR #1802, opened 2025-01-20
Fix windows auth login

bgregoinductiva:
Related to https://github.com/inductiva/tasks/issues/715

The `--private` flag can be used to hide the API Key during live demos.

### DIFF
--- a/inductiva/utils/authentication.py
+++ b/inductiva/utils/authentication.py
@@ -1,6 +1,35 @@
 """This module contains utility functions for handling authentication. """
 
+import base64
+import json
+import re
+
 from inductiva import constants
+
+
+def is_valid_jwe(token):
+    try:
+        # Split the token into its parts
+        parts = token.split(".")
+        if len(parts) != 5:
+            return False
+
+        # Check if the token parts contain only valid base64url characters
+        base64url_pattern = re.compile(r'^[A-Za-z0-9_-]+$')
+        if not all(base64url_pattern.match(part) for part in parts):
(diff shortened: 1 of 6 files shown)


rcvalerio (approved):
The downside is that we are adding a new dependency

bgregoinductiva:
@hpenedones, I refactored the JWE validation to not use any external dependency.

inductiva/utils/authentication.py
```diff
@@ -1,8 +1,37 @@
...
+import re
+
 from inductiva import constants
 
 
+def is_valid_jwe(token):
+    try:
+        # Split the token into its parts
+        parts = token.split(".")
+        if len(parts) != 5:
```
hpenedones, line 14: Does this always hold?
https://stackoverflow.com/questions/58324325/why-is-a-jwt-split-into-three-dot-delimited-parts

We don't need to make a very strict validation of JWT tokens (that can be done by the backend) -- we just need to check there are no weird characters that would break the request to the backend.

hpenedones (commented):
Thanks Bruno!

inductiva/utils/authentication.py
bgregoinductiva: I think it does hold, since it is a JWE. Nevertheless, I think you're right. I refactored the code to just validate the characters.

hpenedones (approved):
Thanks for detecting the root cause of this issue, and for the timely fix, Bruno!

PR #1805, opened 2025-01-21
Add sep char to end of path when listing directory

HenriquePreto:


rcvalerio (approved):
This fixes an error when listing a path which doesn't end in "/"

PR #1811, opened 2025-01-23
Installation instructions on README and docs

hpenedones:
Highlighting how easy is to install and authenticate directly on the README of the Python package and on the docs website, because so far this is only visible inside the web console.

 Note: also removed the hardcoded mention to the version of the package on the README, that easily gets outdated. We have a badge there to show the always up to date version, so...

### DIFF
--- a/docs/preinstallation/system/system-requirements.md
+++ b/docs/preinstallation/system/system-requirements.md
@@ -8,7 +8,7 @@ We’ve gathered tips and feedback from users with different setups who’ve run
 into a few bumps while installing our API. The steps below are here to save 
 you time and help you sidestep the most common setup challenges!
 
-## Pre-requirements
+## System requirements
 
 - Step 1: Check if Python is Installed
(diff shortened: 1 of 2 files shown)


rcvalerio (approved):

docs/preinstallation/system/system-requirements.md
maiyahershey, line 1: it is probably better to create a new file for the installation that sits right before the system requirements so the user can follow them chronologically. should we do that instead @hpenedones?

hpenedones: In the docs (see file below) the requirements appear listed before the installation procedure. Here it's just the README of the python package that people see in GitHub or PyPi. I will leave a small note that user needs Python and pip -- but for the audience that goes directly to pypi this is "obvious" because pypi is a repository of python packages, that are installed via pip. For the users going directly to the GitHub project, that could potentially be less evident -- I will leave a note.

PR #1814, opened 2025-01-23
Add full path in list-files command

rodrigotuna:
This PR changes the `list-files` command accomplishing two improvements:
- print the entire path of the files in the task for reusage of the files in other commands
- appearance
image

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -550,7 +550,10 @@ class Task:
 
         return "".join(new_lst)
 
-    def _format_directory_listing(self, directories: list, indent=0) -> str:
+    def _format_directory_listing(self,
+                                  directories,
+                                  current_path="",
+                                  indent=0) -> str:
         """Formats a dictionary with directory information.
 
         This method formats a dictionary with directory information and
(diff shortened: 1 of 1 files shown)


inductiva/tasks/task.py
```diff
@@ -550,30 +550,44 @@ def _format_list_of_lines(self,
...
+            indent: The current indentation level.
         """
-        color_code = "\033[34m"
-        reset_color = "\033[0m"
-        contents = (f"{color_code}Directory contents:{reset_color}\n"
-                    if indent == 0 else "")
 
-        for item in directories:
+        def build_prefix(indent, is_last):
+            """Creates the correct prefix for the current level."""
```
hpenedones, line 569: Looks quite nice! But just to have an idea of the generality of this implementation: does the color code work in any kind of terminal (e.g. on powershell on windows?) or at least, if it does not work, does it fall back to having no colors but prints the things correctly?

hpenedones (commented):
Thanks @rodrigotuna : this way it is much easier to copy paste the path of a specific file from here, and then use it to tail the file!

inductiva/tasks/task.py
rodrigotuna: Updated it to work everywhere. To answer your question new versions of windows support it: https://en.wikipedia.org/wiki/ANSI_escape_code#DOS_and_Windows

hpenedones: Awesome! Thanks @rodrigotuna !

hpenedones (approved):

PR #1821, opened 2025-01-24
Fix bug  when showing available machines

### LINKED ISSUE #1820: Fix error when trying to list available machine types in Python Client.

catarinalegria:
closes #1820


Pbarbosa92 (changes requested):
This seems to fiz the issue. But could you implement a test to catch other issues like this in the future?

Pbarbosa92:
Something here is breaking the precious listing

Captura de ecrã 2025-01-27, às 11 58 40

Pbarbosa92:
This just merged to main. Was that intended?

PR #1822, opened 2025-01-27
Resources are now started on run if not started.

Pbarbosa92:
With this PR the user no longer needs to use mg.start before running a simulation. The run method will start a machine if not started.

### DIFF
--- a/inductiva/simulators/simulator.py
+++ b/inductiva/simulators/simulator.py
@@ -156,10 +156,8 @@ class Simulator(ABC):
                 "to learn how to create your own computational resource.")
 
         if not on.started:
-            logging.info("")
-            logging.info(
-                "■ The computational resource is not started. Starting it now.")
-            logging.info("")
+            logging.info("\n■ The computational resource is not started."
+                         " Starting it now.\n")
             on.start()
 
         self.validate_computational_resources(on)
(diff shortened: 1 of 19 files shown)


rcvalerio (approved):
I like this feature, however, should we have a `auto_start_mg` parameter which is `True` by default in the `simulation.run()`?

inductiva/simulators/simulator.py
```diff
@@ -155,6 +155,13 @@ def run(
                 "run-parallel_simulations.html "
                 "to learn how to create your own computational resource.")
 
+        if not on.started:
+            logging.info("")
+            logging.info(
+                "■ The computational resource is not started. Starting it now.")
+            logging.info("")
```
Franreno, line 162: Why do we need the empty logging.info before and after the main log message? Couldn't if be like

```python
logging.info("■ The computational resource is not started. Starting it now.")
```

or even

```python
logging.info("\n■ The computational resource is not started. Starting it now.\n")
```

Pbarbosa92: Done

Franreno (approved):

Pbarbosa92:
> I like this feature, however, should we have a `auto_start_mg` parameter which is `True` by default in the `simulation.run()`?

Maybe that would be a property of the resource. When creating a resource have a allow_auto_start.

I don't like having an argument on a method that will control the way we do things to another argument.

Issue #664, opened 2023-09-21
Tasks should automatically to the user launched machines 


begaiym-k:
I have been testing launching machines and running my scenarios in there. At the moment for the simulation to run my machines I need to pass machine_group param to the simulate method. However, in case I restart the script or reload the window I will have to retrieve the machine group object and pass it to the simulate method again. 
I suggest automating this for the user. If I for example, have machines running in the background my tasks should automatically go there even without me explicitly passing machine_group parameter.

IvanPombo:
I like this suggestion! However, there is more to this than it seems... For example, users may forget easily that they have left machines up, since there is no reference to them in their scripts anymore. Could you elaborate on this proposal a bit more - with ideas on the implementation side - so we can more easily discuss the details.

hpenedones:
Right, this is a good question, and I like that you are thinking of ways of removing user pain. But indeed it's a bit unclear what the best option is, because:
- what happens if the user has more than one machine group running? where should the task go to?
- what happens if the user really wanted to send some things to the default pool? does this mean it's now impossible to do it, until the machine group is terminated? Maybe that's a fair "rule", but it's not easy to estimate if this is "intuitive behavior" for the user...
At the moment, I think we should keep this kind of things "explicit choices". It's always easy to implement a default hidden choice, but until we know what is really intuitive, I would keep them explicit.

But maybe you have some idea in mind?

sarmento:
I think the users needs to be explicit about where to go.
What we can have is a method that allows us to select the "most recent" group, so that if the user does something like

```
machine_group=admin.most_recent_machine_group()
```

it's always the last group the user started.

If 

```
machine_group=None
```

then is should go to de default pool.

IvanPombo:
What do you guys think of having this suggestion as an environment variable, like with the `inductiva.api_key`, `inductiva.working_dir`, etc... Obviously, users would still need to fetch the machine and decide which one they have, but then, they wouldn't need to pass it to the simulate - which could be remove from all of them. In this way, we clean up a bit the simulator part and still offer the user the explicit choice. Obviously, @sarmento suggestion still makes sense, to make things easier.

Example for creating a machine group:
```python
import inductiva

# Create a machine group
machines = inductiva.resources.MachineGroup(...)

inductiva.machine_group = machines

machines.start()

scenario = inductiva.fluids.WindTunnel()
scenario.simulate()
```

Example for using an already started machine group:
```python
import inductiva

# Create a machine group
[...]

PR #1832, opened 2025-01-29
Bump minimum supported Python version to 3.9

luispcunha:
Making Python 3.9 the minimum supported version since 3.8 has reached [end-of-life](https://devguide.python.org/versions/) and some of our dependencies no longer support it (e.g., `aiortc`).

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.9]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


hpenedones (approved):
Awesome! 
What happens when a user with an older version of Python tries to install Inductiva? We need to make sure it fails and does not fallback to installing an older version of our package. Is that "guaranteed"?

hpenedones:
Ok, do you guys know how to do that? Is it simple?

hpenedones:
> Ok, do you guys know how to do that? Is it simple?

turned out to be a bit of pain, but it's done. all old versions are yanked. Closed #664

hpenedones:
This closes https://github.com/inductiva/tasks/issues/661

PR #1835, opened 2025-01-30
Update start and terminate MachineGroup's methods to use ```machine_group_id``` parameter.

catarinalegria:
depends of https://github.com/inductiva/inductiva-web-api/pull/1553


inductiva/resources/machine_cluster.py
```diff
@@ -108,20 +108,11 @@ def start(self, wait_for_quotas: bool = False):
...
               become available before starting the resource.
         """
-        return super().start(
-            wait_for_quotas=wait_for_quotas,
-            is_elastic=self.__is_elastic,
-            num_vms=self.num_machines,
-            spot=self.__spot,
-            type=self.__type,
-        )
+        return super().start(wait_for_quotas=wait_for_quotas,)
```
HenriquePreto, line 111: I don't think this method needs to be implemented since it simply calls the base class method without adding any parameters.

inductiva/resources/machine_cluster.py
```diff
@@ -108,20 +108,11 @@ def start(self, wait_for_quotas: bool = False):
...
 
     def terminate(self, verbose: bool = True):
         """Terminates the MPI Cluster."""
-        return super().terminate(num_vms=self.num_machines,
-                                 is_elastic=self.__is_elastic,
-                                 spot=self.__spot,
-                                 verbose=verbose)
+        return super().terminate()
```
HenriquePreto, line 115: > I don't think this method needs to be implemented since it simply calls the base class method without adding any parameters.

Same here

inductiva/resources/machines.py
```diff
@@ -118,19 +118,11 @@ def start(self, wait_for_quotas: bool = False):
               become available before starting the resource.
         """
 
-        return super().start(
-            wait_for_quotas=wait_for_quotas,
-            is_elastic=self.__is_elastic,
-            num_vms=self.num_machines,
-            spot=self.spot,
-        )
+        return super().start(wait_for_quotas=wait_for_quotas,)
```
HenriquePreto, line 121: > I don't think this method needs to be implemented since it simply calls the base class method without adding any parameters.

Same here

HenriquePreto: This happens in other places as well.

inductiva/resources/machines_base.py
```diff
@@ -362,15 +362,12 @@ def can_start_resource(self) -> bool:
 
         return is_cost_ok and is_vcpu_ok and is_instance_ok
 
-    def start(self, wait_for_quotas: bool = False, **kwargs):
```
hpenedones, line 365: @HenriquePreto : were you still relying on something like this on the benchmarks to introduce some sort of "labels"?

HenriquePreto: That happens here: https://github.com/inductiva/inductiva/blob/f81a1df92dfe075e7f7018e265e9c25c7c92ea1f/inductiva/benchmarks/benchmark.py#L154

.github/workflows/python-package.yml
```diff
@@ -51,7 +51,7 @@ jobs:
     - name: Link Checker
       uses: lycheeverse/lychee-action@v1.9.0
       with:
-        fail: true
```
hpenedones, line 54: @Pbarbosa92 : if you don't mind, this will be more silent here for now (but you can re-enable it, perhaps on a separate action, that is not related with the python code, but it's just about documentation)?

hpenedones (approved):
Love it!

PR #1847, opened 2025-01-31
Improve error messages

rodrigotuna:
This PR adds error messages from the client for file operations.

### DIFF
--- a/inductiva/_cli/cmd_tasks/_list_files.py
+++ b/inductiva/_cli/cmd_tasks/_list_files.py
@@ -5,24 +5,15 @@ import sys
 import asyncio
 
 from inductiva import _cli, tasks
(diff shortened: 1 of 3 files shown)


hpenedones (commented):
Awesome, thanks for this! (small note: in case there is a way to avoid the duplication of code, and reuse it in both places, the better).
But it's really great that you went into the detail of polishing the user experience and giving helpful feedback! Thanks

rodrigotuna:
> Awesome, thanks for this! (small note: in case there is a way to avoid the duplication of code, and reuse it in both places, the better). But it's really great that you went into the detail of polishing the user experience and giving helpful feedback! Thanks

Mb got lazy :sweat_smile:

hpenedones (approved):
Thanks!

PR #1853, opened 2025-02-05
Refactor MachineGroup Classes

catarinalegria:
closes [#848](https://github.com/inductiva/tasks/issues/848)

After doing some research regarding the best way to reduce repetitive code and only have "class-specific" arguments in the `MachineGroup`, `ElasticMachineGroup` and `MPICluster` classes, I came across this [StackOverflow comment](https://stackoverflow.com/a/74076953), which suggested using [Dataclasses](https://docs.python.org/3/library/dataclasses.html) to simplify the classes code and inheritance logic.

After some reading, I do believe the @dataclass approach is much simpler:

- Automatically generates `__init__` and handles attributes inheritance.
- Allows to use `__post_init` method to do custom initialization logic.

The initialization order is:
2. The `BaseMachineGroup` class `__init__` is called automatically
3. The child specific arguments are initialized (for example `num_machines` for `MachineGroup` class)
4. The child class `__post_init__` is called
5. The parent class `__post_init__` is called (manually called inside child's `__post_init`)

In the perspective of the user and usability of the classes, everything stays the same! So, although this may seem quite a big change, everything will "stay the same" and we gain by introducing simplicity and reducing a lot of code. WDYT?


inductiva/resources/__init__.py
```diff
@@ -1,7 +1,6 @@
 #pylint: disable=missing-module-docstring
-from .machines import MachineGroup, ElasticMachineGroup
+from .machines import MachineGroup, ElasticMachineGroup, MPICluster
```
hpenedones, line 2: consider consolidating everything in just two files:
1 ) machine_groups.py would have the BaseMachineGroup, MachineGroup, ElasticMachineGroup, MPICluster classes (and maybe the get function)
2) utis.py (or some better name) -- to expose these auxiliary functions like list_available_machines

hpenedones (commented):
I think it's great, you deleted quite a bit of code!! Awesome

inductiva/resources/__init__.py
catarinalegria: I agree! I think that we should probably still split the machine_groups into two files (otherwise the resulting file would be extremely long): `machine_groups_base.py` and `machine_groups.py`. WDYT?

catarinalegria: Also, will do this change in a following PR (started doing it in this one but realized it was too much changes for a single PR).

inductiva/resources/machines.py
```diff
@@ -1,96 +1,69 @@
...
+            therefore the user must set an upper limit to the disk size, to
+            prevent uncontrolled costs. Once that limit is reached, the disk
+            is no longer automatically resized, and if the task continues to
+            output files, it will fail.
+        num_machines: The number of virtual machines to launch.
+        spot: Whether to use spot machines.
+    """
+    # __init__ arguments
+    provider: Union[machine_types.ProviderType, str] = "GCP"
+    auto_resize_disk_max_gb: Optional[int] = None
```
catarinalegria, line 46: I am a little confused by this argument (`auto_resize_disk_max_gb`). I am copying the previous implementation's behaviour, although I am not sure if it is correct:

- `MachineGroup` and `ElasticMachineGroup`: is a class argument set by the user and by default is `None`. However, this argument is not mentioned in our [documentation](https://docs.inductiva.ai/en/latest/api_reference/computational_resources/machinegroup_class.html).
- `MPICluster`: is not an argument in the class initialization and it's value by default is 500.

Is this behaviour correct?

catarinalegria: A similar thing happens with the `provider` argument which can be set when creating a `MachineGroup` -- but does not appear in the documentation. 

In addition, `ElasticMachineGroup` and `MPICluster` can only have _GCP_ as the `provider` arg value -- is this behaviour also correct?

inductiva/resources/__init__.py
hpenedones: I think it's ok if it's a long file -- as long as the things we have in that file are closely related (and they are) -- it actually helps you understand what each of those subclasses will do, because you can also see the code of the base class in the same file. I.e. we don't need to separate or aggregate things in order to minimize file sizes, we should be guided by the concepts of "separation of concerns" -- put in the same class / file / module / component things that are really very closely related and interdependent -- separate things that are unrelated and can live independently of others.

inductiva/resources/machines.py
```diff
@@ -1,96 +1,69 @@
 """Classes to manage different Google Cloud machine group types."""
+from dataclasses import dataclass
 import logging
 from typing import Optional, Union
-import datetime
 
 from inductiva.resources import machine_types, machines_base
 
 
+@dataclass
```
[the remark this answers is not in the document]
catarinalegria: @hpenedones the features we are using are available from python3.7, so all good ! (tested running an example using Python 3.9.21)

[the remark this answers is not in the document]
hpenedones: Great!!

hpenedones:
@bgregoinductiva : if we merge this to development soon this will be extensively tested by the end-to-end tests, right? This should be a pure refactoring, with 0% change to the behaviour.

bgregoinductiva:
> @bgregoinductiva : if we merge this to development soon this will be extensively tested by the end-to-end tests, right? This should be a pure refactoring, with 0% change to the behaviour.

Yes, it will. And we can run the tests using this branch before merging. Nevertheless, I would like @Pbarbosa92 to take a look at this PR as well.

Pbarbosa92:
This is a very hard PR to review. There are lots of changes.

I would not recommend releasing this today. The end-to-end tests will launch simulations. But the classes changed affect everything (cli, prints on the logs).

Shouldn't we merge to dev (after the release) and use it until the next release? Just to make sure nothing breaks.

hpenedones:
Ok, sounds good. Let's not merge yet and review more carefully. Thanks!

catarinalegria:
> I have one question. I'm failing to see the clear benefit of changing things to a DataClass. Did we do it just to avoid the **init** ?
> 
> Can I have a clear example on the simplification this brings?
Yes, i think the biggest plus is the automatic initialization process.

Also, before we were repeating attributes in the parent and child classes - because of the explicit `__init__` in each class (which was much more error prone), and it was really difficult to understand the new attributes / changes in each child class. This way when we add / change an attribute we only need to change it one time instead of four times.

inductiva/resources/machines.py
```diff
@@ -1,112 +1,80 @@
...
-        self.__is_elastic = False
-
-        if register:
-            logging.info("■ Registering MachineGroup configurations:")
-            self._register_machine_group(num_vms=self.num_machines,
-                                         spot=self.spot,
-                                         is_elastic=self.__is_elastic)
+    @property
+    def short_name(self) -> str:
+        return "MachineGroup"
```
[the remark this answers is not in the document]
catarinalegria: It is being used in the logging of `_register_machine_group`

catarinalegria:
> > This way when we add / change an attribute we only need to change it in one place instead of four different places (or more if we eventually add more classes in the future).
> 
> What about the docstring? In that case we still to add all the arguments everywhere?

I tried to find a way to inherit the docstring but could not find a way

hpenedones:
> I'm noticing one difference with the previous behavior. Some arguments are set to null by the user and the defaults are specified in the backend. In this case, while previously we printed the information including the values coming from the backend, we now are showing `N/A`. Previously:

Is this different behaviour that @luispcunha detected now solved ? @catarinalegria

catarinalegria:
> > I'm noticing one difference with the previous behavior. Some arguments are set to null by the user and the defaults are specified in the backend. In this case, while previously we printed the information including the values coming from the backend, we now are showing `N/A`. Previously:
> 
> Is this different behaviour that @luispcunha detected now solved ? @catarinalegria

yes! @hpenedones

inductiva/resources/machines_base.py
```diff
@@ -29,112 +29,89 @@ class ResourceType(enum.Enum):
...
+    data_disk_gb: int = 10
+    max_idle_time: Optional[Union[datetime.timedelta, int]] = None
+    auto_terminate_ts: Optional[datetime.datetime] = None
+
+    # Internal attributes
+    _free_space_threshold_gb = 5
+    _size_increment_gb = 10
+    _id = None
+    _name = None
+    create_time = None
```
hpenedones, line 62: maybe group these ones that do not start by underscores? those are being accessed in the code somewhere else, right? (now they are marked here as "internal attributes"...)

catarinalegria:
> > ```python
> > mg = inductiva.resources.MachineGroup("e2-standard-4")
> > ■ Registering MachineGroup configurations:
> >         · Name:                       api-cn54k5xrpqz8xdno8j8pigp77
> >         · Machine Type:               e2-standard-4
> >         · Data disk size:             10 GB
> >         · Total memory (RAM):         16 GB
> >         · Maximum idle time:          N/A
> >         · Auto terminate timestamp:   N/A
> >         · Number of machines:         1
> >         · Spot:                       True
> >         · Estimated cloud cost of machine group: 0.060 $/h
> >         · You are spending 2.6x less by using spot machines.
> > ```
> 
> I think we should also log the provider next to the machine type. It could be either a different row, or to make it clear that e2-standard-4 is a GCP machine type, it could be in parenthesis after the machine type. Wdyt?

yes! added a new line (after the Name row) with this info

inductiva/resources/machines.py
```diff
@@ -244,3 +185,82 @@ def _log_machine_group_info(self):
...
+        auto_terminate_ts: Moment in which the resource will be
+            automatically terminated.
+        num_machines: The number of virtual machines to launch.
+    """
+    # Constructor arguments
+    num_machines: int = 2
+
+    # Internal attributes
+    auto_resize_disk_max_gb = None
+    _type = machines_base.ResourceType.MPI.value
```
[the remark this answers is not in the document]
catarinalegria: I believe we need to pass the string value:
`File "/Users/catarinaalegria/inductiva/inductiva/client/schemas.py", line 2076, in cast_to_allowed_types
    raise ValueError('Invalid type passed in got input={} type={}'.format(
ValueError: Invalid type passed in got input=ResourceType.MPI type=<enum 'ResourceType'>`

inductiva/resources/machines.py
catarinalegria: from the discussion it does not seem to be really straightforward solution and will probably require some additional discussion, I believe we should open a different issue to to take care of this issue. wdyt @bgregoinductiva @hpenedones @luispcunha ?

If not and the solution is simple, i can handle it rn

bgregoinductiva:
> @bgregoinductiva : if we merge this to development soon this will be extensively tested by the end-to-end tests, right? This should be a pure refactoring, with 0% change to the behaviour.

I used this branch to run the e2e tests and got this error:

```
>       mg = inductiva.resources.MachineGroup(
            machine_type="e2-small",
            num_machines=1,
            register=not custom_vm_image,
        )
E       TypeError: MachineGroup.__init__() got an unexpected keyword argument 'register'
```

Just to confirm, was the removal of the `register` attribute intended? @catarinalegria, @hpenedones

catarinalegria:
> > @bgregoinductiva : if we merge this to development soon this will be extensively tested by the end-to-end tests, right? This should be a pure refactoring, with 0% change to the behaviour.
> 
> I used this branch to run the e2e tests and got this error:
> 
> ```
> >       mg = inductiva.resources.MachineGroup(
>             machine_type="e2-small",
>             num_machines=1,
>             register=not custom_vm_image,
>         )
> E       TypeError: MachineGroup.__init__() got an unexpected keyword argument 'register'
> ```
> 
> Just to confirm, was the removal of the `register` attribute intentional? @catarinalegria, @hpenedones

Yes! @bgregoinductiva it was intentionally removed (the behavior is the same tho but without that attribute)

inductiva/resources/machines.py
hpenedones: I'm not sure why this is hard (would need to know more of the details of the code) -- but I'm fine in general with making that change in a separate pull request.

bgregoinductiva:
> > > @bgregoinductiva : if we merge this to development soon this will be extensively tested by the end-to-end tests, right? This should be a pure refactoring, with 0% change to the behaviour.
> > 
> > 
> > I used this branch to run the e2e tests and got this error:
> > ```
> > >       mg = inductiva.resources.MachineGroup(
> >             machine_type="e2-small",
> >             num_machines=1,
> >             register=not custom_vm_image,
> >         )
> > E       TypeError: MachineGroup.__init__() got an unexpected keyword argument 'register'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Just to confirm, was the removal of the `register` attribute intentional? @catarinalegria, @hpenedones
> 
> Yes! @bgregoinductiva it was intentionally removed (the behavior is the same tho but without that attribute)

After removing the attribute from the tests code, e2e tests passed: https://github.com/inductiva/inductiva-web-api/actions/runs/13503232363

inductiva/resources/machines.py
bgregoinductiva: New issue: https://github.com/inductiva/tasks/issues/852

PR #1854, opened 2025-02-05
Fix task-runner launch windows compatibility issue

rodrigotuna:
This PR fixes https://github.com/inductiva/tasks/issues/749.


hpenedones:
Thanks for the fix, I think we can then merge it to main as a hot fix... Right @Pbarbosa92 ?

hpenedones (approved):

PR #1863, opened 2025-02-07
Tail more than one file

rodrigotuna:
This PR adds the possibility to tail multiple files at the same time.
Eg.
```
inductiva tasks tail atc7vurp18btp89xkpxvt0fx6 stdout.txt stderr.txt -f
```


hpenedones:
Let's do a soft deprecation of the old logs functionality, and on the CLI start invoking this new mechanism. But we let the old system in production for a while for back-compatibility reasons (with old python client versions that people might still have)

bgregoinductiva (approved):

hpenedones (approved):
Awesome!

PR #1873, opened 2025-02-11
Display GPU information when listing available machines.

catarinalegria:


catarinalegria:
**Preview**:
image

catarinalegria:
> **Preview**: image

@Pbarbosa92 WDYT?

catarinalegria:
Also, I did not find another command / log where I think we also need to display this info. @Pbarbosa92 do you know if we should add the gpu information in other places?

Pbarbosa92:
I'm note sure if we are already printing the GPU info when we register a machine group. If we are not we should.

Pbarbosa92:
> **Preview**: image

This is confusing

For example, if a user wants a machine with 3 GPUs, what machine type should they choose?  
A common guess might be:  
**`g2-standard-16-8-nvidia-l4`**  

But we know this is incorrect—yet the user doesn’t.  

Previously, CPU-only machine names followed a clear and logical pattern. However, the new GPU machine names seem inconsistent and difficult to decode. For instance:  
**`a3-ultragpu-8g`** doesn’t even mention the number of CPUs in the name.  

To improve clarity, we could adjust the column names in the table to:  
**Family | Machine Type | Suffix**  

- For CPU machines, the suffix could represent the number of vCPUs.  
- For GPU machines, the suffix could indicate GPU specs (e.g., `8g`).  

This way, we can establish a clear pattern:  
**`Family-MachineType-Suffix`**  

Additionally, we could include extra columns with details about each machine’s specifications to further help users understand their options.  

Something like this:
Family: C2
| Machine Type | Suffix |   | vcpus | gpus |
|--------------|--------|---|-------|------|
| standard     | 2      |   | 2     | 0    |
|              | 4      |   | 4     | 0    |
|              | 5      |   | 8     | 0    |
| hicpu     | 2      |   | 2     | 0    |
|              | 4      |   | 4     | 0    |
|              | 5      |   | 8     | 0    |

Family: a3
| Machine Type | Suffix |   | vcpus | gpus |
|--------------|--------|---|-------|------|
| megagpu     | 8g      |   | 208     | 8    |
| highgpu     | 1g      |   | 26     | 1    |
|              | 2g      |   | 52     | 2    |
|              | 4g      |   | 104     | 4    |
|              | 8g      |   | 208     | 8    |

This will be painful to translate to code.

What do you think @hpenedones ?

catarinalegria:
> > **Preview**: image
> 
> This is confusing
> 
> For example, if a user wants a machine with 3 GPUs, what machine type should they choose? A common guess might be: **`g2-standard-16-8-nvidia-l4`**
> 
> But we know this is incorrect—yet the user doesn’t.
> 
> Previously, CPU-only machine names followed a clear and logical pattern. However, the new GPU machine names seem inconsistent and difficult to decode. For instance: **`a3-ultragpu-8g`** doesn’t even mention the number of CPUs in the name.
> 
> To improve clarity, we could adjust the column names in the table to: **Family | Machine Type | Suffix**
> 
> * For CPU machines, the suffix could represent the number of vCPUs.
> * For GPU machines, the suffix could indicate GPU specs (e.g., `8g`).
> 
> This way, we can establish a clear pattern: **`Family-MachineType-Suffix`**
> 
> Additionally, we could include extra columns with details about each machine’s specifications to further help users understand their options.
> 
> Something like this: Family: C2
> 
> Machine Type	Suffix		vcpus	gpus
> standard	2		2	0
> 4		4	0
> 5		8	0
> hicpu	2		2	0
> 4		4	0
> 5		8	0
> Family: a3
> 
> Machine Type	Suffix		vcpus	gpus
> megagpu	8g		208	8
> highgpu	1g		26	1
> 2g		52	2
> 4g		104	4
> 8g		208	8
> This will be painful to translate to code.
> 
> What do you think @hpenedones ?

@Pbarbosa92 I agree! WDYT?
I also added the Config column since for some machine types it is also included in the name. Do you want me to included it in the Suffix column, for ex "8-lssd"? Or the extra column is more clear?

image

Pbarbosa92:
I would put one  suffix per line. Otherwise it gets hard to understand the specs for a specific config.

For instance:
Suffix | vcpus | gpus 
2         | 2.        | 0
4         | 4.        | 0

catarinalegria:
> I would put one suffix per line. Otherwise it gets hard to understand the specs for a specific config.
> 
> For instance: Suffix | vcpus | gpus 2 | 2. | 0 4 | 4. | 0

@Pbarbosa92 done!
image

hpenedones:
@catarinalegria : can you try to reduce code duplication a bit (i.e. DRY principle), please? Often what is different in the two branches of the if/else is a small detail, but whole function calls and arguments are the same, so we can maybe re-use those.

Pbarbosa92:
> what is the CONFIG column that seems to be always N/A? @catarinalegria @Pbarbosa92

I believe it is related with lssd machines

hpenedones:
> @catarinalegria : can you try to reduce code duplication a bit (i.e. DRY principle), please? Often what is different in the two branches of the if/else is a small detail, but whole function calls and arguments are the same, so we can maybe re-use those.

@catarinalegria : regarding this -- perhaps we can do this as a follow up PR (less urgent) and merge this one, so that we have the functionality as part of this release.

catarinalegria:
> > @catarinalegria : can you try to reduce code duplication a bit (i.e. DRY principle), please? Often what is different in the two branches of the if/else is a small detail, but whole function calls and arguments are the same, so we can maybe re-use those.
> 
> @catarinalegria : regarding this -- perhaps we can do this as a follow up PR (less urgent) and merge this one, so that we have the functionality as part of this release.

Already did a little refactor!

hpenedones (approved):

PR #1910, opened 2025-02-24
improve docstrings for auto_terminate_minutes

Pbarbosa92:


inductiva/resources/machines_base.py
```diff
@@ -73,10 +73,15 @@ def __init__(
...
-                this method.
+            auto_terminate_minutes: Duration, in minutes, the Machine will be
+                kept alive. After auto_terminate_minutes minutes the Machine
+                will be terminated. This time will start counting after calling
+                this method (regardless if you started the resource or not).
+                This differs from `max_idle_time`, which terminates the Machine
+                only after it has been idle for a specified duration.  
+                Use `auto_terminate_minutes` to enforce a strict time limit,
+                while `max_idle_time` is better suited for terminating  
+                the Machine when it is no longer actively processing tasks.
```
hpenedones, line 84: Machine --> MachineGroup

hpenedones (approved):
Thanks Paulo!

inductiva/resources/machines_base.py
Pbarbosa92: But this is not for the MachineGroup class. This is the base machine abstract class.

hpenedones: Yes, but it is a BaseMachineGroup (not a BaseMachine)... (Even though the file name where it is defined is misleading -- which is one thing that @catarinalegria had in mind consolidating ... the file names)

PR #1911, opened 2025-02-24
Improve project download

Pbarbosa92:
This PR makes the project.download_outputs download all files into a folder with the project name.


inductiva/projects/project.py
```diff
@@ -390,12 +391,19 @@ def wait(self):
...
+        `inductiva_output/<project_name>/<task_id's>`.
         """
         list_of_tasks = self.list()
 
+        # Set the new output dir based on the old one + project name
+        old_out_dir = inductiva.get_output_dir()
+        inductiva.set_output_dir(os.path.join(old_out_dir, self.name))
+
         for task in list_of_tasks:
             task.download_outputs()
```
hpenedones, line 404: isn't it possible to just pass the path to the download_outputs method, rather than changing the global output dir variable twice? @HenriquePreto

HenriquePreto: I think you could use the ```output_dir``` parameter in the ```download_outputs``` method, something like:

```output_dir=f"{self.name}/{task.id}"```

Right?

HenriquePreto: Btw, maybe projects should use a ```storage_dir``` by default

HenriquePreto: Then it would be just: ```inductiva.storage.download(remote_path=project.name)```
This would include (input.zip and output.zip files)

Pbarbosa92: > I think you could use the `output_dir` parameter in the `download_outputs` method, something like:
> 
> `output_dir=f"{self.name}/{task.id}"`
> 
> Right?

You could but that would create inductiva_oututs/project_name and i don't want it to create inductiva_outputs

hpenedones: But that is what your code is doing, no? from your comment:

        All the files will be stored inside
        `inductiva_output/<project_name>/<task_id's>`.

Pbarbosa92: The previous code was placing everything in project_name/task_id.

I changed it to do it like you just said.

HenriquePreto: Don't you want to separate the output files per task?

bgregoinductiva (approved):

inductiva/projects/project.py
Pbarbosa92: > Don't you want to separate the output files per task? For example: inductiva_outputs/my-project/task-1/output.zip inductiva_outputs/my-project/task-2/output.zip ...

It is already doing that.

HenriquePreto: There's this comment: `inductiva_output/<project_name>/<task_id's>`

HenriquePreto: > > Don't you want to separate the output files per task? For example: inductiva_outputs/my-project/task-1/output.zip inductiva_outputs/my-project/task-2/output.zip ...
> 
> It is already doing that.

Are you sure?

https://github.com/inductiva/inductiva/blob/4c7ad23e97b1158c513ef58fdd20095cd48f63eb/inductiva/tasks/task.py#L982-L984

HenriquePreto (approved):

PR #1918, opened 2025-02-25
Consolidate Machine Groups code in two files: `utils.py` and `machine_groups.py`.

catarinalegria:
closes [848](https://github.com/inductiva/tasks/issues/848)


tutorials/how_to/use-local-task-runner.md
```diff
@@ -99,7 +99,7 @@ Code Example:
 import inductiva
 
 # Specify the machine group
-machine_group = inductiva.resources.machine_groups.get_by_name(
```
hpenedones, line 102: This is a breaking change at the level of the code that the user needs to write -- we need to think whether we *really* want to do it (for example, it could break existing scripts that users wrote). In some situations that is ok -- if it is really a super clear advantage of doing so. But here, it is not yet very clear to me. I think we can still consolidate our python code into two files, but leave the imports that the user needs to make the same, no?

Pbarbosa92:
I'm not a fan of this. I would rather have one class per file.

Pbarbosa92 (approved):
I just ran our test simulations with this branch and everything worked fine.

Apart from changing the path to some method this PR should not cause any problems.

LGTM!

PR #1919, opened 2025-02-25
Delete inductiva/simulators/fenicsx.py

hpenedones:
Removing the proxy class for fenicsx because the support for this simulator is very partial and different from the other simulators (in fact we are not even announcing it on the website). A good implementation would have to use a custom docker image, rather than custom code on the task-runner.

See also:
https://github.com/inductiva/task-runner/pull/122


Joao97ribeiro (approved):
LGTM!

Pbarbosa92 (approved):

PR #1921, opened 2025-02-26
Add default value for `max_idle_time`.

catarinalegria:
closes [838](https://github.com/inductiva/tasks/issues/838)

### DIFF
--- a/inductiva/resources/machine_groups.py
+++ b/inductiva/resources/machine_groups.py
@@ -190,12 +190,11 @@ class BaseMachineGroup(ABC):
         pass
 
     @staticmethod
-    def _timedelta_to_seconds(
-            value: Union[datetime.timedelta, int]) -> Optional[float]:
+    def _timedelta_to_seconds(value: Union[datetime.timedelta, int]) -> float:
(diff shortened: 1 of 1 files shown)


inductiva/resources/machine_groups.py
```diff
@@ -54,7 +54,8 @@ class BaseMachineGroup(ABC):
     provider: Union[ProviderType, str] = "GCP"
     threads_per_core: int = 2
     data_disk_gb: int = 10
-    max_idle_time: Optional[Union[datetime.timedelta, int]] = None
+    max_idle_time: Union[datetime.timedelta,
+                         int] = datetime.timedelta(minutes=3)
```
Pbarbosa92, line 58: You can use `max_idle_time` as an int. It should make this code smaller.
```suggestion
                         int] = 3
```

Pbarbosa92 (approved):
Just a small change. We can use `max_idle_time` as an int because it will be converted into a timedelta later.

hpenedones (approved):
Thanks Catarina!

PR #1928, opened 2025-02-28
Add `zone`, `cpu_info` and `gpu_info` to Machine Groups

catarinalegria:
depends of https://github.com/inductiva/inductiva-web-api/pull/1651
depends of https://github.com/inductiva/inductiva-web-api/pull/1665


inductiva/resources/machine_groups.py
```diff
@@ -283,6 +276,7 @@ def _register_machine_group(self, **kwargs):
             body.get("auto_terminate_ts"))
         self._total_ram_gb = body.get("total_ram_gb")
         self._cost_per_hour = body.get("cost_per_hour")
+        self._cpu_info = body.get("cpu_info")
```
catarinalegria, line 279: with this approach we not only have access to the logical cpus but also the physical cpus. @Pbarbosa92  this is something beneficial when running the simulations right?

Pbarbosa92: Can you give me an example of the cpu_info?

What happens if we request a machine with 2 threads per core?

What happens if we request a machine with 1 threads per core?

catarinalegria: that logic is now on the backend, this is a snippet of the code that computes the `cpu_info`: 
`
...
        return {
            "cpu_count_logical": vm_info.num_cpus,
            "cpu_count_physical": vm_info.num_cpus // config.threads_per_core,
        }`

Pbarbosa92: That does not seem correct to me. the physical should be the fixed value no? and logical would be the available vcpus based on the threads per core.

If it is not to much to ask, could you add the architecture of the cpu? (if it takes little work only, we might need this info in the future)

catarinalegria: yes, but the way the backend code is atm we fetch the vcpu number based on the machine_type and then "reverse engineer" that number to get the physical cpus.

i can double check if we can fetch the cpu number instead and also the cpu architecture but not sure

Pbarbosa92:
How will this work for local task runners? I just checked for the current state of things, and we return 0 for max_vcpus and max_instances.

Will this be fixed for the local task runner?

@hpenedones Should i create a new issue for this? Or should we tackle this issue on this PR ?

Pbarbosa92 (changes requested):
The value for `cpu_cores_logical` does not seem to be correct.

If i request a `c2-standard-4` with 2 threads per core the `cpu_cores_logical` is 4 (correct). But if i ask for the same machine with 1 thread per core the `cpu_cores_logical` comes as 4 again and that is not correct.

Pbarbosa92:
> > The value for `cpu_cores_logical` does not seem to be correct.
> > If i request a `c2-standard-4` with 2 threads per core the `cpu_cores_logical` is 4 (correct). But if i ask for the same machine with 1 thread per core the `cpu_cores_logical` comes as 4 again and that is not correct.
> 
> @Pbarbosa92 this is how we are calculating the `cpu_info` in the backend:
> 
> ```
> return schemas.CPUInfo(
>             cpu_cores_logical=vm_info.num_vcpus,
>             cpu_cores_physical=vm_info.num_vcpus // request.threads_per_core,
>         ) 
> ```
> 
> that saying, shouldn't the `threads_per_core` only influence the number of `cpu_cores_physical`?

I'm getting the same number for physical and logical. But regardless, how can a variable change the number of physical cores on a cpu? The number of logical cpus is the one that can be changed by the threads per core or am i missing something ?

catarinalegria:
> > > The value for `cpu_cores_logical` does not seem to be correct.
> > > If i request a `c2-standard-4` with 2 threads per core the `cpu_cores_logical` is 4 (correct). But if i ask for the same machine with 1 thread per core the `cpu_cores_logical` comes as 4 again and that is not correct.
> > 
> > 
> > @Pbarbosa92 this is how we are calculating the `cpu_info` in the backend:
> > ```
> > return schemas.CPUInfo(
> >             cpu_cores_logical=vm_info.num_vcpus,
> >             cpu_cores_physical=vm_info.num_vcpus // request.threads_per_core,
> >         ) 
> > ```
> >     
> >   
> > that saying, shouldn't the `threads_per_core` only influence the number of `cpu_cores_physical`?
> 
> I'm getting the same number for physical and logical. But regardless, how can a variable change the number of physical cores on a cpu? The number of logical cpus is the one that can be changed by the threads per core or am i missing something ?

i understand. but given the variables that we have (`num_vcpus` (logical cpus) from the gcp API and `threads_per_core` the from user input), i am not understanding other way of computing these values besided this one. do you know other way? @Pbarbosa92

Pbarbosa92:
> > > > The value for `cpu_cores_logical` does not seem to be correct.
> > > > If i request a `c2-standard-4` with 2 threads per core the `cpu_cores_logical` is 4 (correct). But if i ask for the same machine with 1 thread per core the `cpu_cores_logical` comes as 4 again and that is not correct.
> > > 
> > > 
> > > @Pbarbosa92 this is how we are calculating the `cpu_info` in the backend:
> > > ```
> > > return schemas.CPUInfo(
> > >             cpu_cores_logical=vm_info.num_vcpus,
> > >             cpu_cores_physical=vm_info.num_vcpus // request.threads_per_core,
> > >         ) 
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > that saying, shouldn't the `threads_per_core` only influence the number of `cpu_cores_physical`?
> > 
> > 
> > I'm getting the same number for physical and logical. But regardless, how can a variable change the number of physical cores on a cpu? The number of logical cpus is the one that can be changed by the threads per core or am i missing something ?
> 
> i understand. but given the variables that we have (`num_vcpus` (logical cpus) from the gcp API and `threads_per_core` the from user input), i am not understanding other way of computing these values besided this one. do you know other way? @Pbarbosa92

`vm_info.num_vcpus // request.threads_per_core` seems fine to me, but that is logical cores. We can change the logical cores of a machine witht he threads per core.

inductiva/resources/machine_groups.py
```diff
@@ -51,6 +52,7 @@ class BaseMachineGroup(ABC):
     """
     # Constructor arguments
     machine_type: str
+    zone: Optional[str] = "europe-west1-b"
```
catarinalegria, line 55: should `zone ` have a default value so the user knows which zone it is using if no value is passed?

bgregoinductiva: I believe it should.

Pbarbosa92 (approved):

hpenedones (approved):

PR #1940, opened 2025-03-06
Improve selection of docker image

Pbarbosa92:
Issue https://github.com/inductiva/tasks/issues/883

This pr improves on the system used to select the docker image to run on the vm.

The user now only needs to pick the version. We will add the suffix `_gpu` if the resource for the task has a gpu or not.

we will also print error messages if the user uses a resource with a gpu for a simulator with no gpu support and vice versa.

Examples:
```python
mg_gpu = inductiva.resources.MachineGroup(machine_type="g2-standard-4")

#simulator with NO gpu support
cans=inductiva.simulators.CaNS()

#should fail
cans.run(input_dir="/Users/paulobarbosa/Downloads/cp2k-master/benchmarks/Fayalite-FIST",
          sim_config_filename="hello.inp",on=mg_gpu)
```
This would give
```
ValueError: The selected resource `g2-standard-4` has GPU(s), but the simulator CaNS v2.4.0 does not have a GPU version available.
Please select a different resource or simulator.
```


hpenedones:
thanks @Pbarbosa92 : do you mind doing this "Improve selection of docker image" and "terminate resource on wait" in two separate pull requests?

Pbarbosa92:
> thanks @Pbarbosa92 : do you mind doing this "Improve selection of docker image" and "terminate resource on wait" in two separate pull requests?

Will do.

rcvalerio:
This method of selecting the docker image based on the machine type can in the future be extended to support arm machines and images

inductiva/resources/machine_groups.py
```diff
@@ -129,6 +129,11 @@ def _validate_inputs(self):
             self._auto_terminate_ts = datetime.datetime.now(
                 tz=datetime.timezone.utc) + time_delta_minutes
 
+    def has_gpu(self) -> bool:
+        """Check if the machine group has a GPU."""
+        # TODO: update this code to use gpu info
+        return "g2" in self.machine_type
```
hpenedones, line 135: @catarinalegria : FYI.

rcvalerio (approved):

inductiva/tests/simulators/test_simulator_with_resources.py
```diff
@@ -44,6 +44,60 @@ def new_machine_init(self, machine_type):
...
 
+def test_simulator__wrong_version__raises_error():
+    with pytest.raises(ValueError) as excinfo:
+        inductiva.simulators.CaNS(version="999")
+    assert "not available" in str(excinfo.value)
+
+
+@mark.parametrize("use_dev,use_gpu", [(True, True), (True, False),
+                                      (False, True), (False, False)])
+def test_simulator___get_version_suffixes(use_dev, use_gpu):
```
hpenedones, line 55: I think this test is perhaps too low level and trying to replicate the logic of the methods it is testing -- and that is probably why the code needed to access a private method (and the reason then one had to disable the pylint warning). Wouldn't it be possible to a couple of simple tests in which you just test expected inputs and outputs -- with fully constructed strings (by hand) of what you expect to see in the outputs? and avoiding accessing private methods...

Pbarbosa92: I created a new method that is no longer private. Apart from that I'm not sure how we can test this in a better way since this is only used in the .run method.

PR #1941, opened 2025-03-06
Terminates resource on wait method

Pbarbosa92:
This pr allows to terminate a resource on the wait() method.

### DIFF
--- a/inductiva/tasks/run_simulation.py
+++ b/inductiva/tasks/run_simulation.py
@@ -57,6 +57,9 @@ def run_simulation(
             f"Expected result to be a string with task_id, got {type(task_id)}")
     task = tasks.Task(task_id)
 
+    # Save the resource information on the task
+    task.resource = computational_resources
+
     position = task.get_position_in_queue()
     if position is not None:
         pos_info = f"Number of tasks ahead in the queue: {position}."
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -342,6 +342,7 @@ class Task:
         self._summary = None
         # Internal state to track if the method was called from the wait method
         self._called_from_wait = False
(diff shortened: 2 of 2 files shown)


bgregoinductiva (approved):
LGTM

bgregoinductiva:
It would be even nicer if `terminate_on_exit` was a parameter of `tasks/submit` to have the BE handling the MG termination. And I believe it should be `True` by default.

Pbarbosa92:
> It would be even nicer if `terminate_on_exit` was a parameter of `tasks/submit` to have the BE handling the MG termination. And I believe it should be `True` by default.

Yes. This is just a temporary fix for that. @hpenedones do we already have a issue for this ?

hpenedones:
There is this one, that is closely related:
https://github.com/inductiva/tasks/issues/882

hpenedones:
@Pbarbosa92 : what is the advantage of having this flag on the wait method? this way the script needs to be running anyways, no?

Pbarbosa92:
> @Pbarbosa92 : what is the advantage of having this flag on the wait method? this way the script needs to be running anyways, no?

This is just to remove the machine_group.terminate() command. I can rever this

hpenedones:
> > @Pbarbosa92 : what is the advantage of having this flag on the wait method? this way the script needs to be running anyways, no?
> 
> This is just to remove the machine_group.terminate() command. I can rever this.
> 
> Edit: Done

Right, but to remove that 1 line, you need to pass 1 more argument in the wait method -- and in terms of what happens computationally nothing changes (i.e. the user would still need to leave the script running blocked on wait for this to happen, and the script also does not get really shorter or less error prone -- because the person still needs to set that argument). I think we can do this better. Thanks Paulo!

PR #1946, opened 2025-03-11
Implement ```download``` in ```storage``` API

HenriquePreto:
- Implemented the ```inductiva.storage.download``` function for downloading files or folders from remote storage.
- Developed the corresponding CLI command to interact with the ```download``` function, allowing users to specify a remote path (```remote_path```), local directory (```local_dir```), and the option to uncompress content (```uncompress```).

Closes https://github.com/inductiva/tasks/issues/189

Example Usage (CLI):

```bash
inductiva storage download <remote_path> [<local_dir>] [--uncompress]
```

### DIFF
--- a/inductiva/_cli/cmd_storage/download.py
+++ b/inductiva/_cli/cmd_storage/download.py
@@ -21,13 +21,18 @@ def register(parser):
         "The `inductiva storage download` command allows you to download files "
         "or folders from your remote storage.\n"
         "Specify the remote path, the local destination, and whether or not to "
-        "uncompress the content after downloading.\n"
-    )
-    subparser.add_argument("path", type=str,
+        "uncompress the content after downloading.\n")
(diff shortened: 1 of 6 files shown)


Pbarbosa92:
We also have `inductiva tasks download`. Should we deprecate this command ?

HenriquePreto:
> We also have `inductiva tasks download`. Should we deprecate this command ?

Eventually, yes. Or at least ```inductiva tasks download``` should invoke this command internally (since it's a specific case).

rcvalerio (approved):

HenriquePreto:
> Can we please consolidate this with `inductiva tasks download` just to avoid letting one cli method fall behind.

Yes, ofc. I can update this now.

Pbarbosa92 (approved):

PR #1947, opened 2025-03-11
Update docs: remove tiers and update quotas

bgregoinductiva:
References:
- https://github.com/inductiva/tasks/issues/846
- https://github.com/inductiva/tasks/issues/880


docs/api_reference/credits_and_quotas.md
```diff
@@ -0,0 +1,82 @@
+# Credits and Quotas
+
+All users receive 5 USD in credits upon registration. As a user you may also
```
Pbarbosa92, line 3:
```suggestion
All users receive 5 US$ in credits upon registration. As a user you may also
```

Pbarbosa92 (changes requested):
Shouldn't we talk about our estimates in this page ?

docs/api_reference/credits_and_quotas.md
```diff
@@ -0,0 +1,82 @@
+# Credits and Quotas
+
+All users receive 5 US$ in credits upon registration. As a user you may also
+purchase credits via Stripe whenever you want.
+
+Credits are consumed based on the user's resource usage. This includes factors
+like computation time and the specific machine groups utilized. For example,
```
sarmento, line 7: I think we should say just "machines". "machine groups" is an internal term, that then, rather unfortunately, is mapped to the API. But people will understand "machines".

sarmento (approved):

bgregoinductiva:
> Shouldn't we talk about our estimates in this page ?

Maybe not. We mention real time tracking of credits usage, but I'm not sure if we should explain this internal detail.

docs/api_reference/credits_and_quotas.md
bgregoinductiva: Done

PR #1963, opened 2025-03-13
Add default values to cpu/gpu info

Pbarbosa92:
This PR adds default values to the cpu info and gpu info.

By default the machine wont have a gpu.

This won't solve every issue. If the simulator has only gpu images it will fail to find an image with this defaults.

### DIFF
--- a/inductiva/resources/machine_groups.py
+++ b/inductiva/resources/machine_groups.py
@@ -293,9 +293,9 @@ class BaseMachineGroup(ABC):
         self.zone = body.get("zone")
 
         # If the API response does not contain the CPU info, set default values
-        if self._cpu_info == {} or self._cpu_info is None:
+        if not self._cpu_info:
             self._cpu_info = {"cpu_cores_logical": 2, "cpu_cores_physical": 1}
-        if self._gpu_info == {} or self._gpu_info is None:
+        if not self._gpu_info:
             self._gpu_info = {"gpu_count": 0, "gpu_name": "None"}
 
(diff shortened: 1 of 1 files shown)


catarinalegria (approved):

inductiva/resources/machine_groups.py
```diff
@@ -292,6 +292,12 @@ def _register_machine_group(self, **kwargs):
         self._gpu_info = body.get("gpu_info")
         self.zone = body.get("zone")
 
+        # If the API response does not contain the CPU info, set default values
+        if self._cpu_info == {} or self._cpu_info is None:
```
bgregoinductiva, line 296: Suggestion: could be replaced by `if not self._cpu_info:`

inductiva/resources/machine_groups.py
```diff
@@ -292,6 +292,12 @@ def _register_machine_group(self, **kwargs):
...
+            self._cpu_info = {"cpu_cores_logical": 2, "cpu_cores_physical": 1}
+        if self._gpu_info == {} or self._gpu_info is None:
```
bgregoinductiva, line 298: Suggestion: could be replaced by `if not self._gpu_info:`

hpenedones (approved):
Does it mean it will not solve it for GX users on local GPUs? Or will it?

Pbarbosa92:
> Does it mean it will not solve it for GX users on local GPUs? Or will it?

This won't solve it for GX. That part of the issue needs to be solved in the task runner.

PR #1969, opened 2025-03-18
docstring fixes

hpenedones:
Using Google style docstring and closing escaping char.


bgregoinductiva (approved):

hpenedones (commented):
small note: I noticed these from automated warnings from sphinx when generating documentation from our python code -- there are more of those that we need to fix in the future...

PR #1972, opened 2025-03-19
Allow users to pick the acceleration method for the docker image

Pbarbosa92:
This PR allows users to pass an acceleration_method to the simulator. This will take priority over the acceleration method picked based on the machine provided.

It is important to have this for local task runners, that can have good cpu's and gpu's and we need a way to let the user pick on what he wants to run.

### DIFF
--- a/inductiva/simulators/amr_wind.py
+++ b/inductiva/simulators/amr_wind.py
@@ -10,11 +10,7 @@ class AmrWind(simulators.Simulator):
     """Class to invoke a generic AmrWind simulation on the API.
     """
 
-    def __init__(self,
-                 /,
-                 version: Optional[str] = None,
-                 use_dev: bool = False,
-                 device: str = None):
+    def __init__(self, /, version: Optional[str] = None, use_dev: bool = False):
         """Initialize the AmrWind simulator.
 
         Args:
@@ -23,14 +19,8 @@ class AmrWind(simulators.Simulator):
             use_dev (bool): Request use of the development version of
                 the simulator. By default (False), the production version
                 is used.
-            device (str): Pick whether to use the CPU or GPU
-                version of the Docker image. By default, the appropriate option
-                is selected based on the hardware used to run the simulation. If
-                you explicitly request a specific device, ensure
-                that the corresponding Docker image exists with
-                `inductiva simulators ls`.
         """
-        super().__init__(version=version, use_dev=use_dev, device=device)
+        super().__init__(version=version, use_dev=use_dev)
         self.simulator = "arbitrary_commands"
         self.simulator_name_alias = "amrwind"
 
--- a/inductiva/simulators/amr_wind.py
+++ b/inductiva/simulators/amr_wind.py
@@ -10,7 +10,11 @@ class AmrWind(simulators.Simulator):
     """Class to invoke a generic AmrWind simulation on the API.
     """
 
-    def __init__(self, /, version: Optional[str] = None, use_dev: bool = False):
+    def __init__(self,
+                 /,
+                 version: Optional[str] = None,
+                 use_dev: bool = False,
+                 acceleration_method: str = None):
         """Initialize the AmrWind simulator.
 
         Args:
@@ -19,8 +23,16 @@ class AmrWind(simulators.Simulator):
             use_dev (bool): Request use of the development version of
                 the simulator. By default (False), the production version
                 is used.
+            acceleration_method (str): Pick whether to use the CPU or GPU
+                version of the Docker image. By default, the appropriate option
+                is selected based on the hardware used to run the simulation. If
+                you explicitly request a specific acceleration_method, ensure
+                that the corresponding Docker image exists with
+                `inductiva simulators ls`.
         """
-        super().__init__(version=version, use_dev=use_dev)
+        super().__init__(version=version,
+                         use_dev=use_dev,
+                         acceleration_method=acceleration_method)
         self.simulator = "arbitrary_commands"
         self.simulator_name_alias = "amrwind"
 
--- a/inductiva/simulators/amr_wind.py
+++ b/inductiva/simulators/amr_wind.py
@@ -14,7 +14,7 @@ class AmrWind(simulators.Simulator):
                  /,
                  version: Optional[str] = None,
                  use_dev: bool = False,
-                 acceleration_method: str = None):
+                 device: str = None):
         """Initialize the AmrWind simulator.
 
         Args:
@@ -23,16 +23,14 @@ class AmrWind(simulators.Simulator):
             use_dev (bool): Request use of the development version of
                 the simulator. By default (False), the production version
                 is used.
-            acceleration_method (str): Pick whether to use the CPU or GPU
+            device (str): Pick whether to use the CPU or GPU
                 version of the Docker image. By default, the appropriate option
                 is selected based on the hardware used to run the simulation. If
-                you explicitly request a specific acceleration_method, ensure
+                you explicitly request a specific device, ensure
                 that the corresponding Docker image exists with
                 `inductiva simulators ls`.
         """
-        super().__init__(version=version,
-                         use_dev=use_dev,
-                         acceleration_method=acceleration_method)
+        super().__init__(version=version, use_dev=use_dev, device=device)
         self.simulator = "arbitrary_commands"
         self.simulator_name_alias = "amrwind"
 
--- a/inductiva/simulators/dualsphysics.py
+++ b/inductiva/simulators/dualsphysics.py
@@ -12,7 +12,7 @@ class DualSPHysics(simulators.Simulator):
                  /,
                  version: Optional[str] = None,
                  use_dev: bool = False,
-                 device: str = None):
+                 device: str = "auto"):
         """Initialize the DualSPHysics simulator.
 
         Args:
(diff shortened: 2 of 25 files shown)


inductiva/simulators/amr_wind.py
```diff
@@ -10,7 +10,11 @@ class AmrWind(simulators.Simulator):
     """Class to invoke a generic AmrWind simulation on the API.
     """
 
-    def __init__(self, /, version: Optional[str] = None, use_dev: bool = False):
+    def __init__(self,
+                 /,
+                 version: Optional[str] = None,
+                 use_dev: bool = False,
+                 acceleration_method: str = None):
```
hpenedones, line 17: please use the term "device" which is more standard, instead of "acceleration_method",

Pbarbosa92: Done

inductiva/simulators/amr_wind.py
```diff
@@ -19,8 +23,14 @@ def __init__(self, /, version: Optional[str] = None, use_dev: bool = False):
             use_dev (bool): Request use of the development version of
                 the simulator. By default (False), the production version
                 is used.
+            device (str): Pick whether to use the CPU or GPU
+                version of the Docker image. By default, the appropriate option
```
hpenedones, line 27: I think it is unclear what will happen by default ("By default, the appropriate option is selected based on the hardware used to run the simulation. ")... What happens if the machine has both CPU and GPU, and the simulator also supports both? E.g. I think GROMACS or DualsPhysics are in this category...

HenriquePreto: Also, should we allow passing a ```device``` parameter even though some simulators do not support multiple devices?

Pbarbosa92: Would this be better?

```
device (str): Specifies whether to use the CPU or GPU version of
                the Docker image. By default, the GPU version is selected if
                the assigned machine has a GPU. However, if the simulator does
                not support a GPU version and a GPU-equipped machine is
                provided, an error will occur. To explicitly choose a specific
                version, use the `device` argument. Before specifying a device,
                verify that the corresponding Docker image is available by
                running `inductiva simulators ls`.
```

Pbarbosa92: > Also, should we allow passing a `device` parameter even though some simulators do not support multiple devices?

Yes.

This is the current behavior. If the machine has a GPU we try to pick a GPU docker image. If the simulator has no GPU version we will give an error.

This makes sense on the cloud. You shouldn't use a GPU machine to run a simulation only on the CPU.

The issue comes for the local task runner, that can have a decent GPU and CPU. In this case we need to allow the user to pick a device.

Let's say you wanted to run openfast on a local task runner (that has a GPU). You would never be able because we would try to get the docker image for the GPU and there is none. In this case you would have to force the 
`device="CPU"` to force the CPU docker image.

inductiva/simulators/amr_wind.py
```diff
@@ -10,7 +10,11 @@ class AmrWind(simulators.Simulator):
     """Class to invoke a generic AmrWind simulation on the API.
     """
 
-    def __init__(self, /, version: Optional[str] = None, use_dev: bool = False):
+    def __init__(self,
+                 /,
+                 version: Optional[str] = None,
+                 use_dev: bool = False,
+                 device: str = None):
```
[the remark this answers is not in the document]
Pbarbosa92: I had `acceleration_method` Hugo suggested device. I'm open for better names.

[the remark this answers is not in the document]
HenriquePreto: > I think the "device" argument could be a bit confusing. What does it mean to choose CPU? Doesn't everything need a CPU, even if it also runs on GPU?

That's true. But, for example, PyTorch allows you to pass a ```device``` parameter following (more or less) this same concept:

```python
import torch

# Check if CUDA (GPU) is available, otherwise use CPU
device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

print(device)  # Output: 'cuda' or 'cpu'

# Moving a Tensor to a Device
tensor = torch.tensor([1.0, 2.0, 3.0])
tensor = tensor.to(device)
```

I think the idea is to specify where the computationally intensive operations will be executed, like using multiple CPUs, or GPUs, or other accelerators.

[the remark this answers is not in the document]
HenriquePreto: But I'm also not sure why a user would need to specify this. Are there cases where a simulator might not provide enough information for us to determine the best option? For example, GROMACs runs faster on GPUs than on CPUs. Is there any simulator where it wouldn't be obvious for us to make this decision based on the type of machine group being used?

inductiva/simulators/amr_wind.py
hpenedones: > Would this be better?
> 
> ```
> device (str): Specifies whether to use the CPU or GPU version of
>                 the Docker image. By default, the GPU version is selected if
>                 the assigned machine has a GPU. However, if the simulator does
>                 not support a GPU version and a GPU-equipped machine is
>                 provided, an error will occur. To explicitly choose a specific
>                 version, use the `device` argument. Before specifying a device,
>                 verify that the corresponding Docker image is available by
>                 running `inductiva simulators ls`.
> ```

It is more clear to me now what you had in mind -- but I don't think we need the user to worry about this -- we can only expose the "device" argument that support both CPU and GPU, and let the user chose for those explicitly. The user should not need to "verify that the corresponding Docker image is available by running `inductiva simulators ls`."

inductiva/simulators/amr_wind.py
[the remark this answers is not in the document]
hpenedones: We don't want to guess that for the user, you can be interested in running a benchmark on the same machine to compare CPU vs GPU, for example. We can make an educated guess if we decide to set a "default", but not hide the option if both would work...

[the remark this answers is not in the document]
Pbarbosa92: > But I'm also not sure why a user would need to specify this. Are there cases where a simulator might not provide enough information for us to determine the best option? For example, GROMACs runs faster on GPUs than on CPUs. Is there any simulator where it wouldn't be obvious for us to make this decision based on the type of machine group being used?

You don't need to specify this. If you don't we will pick the fastest option (GPU) when available.

But, if you picked a CPU only simulator on a machine with a GPU we will give an error. Because it does not make sense for an user to pay a lot of money for a machine with one or more GPUs to run a simulator that will use only the CPU.

inductiva/simulators/amr_wind.py
Pbarbosa92: > > Would this be better?
> > ```
> > device (str): Specifies whether to use the CPU or GPU version of
> >                 the Docker image. By default, the GPU version is selected if
> >                 the assigned machine has a GPU. However, if the simulator does
> >                 not support a GPU version and a GPU-equipped machine is
> >                 provided, an error will occur. To explicitly choose a specific
> >                 version, use the `device` argument. Before specifying a device,
> >                 verify that the corresponding Docker image is available by
> >                 running `inductiva simulators ls`.
> > ```
> 
> It is more clear to me now what you had in mind -- but I don't think we need the user to worry about this -- we can only expose the "device" argument on the simulators classes that support both CPU and GPU, and let the user chose for those explicitly. The user should not need to "verify that the corresponding Docker image is available by running `inductiva simulators ls`."

So, my question is. What should happen if an user picks a g2 machine (has GPU) to run a CPU only simulator? What should happen?

I need to have the `device`in CPU only simulators due to the local task runner.

inductiva/simulators/amr_wind.py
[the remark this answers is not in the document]
Pbarbosa92: > We don't want to guess that for the user, you can be interested in running a benchmark on the same machine to compare CPU vs GPU, for example. We can make an educated guess if we decide to set a "default", but not hide the option if both would work...

I'm ok with going either way. @sarmento can you comment on this?

inductiva/simulators/simulator.py
```diff
@@ -29,7 +29,13 @@ class Simulator(ABC):
...
     _logger = logging.getLogger(__name__)
 
-    def __init__(self, /, version: Optional[str] = None, use_dev: bool = False):
+    _supported_device_values = ["gpu", "cpu", "auto"]
+
+    def __init__(self,
+                 /,
+                 version: Optional[str] = None,
+                 use_dev: bool = False,
+                 device: str = "auto"):
```
bgregoinductiva, line 38: Cloud the type annotation be the following?

```
device: Literal["auto", "cpu", "gpu"] = "auto"
```

Otherwise the docstring should explicitly include all the options.

Pbarbosa92: Done

bgregoinductiva (approved):

inductiva/simulators/dualsphysics.py
```diff
@@ -17,8 +21,13 @@ def __init__(self, /, version: Optional[str] = None, use_dev: bool = False):
             use_dev (bool): Request use of the development version of
                 the simulator. By default (False), the production version
                 is used.
+            device (str): Specifies whether to use the CPU or GPU version of
+                the Docker image. You can also pick `auto` and the device will
```
hpenedones, line 25: Simplified text: "Chooses between CPU or GPU for running the simulation. Default is "auto", which will auto-detect if the machine has a GPU and use if it available, otherwise use the CPU."

hpenedones (approved):
just left a suggestion to simplify the docstring text. thanks!

inductiva/simulators/dualsphysics.py
Pbarbosa92: Done. I'll merge it now.

PR #1978, opened 2025-03-21
Fix ```--decompress``` flag on CLI

HenriquePreto:
Users couldn't choose not to decompress files when downloading.

Example without automatic decompression:

```bash
inductiva storage download kjazsmk7jxao6e0a71e2neh1d/input.zip
```

Example with automatic decompression:

```bash
inductiva storage download kjazsmk7jxao6e0a71e2neh1d/input.zip --decompress
```

### DIFF
--- a/inductiva/_cli/cmd_storage/download.py
+++ b/inductiva/_cli/cmd_storage/download.py
@@ -40,7 +40,6 @@ def register(parser):
     subparser.add_argument(
         "--decompress",
         action="store_true",
-        default=True,
         help=("Decompress the downloaded file or folder if it is compressed. "
               "This option is enabled by default."),
     )


hpenedones (approved):

inductiva/_cli/cmd_storage/download.py
```diff
@@ -40,7 +40,6 @@ def register(parser):
     subparser.add_argument(
         "--decompress",
         action="store_true",
-        default=True,
         help=("Decompress the downloaded file or folder if it is compressed. "
               "This option is enabled by default."),
```
hpenedones, line 44: @HenriquePreto : we need to remove this text, "This option is enabled by default.",  right? 
@Pbarbosa92

hpenedones: or change "enabled" to "disabled"

HenriquePreto: > @HenriquePreto : we need to remove this text, "This option is enabled by default.", right? @Pbarbosa92

Right.

PR #1983, opened 2025-03-25
Kill task if `upload_file` is interrupted by user

catarinalegria:
references [#874](https://github.com/inductiva/tasks/issues/874)

**Before**:

https://github.com/user-attachments/assets/58536816-e2dd-42ba-a1a4-e5d6e80ecc24

**Now:**

https://github.com/user-attachments/assets/d75582ca-bd0e-4f3f-a199-0ced6faa88ae

### DIFF
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -138,29 +138,36 @@ def upload_input(api_instance: TasksApi, task_id, original_params,
         original_params: Params of the request passed by the user.
         type_annotations: Annotations of the params' types.
     """
+    try:
+        # Use the blocking task context
+        with blocking_task_context(api_instance, task_id):
+            input_zip_path, zip_file_size = prepare_input(task_id, original_params,
+                                                        type_annotations)
 
-    input_zip_path, zip_file_size = prepare_input(task_id, original_params,
-                                                  type_annotations)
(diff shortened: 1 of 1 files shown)


bgregoinductiva (approved):

inductiva/api/methods.py
```diff
@@ -138,29 +138,34 @@ def upload_input(api_instance: TasksApi, task_id, original_params,
...
-        notify_upload_complete(
-            api_instance.notify_input_uploaded,
-            path_params={"task_id": task_id},
-        )
-
-    logging.info("Local input directory successfully uploaded.")
-    logging.info("")
-    os.remove(input_zip_path)
+    # Use the blocking task context
+    with blocking_task_context(api_instance, task_id):
```
Franreno, line 142: Instead of calling the blocking_task_context inside of the function, i think it would look better if you call the function with the context wrapped around it. 

Like

```python
with blocking_task_content(...):
    upload_input(....)
```

instead of 

```python
def upload_input(...):
    with blocking_task_context(...):
```

Doing this you don't need to change the function definition

rcvalerio (approved):

Franreno (approved):

hpenedones:
Awesome @catarinalegria ! And love the screencasts! :)

PR #2002, opened 2025-04-01
Run task logs during task wait

Pbarbosa92:
This PR required a small refactor.

We had some logic being done on the CLI module. I have moved this logic to the task class.

Now the cli has almost no logic and only calls methods from the task, this will allow us to simply call the logs method during the task.wait().


inductiva/tasks/task.py
```diff
@@ -743,6 +753,64 @@ def wait(self,
...
+            return (False, f"Task {self.id} has terminated.\n"
+                    "Access its output using:\n\n"
+                    f"  inductiva tasks download --id {self.id}")
+        if not info.status == "computation-started":
+            return (False, f"Task {self.id} has not started yet.\n"
+                    "Wait for computation to start.")
+
+        return (True, None)
+
+    def print_tail_files(self, tail_files: List[str], lines: int, follow: bool,
```
hpenedones, line 791: Consider changing to:

```python
tail_files(self, files: List[str], num_lines: int, follow: bool, fout: TextIO):
```

inductiva/tasks/task.py
```diff
@@ -743,6 +753,64 @@ def wait(self,
...
+                         fout: TextIO):
+        """
+        Prints the result of tailing a list of files.
+
+        Args:
+            tail_files: A list of files to tail.
+            lines: The number of lines to print.
+            follow: Whether to follow the file or not.
```
hpenedones, line 799: can you explain better in the docstring what "following" a file is?

inductiva/tasks/task.py
```diff
@@ -669,13 +671,15 @@ def _handle_terminal_status(self, download_std_on_completion: bool,
...
              polling_period: int = 1,
+             silent_mode: bool = False,
              download_std_on_completion: bool = True) -> models.TaskStatusCode:
         """Wait for the task to complete.
 
         This method issues requests to the API.
 
         Args:
             polling_period: How often to poll the API for the task status.
+            silent_mode: If True, do not print the task logs to the console.
```
hpenedones, line 682: logs (stdout and stderr)

inductiva/tasks/task.py
```diff
@@ -1153,15 +1221,24 @@ async def close_stream(self):
...
-
+    async def _list_files(self) -> str:
         directory = [
             files async for files in self._file_operation(
                 Operations.LIST, formatter=self._format_directory_listing)
         ]
         return directory[0]
 
+    def list_files(self):
+        """List the files in the task's working directory.
```
hpenedones, line 1232: maybe we can expand the docstring of this method, that now will be public for task.list_files() so that the automatic python documentation looks neat. 

You can see the current DEV version (which is still a bit ugly) at:
https://docs.inductiva.ai/en/development/python/inductiva.tasks.html

inductiva/tasks/task.py
```diff
@@ -714,6 +718,12 @@ def wait(self,
                     requires_newline = False
                     sys.stdout.write("\n")
                 self._handle_status_change(status, description)
+
+                if (status == models.TaskStatusCode.COMPUTATIONSTARTED) and (
+                        not silent_mode):
+                    self.print_tail_files(["stdout.txt", "stderr.txt"], 50,
+                                          True, sys.stdout)
```
bgregoinductiva, line 725: Should this be inside the condition `if status != prev_status:`? Should the logs only be printed when the task changes status?

inductiva/tasks/task.py
```diff
@@ -1239,6 +1239,36 @@ def list_files(self):
             return 1
         return asyncio.run(self._list_files())
 
+    async def list_top(self):
```
hpenedones, line 1242: (not sure if now or later) but probably we should move this to another place, because "top" is an operation over a machine (part of a machine group) -- and not a "task" operation.

We would do something like `task.get_host_machine()` and then do `machine.top_command()` ...

hpenedones:
Super exciting feature, thanks @Pbarbosa92 !!

inductiva/tasks/task.py
Pbarbosa92: I only want to run the logs if the task status changes. There is no need trying to run logs every iteration if the task status is the same (this will be a blocking call).

inductiva/tasks/task.py
hpenedones: ok, but then maybe for now these methods should be "protected" (as in, let's add an underscore) and we use it for internal tests -- what do you think? Otherwise they become part of our public API, and further changes are breaking changes for user's code. Maybe call it for now `_run_top_on_machine()` ?

inductiva/tasks/task.py
Pbarbosa92: The `prev_status` starts at None, so, if the first status is `COMPUTATIONSTARTED` that woud enter the if `if status != prev_status:` and it would execute the logs.

The `tail_files` returns when the simulation ends or the logs end.

inductiva/tasks/task.py
Pbarbosa92: Done

inductiva/_cli/cmd_tasks/tail.py
```diff
@@ -1,48 +1,18 @@
...
-        asyncio.create_task(consume(generator, fout))
-        for generator in generators
-    ]
-    try:
-        await asyncio.gather(*tail_tasks)
-    except asyncio.CancelledError:
-        for tail_task in tail_tasks:
-            tail_task.cancel()
-        await task.close_stream()
+    task.print_tail_files(args.filename, args.lines, args.follow, fout)
```
hpenedones, line 13: isn't it just tail_files now?

Pbarbosa92: Yes.

I might need to do some refactor to this code i feel like it has repeated code.

But i wanted to at least have a working version in dev for the bug bash and after that i'll focus on refactoring.

inductiva/tasks/task.py
```diff
@@ -1209,6 +1363,48 @@ def formatter(message):
...
+        try:
+            async for lines in generator:
+                print(lines, file=fout, end="", flush=True)
+        except asyncio.CancelledError:
+            pass
+
+    def top(self, fout: TextIO):
+        """Prints the result of the `top -b -H -n 1` command.
+    
+        This command will list the processes and threads (-H) in batch mode
```
hpenedones, line 1398: missing docstring for argument fout

hpenedones (approved):

PR #2006, opened 2025-04-03
Small refactor to task class

Pbarbosa92:
Cleaning up the code a bit.

### DIFF
--- a/inductiva/_cli/cmd_tasks/top.py
+++ b/inductiva/_cli/cmd_tasks/top.py
@@ -16,8 +16,7 @@ def top(args: argparse.Namespace, fout: TextIO = sys.stdout):
     task_id = args.id
     task = tasks.Task(task_id)
     # pylint: disable=protected-access
-    task._top(fout)
-    return 0
+    return task._top(fout)
 
 
 def register(parser):
--- a/inductiva/_cli/cmd_tasks/top.py
+++ b/inductiva/_cli/cmd_tasks/top.py
@@ -16,7 +16,10 @@ def top(args: argparse.Namespace, fout: TextIO = sys.stdout):
     task_id = args.id
     task = tasks.Task(task_id)
     # pylint: disable=protected-access
-    return task._top(fout)
+    result, return_code = task._top(fout)
+    if result:
+        print(result, file=fout)
+    return return_code
 
 
 def register(parser):
--- a/inductiva/_cli/cmd_tasks/top.py
+++ b/inductiva/_cli/cmd_tasks/top.py
@@ -16,7 +16,7 @@ def top(args: argparse.Namespace, fout: TextIO = sys.stdout):
     task_id = args.id
     task = tasks.Task(task_id)
     # pylint: disable=protected-access
-    result, return_code = task._top(fout)
+    result, return_code = task._top()
     if result:
         print(result, file=fout)
     return return_code
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -1221,21 +1221,12 @@ class Task:
             A string with the formatted directory listing.
             The return code for the command. 0 if successful, 1 if failed.
         """
-        buffer = io.StringIO()
 
-        ret_code = self._run_streaming_command(lambda: self._file_operation(
-            Operations.LIST, formatter=self._format_directory_listing),
-                                               fout=buffer)
-        # if the command failed, print the error message
-        # and return None
-        if ret_code != 0:
-            print(
-                f"Error: {buffer.getvalue()}",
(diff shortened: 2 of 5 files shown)


inductiva/tasks/task.py
```diff
@@ -1543,6 +1494,33 @@ def _get_summary(self) -> str:
...
+                               generator_factory: Callable[[], AsyncGenerator],
+                               fout: TextIO = sys.stdout):
+        if not self._validate_task_computation_started():
+            return 1
+
+        if inductiva.is_notebook():
+            nest_asyncio.apply()
+
+        asyncio.run(self._gather_and_consume([generator_factory()], fout))
+        return 0
```
bgregoinductiva, line 1522: Could this function return `False` instead of `1`? And could it return the result of the command instead of `0`? (We are using the result of list files in e2e tests) Same for `_run_multiple_streaming_commands`.

Pbarbosa92: As for the first part. Yes i can change to False or True. I returned 0 or 1 because the CLI expects to return ints as error codes, but python will just convert True to 0 and False to 1 (i think).

As for returning the result of asyncio.run, I'm not sure about that. Like i said, the cli expects an error code as a result, as I am not sure on the returns we will get from the generators i think its safer to return false or true.

Pbarbosa92: Well, nvm.

I tested this. Python converts True=1 and False=0. So when the method fails we return True, and when it succeeds we return True. This, to me, is confusing, so I rather return the ints.

With the ints its easier to understand what is happening. Return 0 = all went well, return != 0 we got some problem.

bgregoinductiva: Ok. Then could you add an additional argument to force the function to return the results of `list_files`? Or add another function that does that instead of just printing to `fout`.

Pbarbosa92: list_files returns the string with the result.

Returns a string with all the files, or returns 1 if there is an error.

bgregoinductiva: No, it returns 0.

```
FAILED ../tests/user_flow/test_input_resources.py::test_remote_assets[setup_workspace0] - AssertionError: Missing files: {'0'}
assert False
 +  where False = <built-in method issubset of set object at 0x7f24112d52a0>({'0'})
 +    where <built-in method issubset of set object at 0x7f24112d52a0> = {'files.zip', 'images.zip', 'object1.obj'}.issubset
```

inductiva/tasks/task.py
```diff
@@ -1543,6 +1519,35 @@ def _get_summary(self) -> str:
...
+        if inductiva.is_notebook():
+            nest_asyncio.apply()
+
+        asyncio.run(
+            self._gather_and_consume([gen() for gen in generator_factories],
+                                     fout))
+        return 0
+
+    def _run_streaming_command(self,
+                               generator_factory: Callable[[], AsyncGenerator]):
```
bgregoinductiva, line 1538: Please add return types.

inductiva/tasks/task.py
```diff
@@ -1408,11 +1377,18 @@ def _top(self, fout: TextIO = sys.stdout):
...
+                Operations.TOP, formatter=lambda _: _, follow=False))
+
+        if return_code != 0:
+            print(
+                f"Error: {result}",
+                file=sys.stderr,
+            )
+            return None, return_code
+
+        return result, return_code
```
bgregoinductiva, line 1391: Please add return types and remove `fout` argument since it is not used anymore.

inductiva/tasks/task.py
```diff
@@ -1228,67 +1211,37 @@ async def close_stream(self):
...
             await self.file_tracker.cleanup()
 
-    async def _list_files(self) -> str:
-        directory = [
-            files async for files in self._file_operation(
-                Operations.LIST, formatter=self._format_directory_listing)
-        ]
-        return directory[0]
-
     def list_files(self):
```
bgregoinductiva, line 1214: Please add return types.

inductiva/_cli/cmd_tasks/top.py
```diff
@@ -16,8 +16,7 @@ def top(args: argparse.Namespace, fout: TextIO = sys.stdout):
     task_id = args.id
     task = tasks.Task(task_id)
     # pylint: disable=protected-access
-    task._top(fout)
-    return 0
+    return task._top(fout)
```
bgregoinductiva, line 19: Now it's necessary to explicitly print the result of `_top`

bgregoinductiva (approved):

PR #2007, opened 2025-04-07
Simplifying Task submit

hpenedones:
Removed some methods to construct the input zip and input.json that were too generic (from about 2 years ago, when we thought we would serialize things like numpy arrays and stuff and deserialize back again). Now we just need to zip the input_dir and pass some additional keyword arguments (kwargs) via the input.json file.

In particular, we no longer need:
- type_annotations
- pack_param
- get_validate_request_params
- unpack_value
- unpack_output

and also got rid of a method that was a very thin wrapper and I think we can call the submit_task directly:
- invoke_async_api 

Renamed some things to be more consistent with the names we use in the public methods:
- input_resources ---> remote_assets
- resource_pool ---> machine_group

and other generic things like:

- request_params ---> kwargs


inductiva/api/methods.py
```diff
@@ -21,8 +21,8 @@
 from inductiva.client.apis.tags.tasks_api import TasksApi
 from inductiva.client.models import (TaskRequest, TaskStatus, TaskSubmittedInfo,
                                      CompressionMethod)
-from inductiva import types, constants, storage
-from inductiva.utils.data import (get_validate_request_params, pack_input)
+from inductiva import constants, storage
+from inductiva.utils.data import pack_input
```
HenriquePreto, line 25: Suggestion: we could use ```data.pack_input``` instead of just ```pack_input```, as it belongs to a different module. And, we are already have an import stmt:```from inductiva.utils import ...```

inductiva/api/methods.py
```diff
@@ -315,134 +315,98 @@ def task_info_str(
...
 
     current_project = inductiva.projects.get_current_project()
     if current_project is not None:
         if not current_project.opened:
             raise RuntimeError("Trying to submit a task to a closed project.")
         current_project = current_project.name
 
-    if not input_resources:
-        input_resources = []
+    if not remote_assets:
```
HenriquePreto, line 356: It seems that ```remote_assets``` should simply be a ```List[str]``` and not ```Optional```. We could then set it to an empty list by default in the function signature, which eliminates the need for this if statement.

I.e.: ```remote_assets: List[str] = []```

inductiva/api/methods.py
```diff
@@ -315,134 +315,98 @@ def task_info_str(
...
 
 
-def submit_task(api_instance,
-                simulator,
-                request_params,
-                resource_pool,
+def submit_task(simulator,
+                input_dir,
+                machine_group,
+                kwargs,
```
HenriquePreto, line 321: Shouldn't ```kwargs``` be called ```**kwargs``` in the function signature? And placed last in the parameters list?

inductiva/tasks/run_simulation.py
```diff
@@ -12,51 +12,39 @@ def run_simulation(
...
-    type_annotations = {
-        "sim_dir": pathlib.Path,
-    }
-
-    if api_invoker is None:
-        api_invoker = methods.invoke_async_api
-
-    if not input_resources:
-        input_resources = []
+    if not remote_assets:
```
HenriquePreto, line 25: Same logic to ```remote_assets``` could be applied here -- setting it to an empty list by default instead.

inductiva/api/methods.py
HenriquePreto: I didn't know this:

```python
def append_to_list(item, my_list=[]):
    my_list.append(item)
    return my_list

print(append_to_list(1))  # [1]
print(append_to_list(2))  # [1, 2] <-- This is not expected!
```

Thanks @luispcunha

inductiva/api/methods.py
hpenedones: right, but here I don't want the caller of "submit_tasks" to be able to pass aditional kwargs, these are the kwargs that were passed to the run_simulation method that invokes this one... But I will revert back to the original name "request_params", to avoid confusion...

HenriquePreto (approved):

bgregoinductiva (approved):
This is a breaking change in the client interface. We'll need to bump the minimum supported version in BE.

hpenedones:
> This is a breaking change in the client interface. We'll need to bump the minimum supported version in BE.

Hmmmm.. wait, can you help me clarify that?
In my mind:

1) I am not intending to change behaviour of public methods, it is basically a refactoring / clean-up. 
2) However, it is possible that we were exposing some methods publicly that should not be public -- and now they are gone for real.
3) or that I did some simple arguments renamings that would cause a "breaking change" for the user if they are passing arguments naming them explicitly in some user script-- if so, can you help me spot those?
4) even if this PR could suffer from problems 2 or 3 (to be confirmed) -- I think we don't need to bump the minimum version on the backend, right? that would be more for the case where the backend changes in a way that the client would break. But here the backend did not change...

bgregoinductiva:
> > This is a breaking change in the client interface. We'll need to bump the minimum supported version in BE.
> 
> Hmmmm.. wait, can you help me clarify that? In my mind:
> 
> 1. I am not intending to change behaviour of public methods, it is basically a refactoring / clean-up.
> 2. However, it is possible that we were exposing some methods publicly that should not be public -- and now they are gone for real.
> 3. or that I did some simple arguments renamings that would cause a "breaking change" for the user if they are passing arguments naming them explicitly in some user script-- if so, can you help me spot those?
> 4. even if this PR could suffer from problems 2 or 3 (to be confirmed) -- I think we don't need to bump the minimum version on the backend, right? that would be more for the case where the backend changes in a way that the client would break. But here the backend did not change...

I was referring to the renaming of these parameters:
- `computational_resources` => `machine_group`
- `input_resources` => `remote_assets`

But yes, you're right. I misinterpreted, it's not a breaking change, it's possible to continue to use the older client version. Nevertheless, we should update the documentation.

hpenedones:
> > > This is a breaking change in the client interface. We'll need to bump the minimum supported version in BE.
> > 
> > 
> > Hmmmm.. wait, can you help me clarify that? In my mind:
> > 
> > 1. I am not intending to change behaviour of public methods, it is basically a refactoring / clean-up.
> > 2. However, it is possible that we were exposing some methods publicly that should not be public -- and now they are gone for real.
> > 3. or that I did some simple arguments renamings that would cause a "breaking change" for the user if they are passing arguments naming them explicitly in some user script-- if so, can you help me spot those?
> > 4. even if this PR could suffer from problems 2 or 3 (to be confirmed) -- I think we don't need to bump the minimum version on the backend, right? that would be more for the case where the backend changes in a way that the client would break. But here the backend did not change...
> 
> I was referring to the renaming of these parameters:
> 
> * `computational_resources` => `machine_group`
> * `input_resources` => `remote_assets`
> 
> Yes, you're right. Sorry, I misinterpreted, it's not a breaking change, it's possible to continue to use the older client versions.

Thanks for double checking and clarifying, Bruno -- but in that case I think it is fine, because that change was made on an internal method, the tasks.run_simulation(...) --- but the users are not supposed to call it directly. They call the simulator.run, and the interface to that one was not changed.

inductiva/tasks/run_simulation.py
```diff
@@ -12,51 +12,39 @@ def run_simulation(
...
+    remote_assets: Optional[List[str]] = None,
     simulator_name_alias: Optional[str] = None,
     **kwargs: Any,
 ) -> tasks.Task:
     """Run a simulation via Inductiva Web API."""
 
-    params = {
-        "sim_dir": input_dir,
-        **kwargs,
-    }
```
HenriquePreto, line 29: I think this shouldn't be deleted

inductiva/tasks/run_simulation.py
```diff
@@ -12,51 +12,39 @@ def run_simulation(
...
+    if (machine_group.allow_auto_start and not machine_group.started):
         logging.info("\n■ The computational resource is not started."
                      " Starting it now.\n")
-        computational_resources.start()
+        machine_group.start()
+
+    task_id = methods.submit_task(simulator,
+                                  input_dir=input_dir,
+                                  machine_group=machine_group,
+                                  params=kwargs,
```
HenriquePreto, line 38: Shouldn't be ```params=params```?

inductiva/tasks/run_simulation.py
HenriquePreto: Because it breaks the task-runner

hpenedones: Did it happen? Wouldn't some test would have caught it?

HenriquePreto: On the client side there's no test that runs with the task-runner. I think @Pbarbosa92

HenriquePreto: I tried to run the task runner locally and it raised an exception: args.sim_dir not found

hpenedones: Ok, thanks - I can't do it right now, do you mind applying a fix?

HenriquePreto: Sure

PR #2010, opened 2025-04-08
delete default queue prints

hpenedones:


inductiva/api/methods.py
```diff
@@ -301,16 +301,13 @@ def task_info_str(
...
-        info_str += f" \t\t· Default queue with {machine_type} machines.\n"
-        ttl_seconds = task_submitted_info.get("time_to_live_seconds")
-        if ttl_seconds is not None and isinstance(ttl_seconds, decimal.Decimal):
-            ttl_seconds = format_utils.seconds_formatter(ttl_seconds)
-            info_str += (f" \t\t· Task will be killed after the computation "
-                         f"time exceeds {ttl_seconds} (h:m:s).\n")
+    info_str += f" \t\t· {resource_pool}\n"
+    
+    ttl_seconds = task_submitted_info.get("time_to_live_seconds")
+    if ttl_seconds is not None and isinstance(ttl_seconds, decimal.Decimal):
```
hpenedones, line 307: for some reason, this part related to the ttl_seconds info was nested inside the else of default_queue -- can I leave it or should I delete as well? @bgregoinductiva @luispcunha

bgregoinductiva: It depends. Right now the client doesn't allow specifying this parameter so it will always be null. When the default queue existed the BE returned the default TTL for the task. In my opinion we should keep this code and modify the `simulator.run` function to be able to specify `time_to_live_seconds`.

hpenedones: Ok -- for now I am just going to add another validation to check that task_submitted_info is not None...

PR #2014, opened 2025-04-10
Add cleanup_commands to coawst.

Pbarbosa92:
This list of commands will run after the simulation finishes and will allow the user to delete unwanted files, OR (this is why i added it) will allow the user to copy the compiled binary to the simulation files for later use in other simulations.

Because we want to allow the user to copy back the binary and re-use it in other simulations we need some extra arguments like -> simulation_binary that, if used will skip the compilation.

### DIFF
--- a/inductiva/simulators/coawst.py
+++ b/inductiva/simulators/coawst.py
@@ -56,13 +56,13 @@ class COAWST(simulators.Simulator):
     def run(self,
             input_dir: Optional[str],
             sim_config_filename: str,
-            build_coawst_script: str,
             *,
+            build_coawst_script: str = None,
             on: types.ComputationalResources,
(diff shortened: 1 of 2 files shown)


bgregoinductiva (approved):

Pbarbosa92:
> > Because we want to allow the user to copy back the binary and re-use it in other simulations we need some extra arguments like -> simulation_binary that, if used will skip the compilation.
> 
> Are you adding this, or will the combination of "coawst_bin" and "compile_simulator" serve this purpose?

It will be a combo from:
- `cleanup_commands` -> used to copy the binary to the outputs (to be used in another simulations)
- `compile_simulator` -> Set to false to use the binary present in the input dir ( needs to have the same name as `coawst_bin`

PR #2032, opened 2025-04-15
Events

rcvalerio:
closes https://github.com/inductiva/tasks/issues/1043

Currently, a single trigger is supported: TaskOutputUploaded, which is called when the task uploads its output—regardless of whether the task succeeds or fails.

```python
import inductiva
from inductiva.events import triggers, actions

event_id = inductiva.events.register(
    triggers.TaskOutputUploaded(task_id="<task_id>"),
    actions.EmailNotification("<email_address>")
)

event_id = inductiva.events.register(
    triggers.TaskOutputUploaded(task_id="<task_id>"),
    actions.WebhookNotification("<webhook_url>")
)

inductiva.events.list_events()

inductiva.events.remove("<event_id>")
```

To get a url for webhook testing: https://webhook.site/

pip install git+https://github.com/inductiva/inductiva.git@rv-add-events


inductiva/events/event.py
```diff
@@ -0,0 +1,69 @@
+"""Project class"""
```
hpenedones, line 1: Event module

hpenedones (commented):
Great! Do you think you could add a couple of tests? I know that it will be hard to test the event itself here -- but at a simple test invoking the register, get and remove we should be able to test, no?

hpenedones (approved):

PR #2033, opened 2025-04-15
Ls availabe img

ayomidetobi:


inductiva/simulators/methods.py
```diff
@@ -1,62 +1,12 @@
...
-            images[branch][name].append(version)
-
-        url = data["next"]
-
-    return images
+def list_available_images() -> Any:
+    """Fetch the list of available simulator images from the API."""
+    api = SimulatorsApi(inductiva.get_client())
+    response = api.list_available_images(skip_deserialization=True).response
+    return json.loads(response.data.decode("utf-8"))
```
hpenedones, line 12: does this return exactly the same thing as the method before moving it to the backend? (in terms of the data type and format)?

inductiva/simulators/methods.py
```diff
@@ -1,62 +1,12 @@
...
-                continue
-
-            name, version = match.group("name"), match.group("version")
-            branch = "development" if match.group("isdev") else "production"
-            images[branch][name].append(version)
-
-        url = data["next"]
-
-    return images
+def list_available_images() -> Any:
```
hpenedones, line 8: Annotating the return data type as "Any" is not really informative -- what type does it truly have?

ayomidetobi: Okay ,Noted

hpenedones (approved):

PR #2034, opened 2025-04-15
Fix opening a benchmark before running

HenriquePreto:


inductiva/benchmarks/benchmark.py
```diff
@@ -161,7 +164,7 @@ def wait(self) -> Self:
         Returns:
             Self: The current instance for method chaining.
         """
-        tasks = self.get_tasks()
+        tasks = self.list()
```
hpenedones, line 167: do we need this change? I think I will delete in on my PR by the way...

PR #2036, opened 2025-04-15
Deploy to prod v0.15.7

bgregoinductiva:


hpenedones:
A test that I expanded recently is failing on the CLI tests on different OS because of some SSL certificate failure.... It's weird because I don't think I changed the url to download from -- maybe something changed on that server?

rcvalerio (approved):

bgregoinductiva:
> A test that I expanded recently is failing on the CLI tests on different OS because of some SSL certificate failure.... It's weird because I don't think I changed the url to download from -- maybe something changed on that server?

I seems you uncovered a problem in Windows, but the problem doesn't seem to be related with any recent change. I've been trying to solve this problem without success so far, so I think it should not block deploy to production.

I believe this pipeline does not affect the repo badges, right @Pbarbosa92?

Pbarbosa92:
> > A test that I expanded recently is failing on the CLI tests on different OS because of some SSL certificate failure.... It's weird because I don't think I changed the url to download from -- maybe something changed on that server?
> 
> I seems that you uncovered a problem in Windows, but the problem doesn't seem to be related with any recent change. I've been trying to solve this problem without success so far, so I think it should not block deploy to production.
> 
> I believe this pipeline does not affect the repo badges, right @Pbarbosa92?

I don't think it will affect the badges.

hpenedones (approved):

PR #2055, opened 2025-04-22
Add Task's `set_metadata` and `get_metadata`

catarinalegria:
related to [#1010](https://github.com/inductiva/tasks/issues/423)
related to [#423](https://github.com/inductiva/tasks/issues/1010)

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -1549,3 +1549,61 @@ class Task:
 
     def print_summary(self, fhandle=sys.stdout):
         print(self._get_summary(), file=fhandle)
+
+    def set_metadata(self, metadata: Dict[str, str]):
+        """Set metadata for the task.
+        
+        Metadata is stored as key-value pairs, where both
+        keys and values must be strings.
+        Metadata can be useful for categorizing, searching,
+        and filtering tasks.
+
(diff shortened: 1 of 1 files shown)


inductiva/tasks/task.py
```diff
@@ -1549,3 +1549,57 @@ def summary(self) -> str:
...
+            if not isinstance(key, str) or not isinstance(value, str):
+                raise TypeError("Metadata keys and values must be strings.")
+            if not key or not value:
+                raise ValueError(
+                    "Metadata keys and values cannot be empty strings.")
+
+        try:
+            self._api.set_metadata(path_params={"task_id": self.id},
+                                   body=metadata)
+        except exceptions.ApiException as exc:
```
Pbarbosa92, line 1588: We already catch ApiExceptions and print the details that come from the back end. The back end just needs to send us the exceptions with a detail that describes the issue.

I don't think we need to catch it here.

Pbarbosa92 (changes requested):

rcvalerio (approved):
LGTM, just remove the try catch as suggested by @Pbarbosa92

Pbarbosa92 (approved):

PR #2057, opened 2025-04-24
Pass `project` in simulator.run()

catarinalegria:
continuation of https://github.com/inductiva/inductiva/pull/2001

### DIFF
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -343,8 +343,9 @@ def submit_task(simulator,
         simulator_obj: Optional simulator object with additional configuration
         remote_assets: Additional input files that will be copied to the
                 simulation from a bucket or from another task output.
-        project_name: Name of the project to which the task will be assigned.
-                If None, the task will be assigned to the default project.
+        project_name: Name of the project to which the task will be
+                assigned. If None, the task will be assigned to
+                the default project.
     Return:
         Returns the task id.
     """
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -343,6 +343,8 @@ def submit_task(simulator,
         simulator_obj: Optional simulator object with additional configuration
         remote_assets: Additional input files that will be copied to the
                 simulation from a bucket or from another task output.
+        project_name: Name of the project to which the task will be assigned.
+                If None, the task will be assigned to the default project.
     Return:
         Returns the task id.
     """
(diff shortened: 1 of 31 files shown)


Pbarbosa92 (commented):
I'm a bit confused on why we need this change.

If we pass project_name on `amr-wind.run()` will the back end create the project for us ?

catarinalegria:
> I'm a bit confused on why we need this change.
> 
> 
> 
> If we pass project_name on `amr-wind.run()` will the back end create the project for us ?

no, this creates the task and assigns it to the (existing) project you pass in project_name. if no project is passed, the task will be created in the default project.

without this, you have no way of assigning the task to a project without doing project.add_task(...) or using the web console.

Pbarbosa92:
> > I'm a bit confused on why we need this change.
> > If we pass project_name on `amr-wind.run()` will the back end create the project for us ?
> 
> no, this creates the task and assigns it to the (existing) project you pass in project_name. if no project is passed, the task will be created in the default project.
> 
> without this, you have no way of assigning the task to a project without doing project.add_task(...) or using the web console.

And what happens if the project does not exist ?

catarinalegria:
> > > I'm a bit confused on why we need this change.
> 
> > > If we pass project_name on `amr-wind.run()` will the back end create the project for us ?
> 
> > 
> 
> > no, this creates the task and assigns it to the (existing) project you pass in project_name. if no project is passed, the task will be created in the default project.
> 
> > 
> 
> > without this, you have no way of assigning the task to a project without doing project.add_task(...) or using the web console.
> 
> 
> 
> And what happens if the project does not exist ?

it throws an error.

rcvalerio:
> > And what happens if the project does not exist ?
> 
> it throws an error.

Would it be possible to create the project instead of throwing an error? Wdyt of this behavior?

catarinalegria:
> > > And what happens if the project does not exist ?
> > 
> > 
> > it throws an error.
> 
> Would it be possible to create the project instead of throwing an error? Wdyt of this behavior?

it would be possible yes, but i am not sure if this is the intended behaviour (the user could start creating unintended projects). 

right now if the user chooses a project that does not exist we get the following error msg:

`ERROR: Project not found: test-dont-exists`

rcvalerio (approved):

catarinalegria:
@hpenedones @HenriquePreto can this PR be merged?

hpenedones:
Ah, I guess so -- but not sure if my initial choice of "project_name" should just be "project" -- do we have other places where we use just "project" (as a str, assuming it's the project name)? Would be good to be consistent....

hpenedones (approved):

PR #2064, opened 2025-04-30
Prevent Commands from having special characters

Pbarbosa92:
Since we are running commands with:

subprocess.run with `shell=False` our commands can't have special characters. We now check for them in the python client to avoid having strange things happen to the machines.

### DIFF
--- a/inductiva/commands/commands.py
+++ b/inductiva/commands/commands.py
@@ -6,7 +6,6 @@ from .mpiconfig import MPIConfig
 
 class Command:
     """Abstraction class for commands."""
-
     SPECIAL_CHARACTERS = r"[|><&;*?~$]"
     SPECIAL_CHARACTERS_PRETTY_PRINT = r"| > < & ; * ? ~ $"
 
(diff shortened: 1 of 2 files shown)


rcvalerio (approved):

inductiva/commands/commands.py
```diff
@@ -23,11 +26,25 @@ def __init__(self, cmd: str, *prompts: str, mpi_config: MPIConfig = None):
...
+        if self._has_special_chars(cmd):
+            raise ValueError(
+                "Command contains unsupported characters.\n"
+                "Please avoid using any of the following characters:\n"
+                f"{self.SPECIAL_CHARACTERS_PRETTY_PRINT}\n")
         self.cmd = cmd
         self.prompts = list(prompts)
         self.mpi_config = mpi_config
 
+    def _has_special_chars(self, command):
```
hpenedones, line 38: we should probably have some unit tests for this... wdyt?

Pbarbosa92: > we should probably have some unit tests for this... wdyt?

Added them here https://github.com/inductiva/inductiva/pull/2073

PR #2066, opened 2025-04-30
Update personal docker quality of life

Franreno:
There were a few bugs that were harming the quality of life for the personal docker usage.

### 1 Containers hanging

The apptainer-converter container wouldn't be removed after the conversion was finished. Users needed to do `docker ps -a` and `docker rm apptainer-converter-id`

To solve this, i've added a finally clause to remove the apptainer-converter container used in this session.

---

### 2 Multiple conversions

Users couldn't convert/upload multiple containers at the same time. The `apptainer-converter` container was named after itself, meaning that only one could exist at any given time.

To solve this, i've added a mini unique id to the end of the `apptainer-converter` name, so now it should allow more multiple conversions to happen at the same time. 

---

### 3 Overwrite flag

Users couldn't update existing containers on their remote storage. They had to use `inductiva containers rm` and then upload the new container.

Now, the upload command checks if the container already exists. If it does, then it will prompt the user if it wants to overwrite it.

```bash
➜  inductiva containers upload docker://nginx:alpine   
Overwrite the existing file?
  - nginx.sif
Are you sure you want to proceed (y/[N])? 
```

Or if the users use the `--overwrite / -f` flag this confirmation is skipped.

```bash
➜  inductiva containers upload docker://nginx:alpine --overwrite
⚠️  'my-containers/nginx.sif' exists - will overwrite it (because --overwrite).
Removing workspace file(s)...
Workspace file(s) removed successfully.
Converting docker://nginx:alpine -> /Users/franreno/.inductiva/tmp/tmp3aet9uld/my-containers/nginx.sif...
Conversion started with container ID: 02f1387ff7ea
```

---

### 4 Tmp permission error

The upload command tried to use the /tmp directory (in linux). But if the user didn't had permissions to write at that location the command would fail.

To solve this, I've updated the command to have a tmp dir inside of the HOME_DIR of the User. Meaning that the conversion and upload will create a temporary file at `/home/user/.inductiva/tmp` instead of `/tmp/`, which should bypass this permission error.

### DIFF
--- a/inductiva/_cli/cmd_containers/upload.py
+++ b/inductiva/_cli/cmd_containers/upload.py
@@ -4,8 +4,11 @@ Uploads a Docker image as a converted Apptainer .sif to remote storage.
 
 import argparse
 import os
+import pathlib
+import sys
 import tempfile
-from inductiva import storage
+from inductiva import storage, constants
+from inductiva.utils.input_functions import user_confirmation_prompt
 from .convert import convert_image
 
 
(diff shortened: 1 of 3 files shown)


inductiva/_cli/cmd_containers/upload.py
```diff
@@ -51,10 +54,41 @@ def upload_container(args):
     folder_name = output_path.split(os.sep)[0]
     filename = os.path.basename(output_path)
 
+    # ---------- pre-flight: does it already exist? ---------------------------
+    try:
+        contents = storage.listdir(folder_name, print_results=False)
+    except Exception as e:  # pylint: disable=broad-except
```
Pbarbosa92, line 60: Won't this return an APIException ?

inductiva/_cli/cmd_containers/upload.py
```diff
@@ -51,10 +54,41 @@ def upload_container(args):
...
+        print(f"Error accessing remote folder '{folder_name}': {e}",
+              file=sys.stderr)
+        return
+
+    already_there = any(c["content_name"] == filename for c in contents)
+    if already_there:
+        if args.overwrite:
+            print(f"⚠️  '{output_path}' exists - will overwrite it "
```
Pbarbosa92, line 68: What will happen to this message (referring to the emoji) if the user runs this on a terminal that does not support emojis ?

inductiva/_cli/cmd_containers/upload.py
```diff
@@ -51,10 +54,41 @@ def upload_container(args):
...
+                  "(because --overwrite).")
+            try:
+                storage.remove_workspace(f"{folder_name}/{filename}")
+            except Exception as e:  # pylint: disable=broad-except
```
Pbarbosa92, line 72: Again, wont the back end return API Exceptions ?

Pbarbosa92 (approved):
LGTM, only a few comments.

inductiva/_cli/cmd_containers/upload.py
Franreno: usually it prints the unicode warning symbol

“⚠” (U+26A0)

however, if the terminal does not support unicode i don't know what it will print, so i removed it.

PR #2074, opened 2025-05-05
Add gpu support for cans and fixed a little bug

Pbarbosa92:
This PR adds gpu support to cans

and fixes a little bug where we were adding _gpu suffix without checking if the image actually exists.

### DIFF
--- a/inductiva/simulators/cans.py
+++ b/inductiva/simulators/cans.py
@@ -1,5 +1,5 @@
 """CaNS module of the API for numerical simulations of fluid flows."""
-from typing import List, Optional
+from typing import List, Literal, Optional
 
 from inductiva import types, tasks, simulators
 from inductiva.commands.commands import Command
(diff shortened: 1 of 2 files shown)


inductiva/simulators/cans.py
```diff
@@ -12,7 +12,11 @@ class CaNS(simulators.Simulator):
 
     """
 
-    def __init__(self, /, version: Optional[str] = None, use_dev: bool = False):
+    def __init__(self,
+                 /,
+                 version: Optional[str] = None,
+                 use_dev: bool = False,
```
hpenedones, line 18: do we still need this use_dev?

hpenedones (approved):

inductiva/simulators/cans.py
Pbarbosa92: Yes. This allows me to use images that are only in dev (to test before sending to prod).

rcvalerio (approved):

PR #2076, opened 2025-05-06
Update docs sources

arasahcor:
This PR updates docs-sources with the most recent version

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,17 +2,3 @@
 pylint>=2.17,<3.0
 pytest
 yapf
-sphinx
-sphinx-autobuild
-sphinx_rtd_theme
-sphinx-book-theme
-sphinxcontrib-mermaid
-myst-parser @ git+https://github.com/executablebooks/MyST-Parser@fc64840b5f009021d64173d81c2f244fb0d8a43d
-sphinx-copybutton
-sphinx-tabs
-sphinx-togglebutton
-sphinxcontrib-googleanalytics
-sphinxext.opengraph
-sphinx-sitemap
-sphinx-reredirects
-sphinxawesome-theme


.github/workflows/deploy-docs-to-staging-new.yml
Pbarbosa92, line 1: I don't like this. This action is pushing things to dev. This will create a lot of merge conflits in the future. Can't we deploy the docs to staging from the current branch ? instead of comiting to dev ?

requirements.txt
Pbarbosa92, line 1: Won't this bloat our package installation ?

Pbarbosa92 (changes requested):

.github/workflows/deploy-docs-to-staging-new.yml
hpenedones: Ah, good catch, I overlooked that! Yes, let's avoid making commits to repos from the Github actions -- what was this needed for @arasahcor ?

hpenedones: I think that for now we don't need to have a generic github action to deploy to staging from a particular branch. It is enough to go to staging when the changes are merged to development. And if the person wants to have a local build, then it could either run sphinx locally, or run the website-new locally with npm -- wdyt @arasahcor ?

hpenedones: In addition, it is fine to often merge changes to the docs to the development branch, because it does not imply any code release...

requirements.txt
```diff
@@ -2,3 +2,17 @@
 pylint>=2.17,<3.0
 pytest
 yapf
+sphinx
+sphinx-autobuild
+sphinx_rtd_theme
+sphinx-book-theme
+sphinxcontrib-mermaid
+myst-parser @ git+https://github.com/executablebooks/MyST-Parser@fc64840b5f009021d64173d81c2f244fb0d8a43d
```
hpenedones, line 10: I think this particular mention to a commit is no longer needed... Do you mind testing just with "myst-parser"?

Pbarbosa92 (approved):

PR #2083, opened 2025-05-06
Don't setup sigint handler if not supported

luispcunha:
Signal handlers can't be set if not on the main thread of the process. Suppress the error thrown, so that the package can be used in multi threading environments.

### DIFF
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -231,9 +231,13 @@ def block_until_status_is(api_instance: TasksApi,
 
 
 def _configure_sigint_handler(handler):
+    if not handler:
+        return None
+
     try:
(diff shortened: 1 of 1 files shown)


inductiva/api/methods.py
```diff
@@ -230,6 +230,14 @@ def block_until_status_is(api_instance: TasksApi,
     return status
 
 
+def _configure_sigint_handler(handler):
+    try:
+        return signal.signal(signal.SIGINT, handler)
+    except ValueError:
+        # If the signal is not supported, ignore it
```
hpenedones, line 237: should we show some log message?

hpenedones (approved):
Awesome, that was fast! :)

bgregoinductiva (approved):

PR #1091, opened 2024-02-14
Instructions on how to remove inductiva

AugustoPeres:
This pull request adds instruction on how to uninstall the inductiva package. This will be even more important when we have things like [autocomplete](https://github.com/inductiva/inductiva/pull/1090)(still experimental) that will write to the users `.zshrc` or `.bashrc`.

More information and discussion: https://github.com/inductiva/tasks/issues/123

PR #2089, opened 2025-05-07
Improve Benchmark logging

catarinalegria:
closes [#1091](https://github.com/inductiva/tasks/issues/1091)

I need some feedback. I am running the following example (with very little runs -- only 3 Machine Groups):

```
from inductiva import benchmarks, simulators, resources

benchmark = benchmarks.Benchmark(name="benchmark-improve-logging") \
    .set_default(simulator=simulators.SWASH(),
                 input_dir="model_input/",
                 sim_config_filename="S1.sws") \
    .add_run(on=resources.MachineGroup("n2d-highcpu-32")) \
    .add_run(on=resources.MachineGroup("c3-highcpu-22")) \
    .add_run(on=resources.MachineGroup("n2-highcpu-32")) \

benchmark.run(num_repeats=2)

benchmark.wait()
```

and this is the current logging situation, with some big improvements already (I killed all the tasks for the benchmark to complete faster):

https://github.com/user-attachments/assets/9dd9ec84-0e41-4964-b908-2d9edb8e7674

I believe the logging could still be improved:

- should I remove even further the initial logging of the MGs / tasks / download input... ?
  - for example, to only have more generic logging: Setting up resources, Preparing Tasks, etc... and then the Benchmark progress bar?
- There is also the scenario where the user gets stuck "waiting for quotas" to start MGs. In that case the final Benchmark logging is delayed and a lot of tasks have already finished when it is displayed

### DIFF
--- a/inductiva/benchmarks/benchmark.py
+++ b/inductiva/benchmarks/benchmark.py
@@ -160,7 +160,7 @@ class Benchmark(projects.Project):
         self.runs.clear()
         logging.info(
             "Benchmark \033[1m%s\033[0m has started...\n"
-            "Go to https://console.inductiva.ai/projects/%s?project=%s "
+            "Go to https://console.inductiva.ai/projects/%s "
             "for more details.\n", self.name, self.name, self.name)
         return self
 
--- a/inductiva/benchmarks/benchmark.py
+++ b/inductiva/benchmarks/benchmark.py
@@ -5,6 +5,7 @@ import csv
 import logging
 from typing import Optional, Union
 import tqdm
+from tqdm.contrib.logging import logging_redirect_tqdm
 from typing_extensions import Self
 from collections import defaultdict
 from inductiva import types, resources, projects, simulators, client
@@ -176,7 +177,8 @@ class Benchmark(projects.Project):
                      self.name)
 
         with tqdm.tqdm(total=len(running_tasks),
-                       desc="Processing tasks") as pbar:
+                       desc="Processing tasks",
+                       bar_format='{l_bar}{bar}| {n_fmt}/{total_fmt}') as pbar:
             with ThreadPoolExecutor(max_workers=len(running_tasks)) as executor:
                 future_to_task = {
                     executor.submit(
@@ -186,19 +188,20 @@ class Benchmark(projects.Project):
                 }
 
                 for future in as_completed(future_to_task):
-                    task = future_to_task[future]
-                    status = future.result()
-                    logging.info("Task %s completed with status: %s", task.id,
-                                 status)
+                    with logging_redirect_tqdm():
+                        task = future_to_task[future]
+                        status = future.result()
+                        logging.info("Task %s completed with status: %s",
+                                     task.id, status)
 
-                    if status != TaskStatusCode.SUCCESS:
-                        logging.error(
-                            "   · To understand why the task did not complete successfully "
-                            "go to https://console.inductiva.ai/tasks/%s",
-                            task.id)
+                        if status != TaskStatusCode.SUCCESS:
+                            logging.info(
+                                "   · To understand why the task did not complete "
+                                "successfully go to https://console.inductiva.ai/tasks/%s",
+                                task.id)
 
-                    # Update progress bar for each completed task
-                    pbar.update(1)
+                        # Update progress bar for each completed task
+                        pbar.update(1)
 
         return self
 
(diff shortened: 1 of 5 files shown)


Pbarbosa92 (approved):
LGTM.

Reagarding this "should I remove even further the initial logging of the MGs / tasks / download input... ?
for example, to only have more generic logging: Setting up resources, Preparing Tasks, etc... and then the Benchmark progress bar?"

We can have a flag to control how verbose the benchmark is. Normal mode can be like you said ("setting up resources...", "preparing tasks..." with some progress bars would be nice). And we could have a debug mode that prints everything for the user to know what is going on. Maybe you can check with Hugo or Sarmento on what a nice interface would look like

catarinalegria:
> LGTM.
> 
> Reagarding this "should I remove even further the initial logging of the MGs / tasks / download input... ? for example, to only have more generic logging: Setting up resources, Preparing Tasks, etc... and then the Benchmark progress bar?"
> 
> We can have a flag to control how verbose the benchmark is. Normal mode can be like you said ("setting up resources...", "preparing tasks..." with some progress bars would be nice). And we could have a debug mode that prints everything for the user to know what is going on. Maybe you can check with Hugo or Sarmento on what a nice interface would look like

I really like that idea, thanks !! Will do that (probably in a seperate pr)

inductiva/benchmarks/benchmark.py
```diff
@@ -162,9 +171,40 @@ def wait(self) -> Self:
...
-            task.wait(download_std_on_completion=False)
+        running_tasks = self.get_tasks()
+
+        logging.info("Waiting for Benchmark \033[1m%s\033[0m to complete...\n",
+                     self.name)
+
+        with tqdm.tqdm(total=len(running_tasks),
+                       desc="Processing tasks",
+                       bar_format="{l_bar}{bar}| {n_fmt}/{total_fmt}") as pbar:
+            with ThreadPoolExecutor(max_workers=len(running_tasks)) as executor:
```
HenriquePreto, line 182: Why set `max_workers=len(running_tasks)`? We should allow `max_workers` to be determined automatically based on the number of CPU cores available -- the ctor does that.

inductiva/benchmarks/benchmark.py
```diff
@@ -162,9 +171,40 @@ def wait(self) -> Self:
...
+                future_to_task = {
+                    executor.submit(
+                        lambda t: t.wait(download_std_on_completion=False,
+                                         silent_mode=True), task):
+                        task for task in running_tasks
+                }
+
+                for future in as_completed(future_to_task):
+                    with logging_redirect_tqdm():
```
HenriquePreto, line 191: Why using `logging_redirect_tqdm`?

inductiva/benchmarks/benchmark.py
```diff
@@ -162,9 +171,40 @@ def wait(self) -> Self:
...
+                            logging.info(
+                                "   · To understand why the task did not "
+                                "complete successfully go to "
+                                "https://console.inductiva.ai/tasks/%s",
+                                task.id)
+
+                        # Update progress bar for each completed task
+                        pbar.update(1)
+
+        print("\n")
```
HenriquePreto, line 207: I think here you should use the `logging` module to log and not `print` stmts.

inductiva/tasks/run_simulation.py
```diff
@@ -19,6 +19,7 @@ def run_simulation(
     remote_assets: Optional[List[str]] = None,
     simulator_name_alias: Optional[str] = None,
     project_name: Optional[str] = None,
+    silent_mode: bool = False,
```
HenriquePreto, line 22: I think the most common parameter name for this is `verbose` and not `silent_mode`

inductiva/benchmarks/benchmark.py
catarinalegria: https://tqdm.github.io/docs/contrib.logging/#logging_redirect_tqdm

I am using this to not re-print the progress bar in each task's logging update

HenriquePreto: Got it.

inductiva/benchmarks/benchmark.py
```diff
@@ -162,9 +171,40 @@ def wait(self) -> Self:
         Returns:
             Self: The current instance for method chaining.
         """
-        tasks = self.get_tasks()
-        for task in tasks:
-            task.wait(download_std_on_completion=False)
+        running_tasks = self.get_tasks()
```
HenriquePreto, line 174: These might not be `running_tasks` -- for that you must pass a non-terminal `status` to the `get_tasks` method.

HenriquePreto (approved):

PR #1103, opened 2024-02-20
Ap add schism simulator

AugustoPeres:
Adds schism simulator to the front end.

PR #2128, opened 2025-05-14
Fix minor benchmark issues

catarinalegria:
closes [#1103](https://github.com/inductiva/tasks/issues/1103)

This PRs solves two small Benchmark issues:

- Enables `resubmit_on_preemption` in benchmark.run()
- Fixes local machines' names showing as "n/a" in benchmark.export()


catarinalegria:
@bgregoinductiva Would it be possible for this small change to be deployed in PROD? I need to run some benchmarks in PROD and need this change

HenriquePreto (commented):
What happens if the machine is not a spot instance? Is this parameter ignored?

catarinalegria:
> What happens if the machine is not a spot instance? Is this parameter ignored?

I believe so. I tested with a MG where spot was false and it seems to have no effect.

bgregoinductiva (approved):

PR #2141, opened 2025-05-15
Adding python docs for simulator classes

hpenedones:
Note: I had to change little pieces of code to make methods protected rather than public, so that they are not shown in the documentation (as they were not intended for direct usage, but only for internal behaviour).

### DIFF
--- a/docs-sources/python-client/index.md
+++ b/docs-sources/python-client/index.md
@@ -8,6 +8,7 @@ benchmarks <inductiva.benchmarks>
 projects <inductiva.projects>
 resources <inductiva.resources>
 storage <inductiva.storage>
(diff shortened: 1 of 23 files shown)


Pbarbosa92 (approved):

docs-sources/python-client/inductiva.simulators.rst
```diff
@@ -0,0 +1,216 @@
+simulators
```
hpenedones, line 1: @Pbarbosa92 : FYI, next time we add a simulator, we need to add a new subsection here -- this is not automatically finding them... We can also use this file to add more content in each section if we want.

PR #2151, opened 2025-05-20
Add 'verbose' to Benchmark

catarinalegria:
closes [#1111](https://github.com/inductiva/tasks/issues/1111)

This PR adds a new argument - `verbose` - to the Benchmark class, allowing the user to control the level of logging detail.
By default, a new Benchmark object is initialized with `verbose = False` ("silent mode", i.e, summarized logging).

## `verbose = False` (default)

https://github.com/user-attachments/assets/4de90220-6eb5-4207-b5d0-1ef1470482bd

## `verbose = True`

https://github.com/user-attachments/assets/f0aad4b0-24d8-4138-b8c7-08b3ecec2683


HenriquePreto:
A general comment about this implementation: every time `machine_group.start` and `submit_task` functions are called, the stack is inspected -- and these are among the most frequently called functions.

There are alternative ways to manage logs across different modules -- particularly by using the `logging` module.

HenriquePreto (commented):
We could add a `verbose` parameter to the `MachineGroup.start` method, with a default value of `True` to maintain backward compatibility. This would allow the `Benchmark` class to call `machine_group.start(wait_for_quotas=wait_for_quotas, verbose=False)` to suppress logs.

We could do the same for `simulator.run`. Wdyt?

catarinalegria:
> We could add a `verbose` parameter to the `MachineGroup.start` method, with a default value of `True` to maintain backward compatibility. This would allow the `Benchmark` class to call `machine_group.start(wait_for_quotas=wait_for_quotas, verbose=False)` to suppress logs.
> 
> 
> 
> We could do the same for `simulator.run`. Wdyt?

Yes! Believe the trickier part tho is the MachineGroup.register(). I cannot find a way to silence the logs without inspecting the stack.

The solution could simply be to not silence the register logs - since these need to be explicitly called in the benchmarking script (either inside add_run or before). 

WDYT ?

HenriquePreto:
> Yes! Believe the trickier part tho is the MachineGroup.register(). I cannot find a way to silence the logs without inspecting the stack.
> 
> The solution could simply be to not silence the register logs - since these need to be explicitly called in the benchmarking script (either inside add_run or before).
> 
> WDYT ?

I think it's fine to leave those logs as they are, since there's actually no clean way to silence the logs from `MachineGroup.register` without a hack. Even using something like stack inspection wouldn't help in this case:

```python
machine_group = inductiva.resources.MachineGroup(...) # <-- logs!
benchmark = inductiva.benchmarks.Benchmark(...).add_run(machine_group, ...)
```

Pbarbosa92:
I agree. I'm not a big fan of the hack we have to do to hide the logs.

But it should be possible to go into each method and add a verbose argument (this can be a bit of work because we would hve to add the argument and pass it around to each call made within a method).

inductiva/benchmarks/benchmark.py
```diff
@@ -57,12 +57,14 @@ def __init__(self, name: str):
...
         self.on = None
         self.kwargs = {}
+        self.verbose = False
 
     def set_default(
         self,
         simulator: Optional[simulators.Simulator] = None,
         input_dir: Optional[str] = None,
         on: Optional[types.ComputationalResources] = None,
+        verbose: Optional[bool] = None,
```
HenriquePreto, line 67: I don't think this parameter is needed here -- it doesn't vary between runs. I'd suggest passing it to the constructor and setting it to `True` or `False` by default instead of using `Optional`.

HenriquePreto: So `verbose` logs the messages of machine groups and tasks and `not verbose` logs the messages of a benchmark?

inductiva/resources/machine_groups.py
```diff
@@ -364,7 +365,8 @@ def can_start_resource(self) -> bool:
 
         return is_cost_ok and is_vcpu_ok and is_instance_ok
 
-    def start(self, wait_for_quotas: bool = False):
+    @mute_logging()
+    def start(self, wait_for_quotas: bool = False, verbose: bool = True):  # pylint: disable=unused-argument
```
HenriquePreto, line 369: Why not just use an if statement? Would we end up with multiple if statements?

I think an if statement is simpler and more consistent with the rest of the codebase. But, if we end up with a large number of if statements, this approach might be a better choice.

We could introduce these logging decorators during a future refactoring and apply them consistently across the codebase.

inductiva/benchmarks/benchmark.py
catarinalegria: Yes! `not verbose` logs the "summarized" benchmark info. `verbose` logs everything "normally".

inductiva/benchmarks/benchmark.py
```diff
@@ -299,6 +307,8 @@ def select_distinct(info):
         for task in tasks:
             task_input_params = get_task_input_params(task)
             task_info = task.info
+            if not task_info.executer:
```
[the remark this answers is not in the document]
catarinalegria: yes it is, if the task is killed it does not have Executer

HenriquePreto (approved):

PR #2159, opened 2025-05-21
Improve cans usability

Pbarbosa92:
1- Cans needs the folder data to be present, so, we create the folder with `mkdir -p`
2- Cans (to run on gpu) needs to have n_vcpus equal to the number of GPUs. If the user does not provide n_vcpus and we detect the resource has a GPU we will pick the n_vcpus for them.

### DIFF
--- a/inductiva/resources/machine_groups.py
+++ b/inductiva/resources/machine_groups.py
@@ -829,9 +829,7 @@ class MPICluster(BaseMachineGroup):
         """
         Returns the number of GPUs available in the resource.
         """
-        if self._gpu_info is None:
-            return 0
-        return self._gpu_info.get("gpu_count", 0) * self.num_machines
+        return super().gpu_count() * self.num_machines
 
     @property
     def available_vcpus(self):


inductiva/resources/machine_groups.py
```diff
@@ -817,6 +825,14 @@ def n_vcpus(self):
             self._cpu_info["cpu_cores_logical"] * self.num_machines,
             self._cpu_info["cpu_cores_logical"])
 
+    def gpu_count(self) -> bool:
+        """
+        Returns the number of GPUs available in the resource.
+        """
+        if self._gpu_info is None:
+            return 0
+        return self._gpu_info.get("gpu_count", 0) * self.num_machines
```
HenriquePreto, line 834: We could extract the code from the function above and refactor this method (and the other as well):

```python
def gpu_count(self) -> bool:
    ...
    return gpu_count(self._gpu_info) * self.num_machines
```

inductiva/simulators/cans.py
```diff
@@ -72,13 +72,18 @@ def run(self,
                                 remote_assets=remote_assets,
                                 sim_config_filename=sim_config_filename)
 
+        #To run cans on gpu the n_vcpus must be equal to the number of gpus
```
HenriquePreto, line 75: If this is mandatory, shouldn't we not remove the `n_vcpus is None` check? Shouldn't we `assert` or enforce that restriction?

inductiva/simulators/cans.py
```diff
@@ -72,13 +72,18 @@ def run(self,
...
+            n_vcpus = on.gpu_count()
+
         mpi_kwargs = {}
         mpi_kwargs["use_hwthread_cpus"] = use_hwthread
         if n_vcpus is not None:
             mpi_kwargs["np"] = n_vcpus
 
         mpi_config = MPIConfig(version="4.1.6", **mpi_kwargs)
         commands = [
+            "mkdir -p data",
```
HenriquePreto, line 86: We could store this `"data"` in a static variable of the Cans simulator class.

HenriquePreto: This is just a suggestion, as we may need to access it later when downloading the outputs.

inductiva/resources/machine_groups.py
Pbarbosa92: I'm not i follow

HenriquePreto: Something like:

```python
def gpu_count(gpu_info) -> int:
    if gpu_info is None:
        return 0
    return gpu_info.get("gpu_count", 0)

...

    def gpu_count(self) -> bool:
        """
        Returns the number of GPUs available in the resource.
        """
        return gpu_count(self._gpu_info)

    ...

    def gpu_count(self) -> bool:
        """
        Returns the number of GPUs available in the resource.
        """
        return gpu_count(self._gpu_info) * self.num_machines
```

This avoids duplicated code.

PR #2164, opened 2025-05-22
Fix wrong behavior in Benchmark's progress bar.

catarinalegria:
closes [#1114](https://github.com/inductiva/tasks/issues/1114)

### DIFF
--- a/inductiva/benchmarks/benchmark.py
+++ b/inductiva/benchmarks/benchmark.py
@@ -185,16 +185,16 @@ class Benchmark(projects.Project):
                        bar_format="{l_bar}{bar}| {n_fmt}/{total_fmt} {unit}",
                        initial=len(completed_tasks),
                        unit='tasks') as pbar:
-            with logging_redirect_tqdm():
-                with ThreadPoolExecutor() as executor:
-                    future_to_task = {
-                        executor.submit(
-                            lambda t: t.wait(download_std_on_completion=False,
-                                             silent_mode=True), task):
-                            task for task in running_tasks
-                    }
+            with ThreadPoolExecutor() as executor:
+                future_to_task = {
+                    executor.submit(
+                        lambda t: t.wait(download_std_on_completion=False,
+                                         silent_mode=True), task):
+                        task for task in running_tasks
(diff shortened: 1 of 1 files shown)


inductiva/benchmarks/benchmark.py
```diff
@@ -174,33 +174,37 @@ def wait(self) -> Self:
         """
         tasks = self.get_tasks()
 
+        completed_tasks = [task for task in tasks if task.info.is_terminal]
+        running_tasks = [task for task in tasks if not task.info.is_terminal]
```
HenriquePreto, line 178: The `get_tasks` has a `status` parameter. You could use that to filter the running tasks and completed

inductiva/benchmarks/benchmark.py
```diff
@@ -174,33 +174,37 @@ def wait(self) -> Self:
...
+
         logging.info("Waiting for Benchmark \033[1m%s\033[0m to complete...\n",
                      self.name)
 
         with tqdm.tqdm(total=len(tasks),
-                       desc="Processing tasks",
-                       bar_format="{l_bar}{bar}| {n_fmt}/{total_fmt}") as pbar:
+                       desc="Running Benchmark",
+                       bar_format="{l_bar}{bar}| {n_fmt}/{total_fmt} {unit}",
+                       initial=len(completed_tasks),
```
HenriquePreto, line 186: Shouldn’t it be `initial=len(running_tasks)`?

catarinalegria: I don't think so, i believe the progress bar should be "filled" with the completed tasks

inductiva/benchmarks/benchmark.py
catarinalegria: I know! I think it would be trickier to use in this case tho, since the status parameter of `get_tasks` can only be a single status, and, for example, the running tasks are composed of multiple status.

WDYT?

HenriquePreto: Ah ok, I see -- we should allow passing a list of statuses.

PR #2179, opened 2025-05-27
Create simulations from csv file

Pbarbosa92:


docs-sources/how-it-works/recipes/start-simulations-from-csv.md
```diff
@@ -0,0 +1,54 @@
...
+
+        inductiva.TemplateManager.render_dir( \
+            source_dir="my_templates",
+            target_dir="my_rendered_files",
+            overwrite=True,
+            density0=density0,
+            particleRadius=particleRadius,
+            viscosity=viscosity)
+
+        task = splishsplash.run( \
```
hpenedones, line 47: probably we should also set the task_metadata with those params... no?

Pbarbosa92: Yes. Done

PR #2191, opened 2025-05-30
Set `project_type = benchmark` when creating a new Benchmark

catarinalegria:


rcvalerio (approved):

inductiva/projects/project.py
```diff
@@ -117,10 +126,13 @@ def _get_project(self, name: str):
...
         return None
 
-    def _create_project(self, name):
+    def _create_project(self, name, project_type):
         """Creates a project with the given name on the backend."""
         try:
-            response = self._api.create_project({"name": name})
+            args = {"name": name}
+            if project_type:
+                args["project_type"] = project_type
```
HenriquePreto, line 134: It seems that `_create_project` is never called with `project_type=None`. Should we always assign `args["project_type"] = project_type` in that case?

catarinalegria: yes, if the `project_type` is not passed in the request, it assumes the type is "project". However, i think removing the if stmt like you suggested is cleaner

HenriquePreto (approved):

PR #2213, opened 2025-06-04
Fix download path when setting output_dir or input_dir

rcvalerio:
closes https://github.com/inductiva/tasks/issues/1138

This PR removes the prefix `inductiva.get_output_dir()` and the suffix `inputs`/`outputs` when `input_dir`/`output_dir` is set in the methods `download_inputs()`/`download_outputs()`.

It also adds the ability to select a custom `output_dir` in `projects.download_outputs()`.

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -1045,13 +1045,9 @@ class Task:
 
         logging.debug("\nDownload URL: %s\n", download_url)
 
-        if dest_dir is None:
-            dir_path = files.resolve_output_path(self.id)
-        else:
-            dir_path = pathlib.Path(dest_dir)
-
-        if sub_dir is not None:
-            dir_path = dir_path.joinpath(sub_dir)
+        base_path = pathlib.Path(
+            dest_dir) if dest_dir else files.resolve_output_path(self.id)
+        dir_path = base_path.joinpath(sub_dir) if sub_dir else base_path
 
         if (dir_path.exists() and not self._contains_only_std_files(dir_path)):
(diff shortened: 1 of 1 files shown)


sarmento (approved):

inductiva/projects/project.py
```diff
@@ -245,17 +245,23 @@ def wait(self):
...
-        All the files will be stored inside
-        `inductiva_output/<project_name>/<task_id's>`.
+        Args:
+            output_dir (str, optional): The base directory where project outputs
+                                        will be downloaded.
         """
-        all_tasks = self.get_tasks()
 
+        all_tasks = self.get_tasks()
         for task in all_tasks:
```
catarinalegria, line 262:
```suggestion
        for task in self.get_tasks():
```

inductiva/tasks/task.py
```diff
@@ -1046,10 +1046,12 @@ def _download(
...
         if dest_dir is None:
-            dest_dir = self.id
-        dest_dir += f"/{sub_dir}/"
+            dir_path = files.resolve_output_path(self.id)
+        else:
+            dir_path = pathlib.Path(dest_dir)
 
-        dir_path = files.resolve_output_path(dest_dir)
+        if sub_dir is not None:
+            dir_path = dir_path.joinpath(sub_dir)
```
catarinalegria, line 1055: Here we could also simplify the logic:
```
base_path = pathlib.Path(dest_dir) if dest_dir else files.resolve_output_path(self.id)
dir_path = base_path.joinpath(sub_dir) if sub_dir else base_path
```

catarinalegria (approved):

PR #2218, opened 2025-06-05
Update OpenAPI client generator

luispcunha:
Reference: https://github.com/inductiva/tasks/issues/744

This updates the python client code to use the updated openapi generator.

Note that the pipeline is not passing because I split the PR in two, one with the automated changes (https://github.com/inductiva/inductiva/pull/2225) and another with the manual changes I did to the code that uses the generated client to make it easier to review.


bgregoinductiva (approved):

hpenedones (approved):
Nice!

PR #2231, opened 2025-06-12
Improve openfoam list of commands

Pbarbosa92:
This PR introduces two main changes:

1. **Custom MPI Configuration for `MPICluster`**
   Users can now specify a custom MPI configuration when creating an `MPICluster` machine.

   * If no configuration is provided, a default will be used: `mpi_version=4.1.6`, `use_hwthread_cpus=True`, and `np` set to the number of available cores on the cluster.
Example:
```python
cloud_machine = inductiva.resources.MPICluster(
    provider="GCP",
    machine_type="c3d-standard-90",
    num_machines=9,
    data_disk_gb=200,
    spot=True
)
mpi_config-> version="4.1.6", use_hwthread_cpus=True, np=810
```

```python
cloud_machine = inductiva.resources.MPICluster(
    provider="GCP",
    machine_type="c3d-standard-90",
    num_machines=9,
    mpi_version="1.2.3",
    np=5,
    data_disk_gb=200,
    spot=True
)
mpi_config-> version="1.2.3", use_hwthread_cpus=True, np=5
```

2. **Improved OpenFOAM Command Handling**
   Users can once again pass a list of commands to OpenFOAM, now as plain strings (as before).

   * We automatically check each command for the presence of the `-parallel` flag.
   * If detected, the command is converted into a `Command` object, using the MPI configuration defined for the machine.

This will simplify the process of creating the list of commands from (old way):
```python
# MPI configuration
mpiconfig = MPIConfig("4.1.6", np=n_procs, use_hwthread_cpus=True)

# Preprocessing and solver commands
commands = [
    "cp system/controlDict.noWrite system/controlDict",
    "cp system/fvSolution.fixedIter system/fvSolution",
    f"decomposePar -constant {file_handler}",
    "restore0Dir -processor",
    Command(f"renumberMesh -constant -overwrite -parallel {file_handler}", mpi_config=mpiconfig),
    Command(f"potentialFoam -initialiseUBCs -parallel {file_handler}", mpi_config=mpiconfig),
    Command(f"applyBoundaryLayer -ybl '0.0450244' -parallel {file_handler}", mpi_config=mpiconfig),
    Command(f"{solver_rans} -parallel {file_handler}", mpi_config=mpiconfig),
]
```

to
```python
# Preprocessing and solver commands
commands = [
    "cp system/controlDict.noWrite system/controlDict",
    "cp system/fvSolution.fixedIter system/fvSolution",
    f"decomposePar -constant {file_handler}",
    "restore0Dir -processor",
    f"renumberMesh -constant -overwrite -parallel {file_handler}",
    f"potentialFoam -initialiseUBCs -parallel {file_handler}",
    f"applyBoundaryLayer -ybl '0.0450244' -parallel {file_handler}",
    f"{solver_rans} -parallel {file_handler}",
]
```
We will convert the last 3 commands into commands with mpi configs automatically.

> Note: The old way should work exactly the same. We only convert commands that are strings. If you pass a command object in the list we will not touch it.

### DIFF
--- a/inductiva/resources/machine_groups.py
+++ b/inductiva/resources/machine_groups.py
@@ -50,6 +50,9 @@ class BaseMachineGroup(ABC):
                 kept alive. After auto_terminate_minutes minutes the machine
                 will be terminated. This time will start counting after calling
                 this method.
+        mpi_version: The version of MPI to be used on the machines.
+        np: The number of processes to use for MPI commands.
+        use_hwthread_cpus: Whether to use hyperthreading or not.
 
     :meta private:
     """
@@ -63,6 +66,10 @@ class BaseMachineGroup(ABC):
     auto_terminate_ts: Optional[datetime.datetime] = None
     auto_terminate_minutes: Optional[int] = None
     spot: bool = True
+
+    mpi_version: str = "4.1.6"
+    np: int = None
+    use_hwthread_cpus: bool = True
 
     create_time = None
     num_machines = 0
@@ -150,6 +157,29 @@ class BaseMachineGroup(ABC):
         if self._gpu_info is None:
             return 0
         return self._gpu_info.get("gpu_count", 0)
+
+    def set_mpi_config(self,
+                       mpi_version: str = "4.1.6",
+                       np: int = None,
+                       use_hwthread_cpus: bool = True):
+        """Set the MPI configuration for the cluster.
+        Args:
+            mpi_version: The version of MPI to be used on the machines.
+            np: The number of processes to use for MPI commands.
+            use_hwthread_cpus: Whether to use hyperthreading or not.
+        """
+        self.mpi_version = mpi_version
+        self.use_hwthread_cpus = use_hwthread_cpus
+        if np is not None:
+            self.np = self.available_vcpus
+        else:
+            self.np = np
+
+    def get_mpi_config(self):
+        """Get the MPI configuration for the cluster."""
+        return MPIConfig(self.mpi_version,
+                         np=self.np,
+                         use_hwthread_cpus=self.use_hwthread_cpus)
 
     @property
     def id(self):
@@ -626,6 +656,8 @@ class MachineGroup(BaseMachineGroup):
         self._register_machine_group(num_vms=self.num_machines,
                                      spot=self.spot,
                                      is_elastic=self._is_elastic)
+        if self.np is None:
+            self.np = self.available_vcpus
 
     def _validate_inputs(self):
         super()._validate_inputs()
@@ -725,6 +757,8 @@ class ElasticMachineGroup(BaseMachineGroup):
                                      is_elastic=self._is_elastic,
                                      num_vms=self._active_machines,
                                      spot=self.spot)
+        if self.np is None:
+            self.np = self.available_vcpus
 
     def _validate_inputs(self):
         super()._validate_inputs()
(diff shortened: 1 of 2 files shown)


inductiva/resources/machine_groups.py
```diff
@@ -63,6 +67,10 @@ class BaseMachineGroup(ABC):
     auto_terminate_minutes: Optional[int] = None
     spot: bool = True
 
+    mpi_version: str = "4.1.6"
+    np: int = None
+    use_hwthread_cpus: bool = True
```
HenriquePreto, line 72: Why are these new parameters part of the `BaseMachineGroup`? Shouldn't they be specific to `MPICluster`?

inductiva/resources/machine_groups.py
```diff
@@ -150,6 +158,29 @@ def gpu_count(self) -> bool:
             return 0
         return self._gpu_info.get("gpu_count", 0)
 
+    def set_mpi_config(self,
+                       mpi_version: str = "4.1.6",
+                       np: int = None,
```
HenriquePreto, line 163: Looks like `np` parameter is optional

inductiva/resources/machine_groups.py
```diff
@@ -150,6 +158,29 @@ def gpu_count(self) -> bool:
...
+                       use_hwthread_cpus: bool = True):
+        """Set the MPI configuration for the cluster.
+        Args:
+            mpi_version: The version of MPI to be used on the machines.
+            np: The number of processes to use for MPI commands.
+            use_hwthread_cpus: Whether to use hyperthreading or not.
+        """
+        self.mpi_version = mpi_version
+        self.use_hwthread_cpus = use_hwthread_cpus
+        if np is not None:
```
HenriquePreto, line 173: Shouldn't be `if np is None`?

HenriquePreto: Suggestion: `self.np = np or self.available_vcpus`

inductiva/resources/machine_groups.py
```diff
@@ -150,6 +158,29 @@ def gpu_count(self) -> bool:
...
+            self.np = self.available_vcpus
+        else:
+            self.np = np
+
+    def get_mpi_config(self):
+        """Get the MPI configuration for the cluster."""
+        return MPIConfig(self.mpi_version,
+                         np=self.np,
+                         use_hwthread_cpus=self.use_hwthread_cpus)
```
HenriquePreto, line 182: Again, shouldn't this be specific to the `MPICluster` class?

inductiva/resources/machine_groups.py
Pbarbosa92: I had it like that. But you can also define commands with mpi configs in normal machines.

HenriquePreto: Oh, are you saying, for example, that I can run a simulator on a `MachineGroup` that doesn't support MPI / it's not a `MPICluster`? What's the point of that? Sorry, but I'm not sure I understand…

inductiva/resources/machine_groups.py
```diff
@@ -625,6 +653,7 @@ def __post_init__(self):
         self._register_machine_group(num_vms=self.num_machines,
                                      spot=self.spot,
                                      is_elastic=self._is_elastic)
+        self.np = self.np or self.available_vcpus
```
HenriquePreto, line 656: Why is this repeated here if `super().__post_init__()` is already being called?

HenriquePreto: The same applies for the other sub classes

inductiva/resources/machine_groups.py
```diff
@@ -798,6 +828,9 @@ class MPICluster(BaseMachineGroup):
     """
     # Constructor arguments
     num_machines: int = 2
+    mpi_version: str = "4.1.6"
+    np: int = None
+    use_hwthread_cpus: bool = True
```
HenriquePreto, line 833: Why is this repeated here if these attributes are already in the `BaseMachineGroup` base class?

inductiva/resources/machine_groups.py
HenriquePreto: Oh, I see -- it looks like we have some duplicated code here, but we can refactor it later.

PR #2237, opened 2025-06-12
Openapi client update fixes

luispcunha:
Reference: https://github.com/inductiva/tasks/issues/744

Fix a few places where the usage of the openapi client was wrong.


HenriquePreto:
probably we should run the e2e tests using the new generated client...

PR #2274, opened 2025-06-23
Fix openfoam command creation

Pbarbosa92:
Last release we added the feature for openfoam to build the mpi commands automatically if `-parallel` is present in the command. But we don't detect if the user already passed `mpirun` in the command.

We now only build the mpi command if the command is a string and mpirun is not present in the command.


Pbarbosa92:
> Is this behaviour properly documented in the simulator guide or something like that?

I've added something in the FAQ. Since we don't have (yet) a tutorial with a list of commands FAQ is the best place for now. But we will add a tutorial with mpi clusters that can go into more detail.

PR #2279, opened 2025-06-23
Fix creation of image_uri in simulator.run

luispcunha:
The image_uri was being updated more than once if several tasks were run with the same simulator object. This fixes that by using a local variable for the final image_uri instead of updating the image_uri property of the object.


Pbarbosa92:
> > Oh, good catch! This one we might want to deploy as hot fix to PROD, right?
> 
> Yes, I think so

Yes. There is also another hot fix in dev that would be nice to have in prod.

PR #2291, opened 2025-06-27
Add "Scale Up and Organize" doc

HenriquePreto:
This is an end-to-end tutorial that shows how to submit multiple simulations in parallel, organize them using projects and metadata, monitor their progress, and download the outputs cleanly and in an organized way.

I believe this is a common use case, and it could be valuable to include it under "Resources > Scale-up > Parallel Simulations".

@eunmoliveira @ClaudiaRSilva @Pbarbosa92 wdyt? Also, could you point me to the right place to store this document?


eunmoliveira:
> This is an end-to-end tutorial that shows how to submit multiple simulations in parallel, organize them using projects and metadata, monitor their progress, and download the outputs cleanly and in an organized way.
> 
> I believe this is a common use case, and it could be valuable to include it under "Resources > Scale-up > Parallel Simulations".
> 
> @eunmoliveira @ClaudiaRSilva @Pbarbosa92 wdyt? Also, could you point me to the right place to store this document?

I agree, it's a recipe so it should go to that part of the documentation. To store this document, go to our repo > docs-sources > scale-up (it's among the simulators)

eunmoliveira:
> > > This is an end-to-end tutorial that shows how to submit multiple simulations in parallel, organize them using projects and metadata, monitor their progress, and download the outputs cleanly and in an organized way.
> > > I believe this is a common use case, and it could be valuable to include it under "Resources > Scale-up > Parallel Simulations".
> > > @eunmoliveira @ClaudiaRSilva @Pbarbosa92 wdyt? Also, could you point me to the right place to store this document?
> > 
> > 
> > I agree, it's a recipe so it should go to that part of the documentation. To store this document, go to our repo > docs-sources > scale-up (it's among the simulators)
> 
> Thanks! Done.

In terms of review, I leave that to @ClaudiaRSilva, as she's more in touch with that part of the documentation now

HenriquePreto:
> The timing for this is perfect, because a few users who I've been in contact with are struggling with organizing their data. Yet, I'm also concerned about creating repetitive documentation. @HenriquePreto are you aware of these 2 pre existing docs: https://inductiva.ai/guides/scale-up/parallel-simulations/run-parallel_simulations https://inductiva.ai/guides/how-it-works/recipes/download-file-from-project
> 
> It's ok if this new one replaces 1 or both previously existing docs, as long as it adds value. I'm only trying to avoid that we disperse the same information throughout many docs, as it will become harder for us to manage and update, and confusing for the user.

Thanks, Claudia! Yes, I'm aware of those docs. I wouldn't say this one is repetitive, though — it's more of a complement to the others (or even an introduction to submitting multiple simulations to run in parallel on the cloud).

For example, this example doesn't use the `TemplateManager` class, which is more for advanced use cases like parameter sweeping. Here, it's just showing a simpler way to run multiple simulations in parallel on the cloud without that extra layer.

I will go by this rule: if a user's use case doesn't clearly match anything we've documented, then we probably need more docs.

Why I think this doc adds value:
- Shows how to run multiple parallel simulations without using the template manager
- Monitors using the Console (not the terminal)
- Demonstrates grouping tasks into a project to easily download them later (on the tutorial you sent, we don't show how a user could download the multiple tasks in a systematic and organized way -- and as the script is written, it doesn't actually support that flow)
- Uses task metadata to help distinguish between runs

All of these are not shown here: https://inductiva.ai/guides/scale-up/parallel-simulations/run-parallel_simulations

Wdyt?

catarinalegria (approved):
Really cool tutorial !! Everything seems clear and very informative :)) 

I just have a small note (that Cláudia already commented): maybe the ["Run Simulation in Parallel"](run-parallel_simulations) tutorial now seems a bit redudant (specially since both tutorials will be on the same Scale Up "section")-- all it's key information and takeaways are compreenhded in this new tutorial imo.

i know it does not use the `TemplateManager` but it would be cool to maybe update the script to use it and even add the water_level of each run to the metadata!

Its just an idea, if there is a great effort of doing this change now we can discuss the possibility of merging them later!

HenriquePreto:
> Really cool tutorial !! Everything seems clear and very informative :))
> 
> I just have a small note (that Cláudia already commented): maybe the ["Run Simulation in Parallel"](run-parallel_simulations) tutorial now seems a bit redudant -- all it's key information and takeaways are compreenhded in this new tutorial imo.
> 
> i know it does not use the `TemplateManager` but it would be cool to maybe update the script to use it and even add the water_level of each run to the metadata!
> 
> Its just an idea, if there is a great effort of doing this change now we can discuss the possibility of merging them later!

I think the use cases are different.

This one is simpler: you are just running multiple simulations in parallel without changing parameters. It's not obvious to users that every tutorial on parallel runs includes the `TemplateManager` class, which can make it seem like it is required, when it is not.

I think it is totally fine (for now) to have 2 or 3 tutorials that overlap in some ways but differ in others. I think that difference actually helps users get a better feel for the API.

ClaudiaRSilva:
> > Really cool tutorial !! Everything seems clear and very informative :))
> > I just have a small note (that Cláudia already commented): maybe the ["Run Simulation in Parallel"](run-parallel_simulations) tutorial now seems a bit redudant -- all it's key information and takeaways are compreenhded in this new tutorial imo.
> > i know it does not use the `TemplateManager` but it would be cool to maybe update the script to use it and even add the water_level of each run to the metadata!
> > Its just an idea, if there is a great effort of doing this change now we can discuss the possibility of merging them later!
> 
> I think the use cases are different.
> 
> This one is simpler: you are just running multiple simulations in parallel without changing parameters. It's not obvious to users that every tutorial on parallel runs includes the `TemplateManager` class, which can make it seem like it is required, when it is not.
> 
> I think it is totally fine (for now) to have 2 or 3 tutorials that overlap in some ways but differ in others. I think that difference actually helps users get a better feel of the API.


Regarding with having 2 or 3 tutorials that overlap, it's ok if it's temporary, but we must avoid it - the same way that we needed to make all simualtors visible in the same menu, the docs pages will also need to be as visible as possible at first glance (I'm working on it). Therefore, docs that are redundant will take much needed space!

PR #2292, opened 2025-06-29
Modify visualization requirements for DELILAH case

Franreno:
Specify the visualization requirements changes for the DELILAH case.


sarmento (approved):
Thank you for this. I will test this later!

PR #2295, opened 2025-06-30
Xbeach - rename sim config file name to params.txt

Pbarbosa92:
It turns out, xbeach does not support passing the sim config file name as input. It always looks for `params.txt`


HenriquePreto:
If a user passes sim config file name different from params.txt, it will appear in the outputs -- right?

PR #2298, opened 2025-06-30
Add "Setting a TTL (Time-to-Live) on Your Simulations" tutorial

HenriquePreto:
Closes [953](https://github.com/inductiva/tasks/issues/953).

### DIFF
--- a/docs-sources/how-it-works/recipes/set-task-ttl/set-task-ttl.md
+++ b/docs-sources/how-it-works/recipes/set-task-ttl/set-task-ttl.md
@@ -29,8 +29,9 @@ task = simulator.run(
 )
 ```
 
-
-You'll see logs like these in your terminal:
+If the simulation exceeds the maximum duration (in this case, 1 minute), you'll see logs 
+like these in your terminal indicating that it was automatically terminated because the 
+TTL was reached:
 
 ```bash
 ■ Your task exceeded its configured time-to-live (TTL) and was automatically stopped.


docs-sources/how-it-works/recipes/set-task-ttl/set-task-ttl.md
```diff
@@ -0,0 +1,37 @@
+# Setting a TTL (Time-to-Live) on Your Simulations
+
+To set a TTL (Time-to-Live) for your simulation, you need to use the `time_to_live` 
+parameter when submitting the task. This parameter defines the maximum duration the 
+task is allowed to run, expressed as a string with a time unit suffix. Supported formats 
+include minutes (e.g., `"10m"`) or hours (e.g., `"2h"`). Once the specified time has
```
hpenedones, line 6: Does "3h30m" work? or "3h30"?

hpenedones (commented):
This is great, thanks @HenriquePreto !

docs-sources/how-it-works/recipes/set-task-ttl/set-task-ttl.md
HenriquePreto: No -- but I'll make another PR to make it work.

I'll add another dependency to the Python Client called `pytimeparse` that does this automatically.

HenriquePreto: wdyt?

HenriquePreto: Done: https://github.com/inductiva/inductiva/pull/2302

Pbarbosa92 (changes requested):
I think we should follow the template used in https://inductiva.ai/guides/how-it-works/recipes/download-file-from-project

We should start by `The Challenge` where we explain the issue. In this case, "I ran my simulation but due to some miss-configuration the simulation ran for longer than what i expected. I ended up wasting credits because i know this simulation should take at a maximum 5 hours but it ended up running for 10 h before i noticed the mistake."

after this you can present the solution where you explain what ttl does and how to use it.

HenriquePreto:
> I think we should follow the template used in https://inductiva.ai/guides/how-it-works/recipes/download-file-from-project
> 
> We should start by `The Challenge` where we explain the issue. In this case, "I ran my simulation but due to some miss-configuration the simulation ran for longer than what i expected. I ended up wasting credits because i know this simulation should take at a maximum 5 hours but it ended up running for 10 h before i noticed the mistake."
> 
> after this you can present the solution where you explain what ttl does and how to use it.

I'm going to be a bit less specific about the challenge, because I think there are multiple reasons why a user may want to use this parameter.

PR #2302, opened 2025-07-01
Add `pytimeparse` for `time_to_live` parsing

HenriquePreto:
This PR adds `pytimeparse` as a dependency to convert the `time_to_live` parameter from a string to seconds. It also updates the simulator's docstrings.

### DIFF
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -350,10 +350,10 @@ def submit_task(simulator,
                 assigned. If None, the task will be assigned to
                 the default project.
         time_to_live: Maximum allowed runtime for the task, specified as a
-            string duration. Supports all formats accepted by the
-            `pytimeparse2` library, such as "10m", "2 hours", "1h30m", or
-            "90s". The task will be automatically terminated if it exceeds
-            this duration after starting.
+            string duration. Supports common time duration formats such as
+            "10m", "2 hours", "1h30m", or "90s". The task will be
+            automatically terminated if it exceeds this duration after
+            starting.
     Return:
         Returns the task id.
     """
--- a/inductiva/api/methods.py
+++ b/inductiva/api/methods.py
@@ -349,11 +349,11 @@ def submit_task(simulator,
         project: Name of the project to which the task will be
                 assigned. If None, the task will be assigned to
                 the default project.
-        time_to_live: Maximum duration the task is allowed to run, 
-            specified as a string with a time unit suffix. Supported formats
-            include minutes ("10m") or hours ("2h"). The task will be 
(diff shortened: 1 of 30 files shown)


rcvalerio (commented):
Are you using the correct version of `pytimeparse`?

https://pypi.org/project/pytimeparse/
https://pypi.org/project/pytimeparse2/

HenriquePreto:
> > > Are you using the correct version of `pytimeparse`?
> > > https://pypi.org/project/pytimeparse/ https://pypi.org/project/pytimeparse2/
> > 
> > 
> > It looks like `pytimeparse2` supports more formats but `pytimeparse` is simpler and more lightweight -- it covers the basics.
> 
> But, I can switch to `pytimeparse2` to use a more recent and maintainable version. Thanks!

Much better with `pytimeparse2` -- just one line of code now:

https://github.com/inductiva/inductiva/blob/449a71eb461d2c6c554dfbc6c020b486ac759147/inductiva/api/methods.py#L367

inductiva/api/methods.py
```diff
@@ -14,7 +14,7 @@
 import ssl
 from contextlib import contextmanager
 from typing import List, Optional
-from pytimeparse.timeparse import timeparse
+from pytimeparse2 import parse
```
hpenedones, line 17: small note: in principle we should import the module, not a particular method, according to the Google Python guidelines that we try to follow:

https://google.github.io/styleguide/pyguide.html#22-imports

So below you will do pytimeparse2.parse(...)

hpenedones: Thanks!

inductiva/api/methods.py
```diff
@@ -351,7 +351,7 @@ def submit_task(simulator,
                 the default project.
         time_to_live: Maximum allowed runtime for the task, specified as a
             string duration. Supports all formats accepted by the
-            `pytimeparse` library, such as "10m", "2 hours", "1h30m", or
+            `pytimeparse2` library, such as "10m", "2 hours", "1h30m", or
```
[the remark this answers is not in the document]
hpenedones: Right... I don't have a super strong opinion -- it is true that it might look a bit to "specific" to refer to a particular library that we are relying on here in the docstring. Ideally, we would just give some examples, and then in some other location provide more detail and a link to the the library... For example in the more detailed docs you wrote here: https://github.com/inductiva/inductiva/pull/2298

bgregoinductiva:
> Is this a breaking change because of the new dependency? @HenriquePreto @bgregoinductiva

I guess not. This is a new optional parameter.

PR #2303, opened 2025-07-01
Add ability to configure OpenMP for fds simulator

luispcunha:
Adding the `n_omp_threads` and `n_mpi_processes` to the FDS simulator.
These two parameters, as well as `n_vcpus` are optional.
If `n_vcpus` is not set, we default to the number of available vcpus on the machine. If one of the other two is not set, then we make it so that `n_omp_threads * n_mpi_processes = n_vcpus`. If none of the others are set, we make it so that `n_mpi_processes = 1` and `n_omp_threads = n_vcpus` (this is because the number of mpi processes can't exceed the number of meshes, so having a default of n_mpi_processes different than one would cause most cases to fail).

If not set, the default is to set `n_vcpus` to the number of available `n_vcpus` on the VM, and set `n_omp_threads``n_omp_threads`

### DIFF
--- a/inductiva/tests/simulators/test_simulator_with_resources.py
+++ b/inductiva/tests/simulators/test_simulator_with_resources.py
@@ -269,6 +269,7 @@ def test_resubmit_on_preemption__is_correctly_handled(resubmit_on_preemption):
     mock_mg.has_gpu.return_value = True
     mock_mg.gpu_count.return_value = 1
     mock_mg.available_vcpus = 16
+    mock_mg.get_available_mpi_slots.return_value = 8
 
     for sim_name, simcls in sim_classes:
 


inductiva/simulators/fds.py
```diff
@@ -70,17 +71,36 @@ def run(self,
                                 remote_assets=remote_assets,
                                 sim_config_filename=sim_config_filename)
 
+        available_vcpus = on.available_vcpus
+
+        if n_vcpus is None and omp_num_threads is None:
```
sarmento, line 76: I think the issue here is that we need to change the nomeclature.
What we need to introduce is a new variable / property n_mpi_processes.
For most simulators we have so far:

n_mpi_processes = n_vcpus = n_available_vcpus
(this is what happens for most of existing cases, and that's the behaviour we need to keep).

For FDS

n_mpi_processes * omp_num_threads = n_vcpus =  available_vcpus

I think the issue here is that we have been using n_vcpus for setting MPI processes but they are two different things. So far, this has worked, but not anymore....

I think we many need to refactor this accross the board, but we can do in one by one. Let's get FDS right first, since it's the most generic one.

Pbarbosa92: If n_mpi_processes=1 we need to call the comand without the mpirun i think.

Using mpirun can slow things down.

lets day the command is fds.exe

if mpi = 1 we just call fds.exe with the omp value defined

Only if mpi>1 we do mpirun fds.exe

inductiva/simulators/fds.py
```diff
@@ -70,17 +73,63 @@ def run(self,
...
-        mpi_kwargs = {}
-        mpi_kwargs["use_hwthread_cpus"] = use_hwthread
-        if n_vcpus is not None:
+        available_vcpus = on.available_vcpus
+
+        if all(arg is not None
+               for arg in [n_vcpus, n_omp_threads, n_mpi_processes]):
+            raise ValueError(
+                "Parameters n_vcpus, omp_n_threads and n_mpi_processes "
+                "can't be set simultaneously.")
```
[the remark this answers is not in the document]
Pbarbosa92: I think we should. We just need to check if n_vcpus<= omp_n_threads*n_mpi_processes

WDYT @sarmento ?

PR #2305, opened 2025-07-02
generate-dataset-recipe

eunmoliveira:
Removed from "Tutorials".
Created a new recipe under Resources.


eunmoliveira:
> We removed a very in depth tutorial to add a very vague recipe. I feel like this recipe is missing actual code.
> 
> We might want to create a very small example with 2 or 3 runs. I don't think that having just this bullet points is enough to help the user.
> 
> And we are also mixing generating datasets with benchmarks, which can be confusing.

Paulo, the proper tutorial of how to create a dataset still exists under the SplishSplash guide. The benchmarks part, I agree, but that was already stated in that doc before - if it doesn't make sense anymore lmk

eunmoliveira:
> > > We removed a very in depth tutorial to add a very vague recipe. I feel like this recipe is missing actual code.
> > > We might want to create a very small example with 2 or 3 runs. I don't think that having just this bullet points is enough to help the user.
> > > And we are also mixing generating datasets with benchmarks, which can be confusing.
> > 
> > 
> > Paulo, the proper tutorial of how to create a dataset still exists under the SplishSplash guide (https://inductiva.ai/guides/splishsplash/synthetic-data-for-piml/index). The benchmarks part, I agree, but that was already stated in that doc before - if it doesn't make sense anymore lmk
> > I can link some of the tutorials we have on datasets - and I think that makes sense - but how would you include 2 or 3 runs without explaining in detail like we do in the tutorials?
> 
> Well you would need to explain in some details. How to create the template, and starting 2 or 3 simulations. But we could pick a much simpler simulations, like swan.

Fine by me, you can include that. @sarmento @ClaudiaRSilva if you have any thoughts on this, let us know

eunmoliveira:
> > > We removed a very in depth tutorial to add a very vague recipe. I feel like this recipe is missing actual code.
> > > We might want to create a very small example with 2 or 3 runs. I don't think that having just this bullet points is enough to help the user.
> > > And we are also mixing generating datasets with benchmarks, which can be confusing.
> > 
> > 
> > Paulo, the proper tutorial of how to create a dataset still exists under the SplishSplash guide (https://inductiva.ai/guides/splishsplash/synthetic-data-for-piml/index). The benchmarks part, I agree, but that was already stated in that doc before - if it doesn't make sense anymore lmk
> > I can link some of the tutorials we have on datasets - and I think that makes sense - but how would you include 2 or 3 runs without explaining in detail like we do in the tutorials?
> 
> Well you would need to explain in some details. How to create the template, and starting 2 or 3 simulations. But we could pick a much simpler simulations, like swan.

Wouldn't that be repetitive, since we already have documentation on how to create the template and it's already linked here? @sarmento @ClaudiaRSilva if you have any thoughts on this, let us know.

ClaudiaRSilva:
A few comments and questions:
- Why did we feel the need to replace the previous in-depth content by this simpler one? What was wrong with the old one that made us go through the extra effort of creating a new one?
- I'm tasked with migrating the old Docs and Tutorials into the website. Does this mean that the old  'generating synthetic data' should not be migrated (being replaced by this new one)?
- If the proper tutorial of how to create a dataset still exists under the SplishSplash guide, why isn't it linked in this new doc?

sarmento:
PLease think of this document as a hub: people will search for varous things on the internet and then they will get there and we point them to various other more specific dociments we have. This should be a traffic magnet. It could even point to HuggingFace, etc. Pls think about this strategically.

Pbarbosa92 (approved):

PR #2331, opened 2025-07-04
Improve task.wait method to support waiting for specific task statuses

rcvalerio:
related to https://github.com/inductiva/tasks/issues/1162

Adds ability to wait for status:
`task.wait(wait_for_status="computation-started")`

### DIFF
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -845,7 +845,6 @@ class Task:
 
             time.sleep(polling_period)
 
-
     def _validate_task_computation_started(self) -> Tuple[bool, Optional[str]]:
         info = self.get_info()
         if info.is_terminal:
--- a/inductiva/tasks/task.py
+++ b/inductiva/tasks/task.py
@@ -665,14 +665,11 @@ class Task:
             if status == models.TaskStatusCode.FAILED:
                 self._print_failed_message(out_dir)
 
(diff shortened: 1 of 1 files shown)


Pbarbosa92:
Hmm this changes what the method does. With this argument the method description is no longer correct. And we are adding even a new functionality to an already complex method. Maybe we can create a new method?

maybe task.wait_for_status(status=abc)

rcvalerio:
> > Hmm this changes what the method does. With this argument the method description is no longer correct. And we are adding even a new functionality to an already complex method. Maybe we can create a new method?
> > maybe task.wait_for_status(status=abc)
> 
> What happens if a user passes a `status` diff from complete and a `download_on_complete=True`?
> 
> I feel like these parameters can be part of the same method (i.e., `task.wait`) but `download_on_complete` would download the files once the task reached to the passed status.

I think `download_on_complete=True` does not make sense if the status is not a terminal status (since there is no output to download) and it should be ignored (I can add that information to the docstring).

> This way, you are just making this function more generic, but the default `task.wait` keeps working as it was (i.e., `status=complete`)

Exactly, that's the goal.

rcvalerio:
I created a new method task.wait_for_status() to avoid changes in the task.wait()

Pbarbosa92 (approved):

PR #2336, opened 2025-07-04
Add wait parameter to task.tail_files()

rcvalerio:


Pbarbosa92:
what does the wait do? isn't that the same as follow?

PR #2354, opened 2025-07-08
Delft3d - review

eunmoliveira:


Pbarbosa92:
Sarmento asked to add some info about this case on the FAQ. Could you do it on this PR please ?

Something short, we can link to the tutorial.

Pbarbosa92:
> > Sarmento asked to add some info about this case on the FAQ. Could you do it on this PR please ?
> > Something short, we can link to the tutorial.
> 
> of course, lmk what it is

Something like this
Q: How can i run a coupled Delft3d simulation?

Then we cant tell them they need to run two commands at the same time with
command 1 &
command 2

just like we did in the tutorial for wave-flow (add link)

Something very simple like this. The tutorial would do the work

PR #2356, opened 2025-07-08
Added image to template tutorial

Pbarbosa92:


eunmoliveira (approved):

Pbarbosa92:
> I can't amplify the image, but from what I can see, I would add legends to each block of the diagram

With what?

PR #2368, opened 2025-07-09
 Reorganize Parallel Simulation Tutorials

HenriquePreto:
After [this discussion](https://github.com/inductiva/inductiva/pull/2291#issuecomment-3052221347), I realized the tutorial titles weren't clearly distinguishing their purpose. So I updated them to better reflect their focus:

- Renamed the original "Run Simulations in Parallel" to "Run Simulations in Parallel **with Templating**"
- Made the simpler version I added the new "Run Simulations in Parallel"

I also adjusted the order of the tutorials under the "Scale-Up" section to reflect the natural learning progression:
- Run Simulations in Parallel (introductory, simpler case, **no `TemplateManager`**)
- Templating Manager (explains the tool itself)
- Run Simulations in Parallel with Templating (applies both concepts together)

This way, users learn to run things in parallel first, then understand the templating mechanism, and finally see how both can be combined in a more advanced workflow.

Wdyt?


ClaudiaRSilva (approved):
Love the way you show the evolving path!
Could we make the titles smaller, though? It will be easier to read if it's not split in 2 lines.
E.g.
Parallel Simulations
Templating
Parallel Simulations with Templating

PR #2380, opened 2025-07-10
Improve Tutorial: Reuse Files Across Multiple Simulations

rcvalerio:
This PR improves the **Reuse Task Outputs in Simulations** section by clearly explaining how adding `remote_assets` copies previous task outputs into a subfolder named after the task ID within the new workspace.

It adds practical examples showing how to reuse files both in commands and in configuration files using the Template Manager.


tutorials/how_to/reuse-files.md
```diff
@@ -151,16 +167,59 @@ task = gromacs.run(
...
-    "gmx solvate -cs tip4p -box 2.3 -o conf.gro -p <task_id>/topol.top",
+    f"gmx solvate -cs tip4p -box 2.3 -o conf.gro -p {previous_task.id}/topol.top",
     ...
 ]
 ```
+Make sure to use an f-string (`f"..."`) so that `previous_task.id` gets correctly inserted into the command string.
+
+
+### How to Reference Task Outputs in a Config File
+In some simulations, you may need to inject the path of a file from a previous task into a config file, rather than referencing it directly in a command. For example, in OpenFAST, to reuse a TurbSim-generated wind field, you must point to its .bts file in the `InflowWind.dat` file.
```
sarmento, line 193: I think this is good for now, but assume that we will be doing more of these for other simulators, and then we can actually place these more specific aspects of the recipe in a dedicated "Recipes" section for each simulator. This will, of course, take its time so, for now, it's good to keep this specific example here.

sarmento (approved):

PR #2453, opened 2025-07-22
Octopus Review

eunmoliveira:
Did small improvements and added the e-h scattering link to the index. Overall looks great :)

On the e-h scatteting tutorial, I  think it would be a nice storyline to show how one would benefit from running the 3d simulation of the atom on the cloud, since the 1d simulation is small and you can't really see much performance improvements through scaling. It would also give a nice visualization + they don't show you how to setup a 3d simulation in their tutorial


sarmento (approved):
This looks great!
Thank you!

PR #2483, opened 2025-07-29
Update CLI & Pyhton Client Intro Tables to still use old links

catarinalegria:


Pbarbosa92:
Internal links don't work from one sphinx project to the other :D

PR #2543, opened 2025-08-08
Make website documentation menus fully configurable

arasahcor:
Related to https://github.com/inductiva/website-new/pull/438

Added `menu.json` to `docs-sources` to centralize control of the website's Resources and Simulators menus.
This gives content editors full control over what appears in the menus, no need to modify any frontend code.


hpenedones:
This is great! I guess it needs the corresponding changes on the website repo, but then the static builder should be able to optimise also the documentation pages, right @luispcunha ?

PR #2546, opened 2025-08-12
Update MachineTypeInfo class

bgregoinductiva:


inductiva/resources/utils.py
```diff
@@ -32,6 +32,7 @@ class MachineTypeInfo:
     zone: str
     num_gpus: Optional[int]
     gpu_name: Optional[str]
+    spot_price: Optional[float]
```
ayomidetobi, line 35: I think the main issues if from the api

Pbarbosa92: Spot_price is always None. At least for now.

PR #2562, opened 2025-08-21
Sanitize file paths for download

rcvalerio:
closes https://github.com/inductiva/tasks/issues/1153

This PR fixes two bugs on windows:
- Downloading files with invalid chars on filename;
- Extracting zips with files with invalid chars;


rcvalerio:
> Thank you Rodrigo for this! Small ask: do you think you can put that line of code behind a tiny function that we can write a couple of unit tests for? This will actually "document" what it does with human readable examples of pairs of strings of paths, before and after the unquote operation. What do you think?

Yes, of course

HenriquePreto (approved):

PR #2563, opened 2025-08-21
Fix windows line ending

rcvalerio:
closes https://github.com/inductiva/tasks/issues/1195

### DIFF
--- a/inductiva/utils/data.py
+++ b/inductiva/utils/data.py
@@ -21,6 +21,41 @@ ARTIFACTS_DIRNAME = "artifacts"
 INPUT_DIRNAME = "sim_dir"
 
 
+def _normalize_file(path: str) -> None:
+    """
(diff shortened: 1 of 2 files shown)


bgregoinductiva (approved):

inductiva/utils/data.py
```diff
@@ -21,6 +22,41 @@
...
+    Args:
+        path (str): Path to the file to normalize.
+    """
+    _, ext = os.path.splitext(path)
+    if ext.lower() not in {".txt", ".sh"}:
+        return
+
+    with open(path, "rb") as f:
+        first_line = f.readline()
+        print(f"Checking Normalizing file {path}...")
```
HenriquePreto, line 42: I'm not sure we should print this to the user... Normalizing a file doesn't look like something the user would care

HenriquePreto (approved):

PR #2580, opened 2025-09-02
Add the ability for users to pass unswan as a command (only for newer versions of swan)

Pbarbosa92:
This PR adds the ability for users to use unswan, swan with unstructured grids.

unswan is only available on the lastest version of swan (41.45A+)

### DIFF
--- a/inductiva/simulators/swan.py
+++ b/inductiva/simulators/swan.py
@@ -92,6 +92,11 @@ class SWAN(simulators.Simulator):
                     ]
         """
 
+        if self.version and command == "unswan" and self.version in {
+                "41.31", "41.45"
+        }:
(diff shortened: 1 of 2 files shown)


inductiva/simulators/swan.py
```diff
@@ -92,6 +92,11 @@ def run(
                     ]
         """
 
+        if self.version and command == "unswan" and self.version in {
+                "41.31", "41.45"
+        }:
+            raise ValueError("Unswan version 41.31 or 41.45 is not supported.")
+
         if command not in ("swanrun", "swan.exe"):
```
bgregoinductiva, line 100: You need to add `unswan` to this list.

bgregoinductiva (approved):

PR #2664, opened 2025-09-26
Launch task-runner on BYOC GCP [alpha version]

rcvalerio:
### Python Client
```python
import inductiva

mg = inductiva.resources.MachineGroup( \
    provider="GCP",
    machine_type="c2d-highcpu-8",
    spot=True,
    mg_name="byoc-gcp-machine",
    byoc=True, # Bring Your Own Cloud
)
```
Once launched, the machine group can be used like any other.

### Command Line
You can view all available flags with:
```
inductiva task-runner launch -h
```

To launch a machine on your own GCP account:
```
inductiva task-runner launch byoc-gcp-machine \
    --provider gcp \
    --machine-type c2d-highcpu-8 \
    --spot
```
After launching, you can use the machine group in python:

```python
mg = inductiva.resources.machine_groups.get_by_name('byoc-gcp-machine')
```

### Limitations of Alpha Version
- **Cost calculation**: Currently always reports 0.
- **Storage**: Only inductiva storage is supported.
- Is registered in the backend as a **LOCAL** machine.

### DIFF
--- a/inductiva/resources/machine_groups.py
+++ b/inductiva/resources/machine_groups.py
@@ -715,7 +715,7 @@ class MachineGroup(BaseMachineGroup):
                 "For managed resources, names are automatically generated.")
 
         if self.mg_name:
-            if not re.match(r'^[0-9a-zA-Z-]+$', self.mg_name):
+            if not re.match(r"^[0-9a-zA-Z-]+$", self.mg_name):
                 raise ValueError(
                     "`mg_name` must contain only letters, numbers, and hyphens."
                 )
@@ -845,10 +845,9 @@ class MachineGroup(BaseMachineGroup):
         """Terminate GCP VMs using client-side management."""
         logging.info("Terminating %s (client-side GCP)...", repr(self))
 
-        success, error_message = byoc_gcp.delete_gcp_vm(self._vm_name,
-                                                        self.zone, verbose)
+        try:
+            byoc_gcp.delete_gcp_vm(self._vm_name, self.zone, verbose)
 
-        if success:
             # Also notify backend that machine group is terminated
             try:
                 self._api.delete_vm_group(machine_group_id=self.id)
@@ -863,9 +862,9 @@ class MachineGroup(BaseMachineGroup):
 
             logging.info("%s terminated.", self)
             return True
-        else:
+        except ValueError as e:
             print("Failed to terminate GCP VM:")
-            print(error_message)
+            print(e)
             return False
 
     @logs.mute_logging()


inductiva/resources/byoc_gcp.py
```diff
@@ -0,0 +1,258 @@
...
+    match = re.search(r"-(\d+)$", machine_type)
+    if match:
+        return int(match.group(1))
+    raise ValueError(
+        f"Could not extract vCPU count from machine type: {machine_type}")
+
+
+def create_gcp_startup_script() -> str:
+    """Create the startup script for GCP VM."""
+    script_content = """#!/bin/bash
```
Pbarbosa92, line 21: Can't we have this in a txt file instead of pasted in the code?

inductiva/_cli/cmd_task_runner/launch.py
```diff
@@ -35,8 +36,44 @@ def join_container_streams(*containers, fout: TextIO = sys.stdout):
...
+    success, error_msg = byoc_gcp.create_gcp_vm(vm_name=args.machine_group_name,
+                                                zone=args.zone,
+                                                machine_type=args.machine_type,
+                                                api_key=api_key,
+                                                api_url=api_url,
+                                                spot=args.spot,
+                                                hostname=args.hostname,
+                                                verbose=True)
+
+    if not success:
```
Pbarbosa92, line 55: Since this is 100% related with the byoc why don't we do this error parsing and prints inside the create_gcp_vm?

Or of we are going to do it here at least we should avoid the if's comparing strings because some day in the future we might change the text and this could all break.

Pbarbosa92 (changes requested):

inductiva/_cli/cmd_task_runner/launch.py
```diff
@@ -35,8 +36,28 @@ def join_container_streams(*containers, fout: TextIO = sys.stdout):
...
 
+def launch_task_runner_gcp(args):
+    """Launches a Task-Runner on GCP."""
+    machine_groups.MachineGroup(
+        machine_type=args.machine_type,
+        zone=args.zone,
+        provider="GCP",
+        byoc=True,
+        spot=args.spot,
+        machine_group_name=args.machine_group_name,
```
bgregoinductiva, line 47: This could be simply `name`.

inductiva/_cli/cmd_task_runner/launch.py
```diff
@@ -35,8 +36,28 @@ def join_container_streams(*containers, fout: TextIO = sys.stdout):
         thread.join()
 
 
+def launch_task_runner_gcp(args):
+    """Launches a Task-Runner on GCP."""
```
bgregoinductiva, line 40: Please document the restriction that this type of MG can only have a single VM.

Pbarbosa92 (approved):

bgregoinductiva (approved):

PR #2672, opened 2025-09-30
Add no examples to auth login

Pbarbosa92:
This PR adds the flag `--no-examples` to the auth login to avoid downloading the examples.


rcvalerio (approved):
Should we also add this flag to the docs?

Pbarbosa92:
> Should we also add this flag to the docs?

Yes, but aren't the docs related with the CLI automatic ?

PR #2678, opened 2025-10-03
added-task-fee

eunmoliveira:
The added fee in the openfoam docs was done on another PR

### DIFF
--- a/docs-sources/amr-wind/quick-start.md
+++ b/docs-sources/amr-wind/quick-start.md
@@ -66,7 +66,10 @@ Data:
 	Size of unzipped output:  52.24 MB
 	Number of output files:   92
 
-Estimated computation cost (US$): 0.000063 US$
+Estimated Task Compute Cost = 0.000063 US$ 
+Task Orchestration Fee = 0.01 US$
+Total Estimated Cost = 0.010063 US$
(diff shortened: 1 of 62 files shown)


eunmoliveira:
> I can only see changes showing the format for when the fee is in effect. How about the format for before the fee is in effect, as we defined - [link](https://docs.google.com/document/d/1yGpElon1psvvKX-gvdYAXnOb02mDjZCQCJtbdavrQkg/edit?tab=t.0#heading=h.fnwxipda22zg)

As I understood it, we defined that for the CLI itself, not for these print task summary blocks in the documentation.

eunmoliveira:
> It's true that I didn't include in the document the detailed description of how to show the guides before and after the fee is in effect, but it includes the [implementation phases](https://docs.google.com/document/d/1yGpElon1psvvKX-gvdYAXnOb02mDjZCQCJtbdavrQkg/edit?tab=t.0#heading=h.5ah37f69wjvu), and we discussed in the meeting that this warning in advance would be a better communication option for the registered users. Go ahead and do what you think is best for the users of this content

I don't think an intermediate message is particularly relevant here, especially since the same information will be provided in the documentation, CLI and Console.

Pbarbosa92 (approved):

PR #2680, opened 2025-10-06
BYOC GCP docs

rcvalerio:


docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp.md
```diff
@@ -0,0 +1,274 @@
...
+For more detailed information about the task-runner, see the [Task-Runner Guide](../use-local-task-runner/index.md).
+
+
+## Prerequisites
+
+Before you begin, ensure you have:
+
+- **A Google Cloud Platform (GCP) account** with a valid billing account attached
+- **A GCP project** with the Compute Engine API enabled
+- **Sufficient permissions** in your GCP project (Compute Instance Admin role recommended)
```
sarmento, line 45: Let's try to find what is the minimal permission we need.
We want to minimize the risk for Inductiva, in case we do something wrong on our side. We want to make sure that there are things that we can never do. Please research this as a next step. Can you please help @bgregoinductiva @luispcunha @HenriquePreto ?(also @Pbarbosa92  can you follow this investigation)

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp.md
```diff
@@ -0,0 +1,274 @@
...
+- **Adequate quotas** for the machine types you plan to use
```
sarmento, line 46: Should we have a note of this? Should we make some recommendations also to make sure we are not liable for anything?

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp.md
```diff
@@ -0,0 +1,274 @@
...
+Make sure the Compute Engine API is enabled in your GCP project:
+
+```bash
+# Enable Compute Engine API (required for launching VMs)
+gcloud services enable compute.googleapis.com
+```
+
+You can also enable this through the [GCP Console](https://console.cloud.google.com/apis/library).
+
+## Python Client Usage
```
sarmento, line 108: I think we are missing something in between, like "Testing it all works together".
Also, don't assume that, at this point users know they need to install Inductiva's Python client.
For example, this document may need to be read by a sys admin of a company that wants to understand what is at stake before allowing some user to actually try this.

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Machine Group Creation
+
+Create a machine group on your own GCP account using the `byoc=True` parameter:
```
sarmento, line 112: Maybe you want to say it's just like running a machine directly from Inductiva with just one minor change.

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp.md
```diff
@@ -0,0 +1,274 @@
...
+    machine_type="c2d-highcpu-8",  # GCP machine type to use
+    spot=True,                     # Use spot instances for cost savings
+    mg_name="byoc-gcp-machine",    # Name for your machine group
+    byoc=True,                     # Enable Bring Your Own Cloud
+)
+
+# Start the machine group
+mg.start()
+```
+
```
sarmento, line 129: PLease explain what is different from a normal MG starting procedure.

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp.md
```diff
@@ -0,0 +1,274 @@
...
+### Configuring Auto-Termination
+
+By default, GCP VMs in BYOC machine groups will automatically terminate after **3 minutes of idle time** to help control costs. You can configure this behavior using the `max_idle_time` parameter:
```
sarmento, line 132: Explain why we do this: as a measure for reducing the risk of leaving a machine up.

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp.md
```diff
@@ -0,0 +1,274 @@
...
+# Set custom idle time (in minutes)
+mg = inductiva.resources.MachineGroup(
+    provider="GCP",
+    machine_type="c2d-highcpu-8",
+    spot=True,
+    mg_name="byoc-gcp-machine",
+    byoc=True,
+    max_idle_time=10  # Terminate after 10 minutes of idle time
+)
+
```
sarmento, line 146: Explain how this is different from how we do it now with non-byoc MGs. This is important for users to understand the potential limitations. Also, explain if there is another GCP-based method to prevent users forget a machine during the weekend. If not, then this is an amazing feature we are providing we should give a bit more flexibility.

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp.md
```diff
@@ -0,0 +1,274 @@
...
+task = simulator.run(input_dir=input_dir, on=mg)
+```
+
+
+## Current Limitations (Alpha Version)
+
+Please be aware of the following limitations in the current alpha release:
+
+### Machine Scaling
+- **Single machine per group**: BYOC machine groups only support one VM per group (unlike regular machine groups where you can specify `num_vms`)
```
sarmento, line 216: I don't think it is num_vms

sarmento (approved):
I think this great! 
Left a few comments.
Also, I would break this in smaller parts so we can actually consolidate everything under a "tutorial": 

https://inductiva.ai/guides/expand/bring-your-own-cloud/index

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp.md
rcvalerio: Nice catch, it's `num_machines`. `num_vms` is what we use internally

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp/sections/section2.md
```diff
@@ -0,0 +1,111 @@
...
+    machine_type="c2d-highcpu-8",
+    spot=True,
+    mg_name="byoc-gcp-machine",
+    byoc=True,
+    max_idle_time=None  # No auto-termination
+)
+```
+
+> **⚠️ Warning**: Disabling auto-termination (`max_idle_time=None`) removes the safety mechanism that prevents unexpected charges. You are fully responsible for manually terminating machines. Consider setting up GCP billing alerts and monitoring to track costs if you disable this feature.
+
```
sarmento, line 59: Maybe also explain what Inductiva will know after this call, and how the user can confirm (using the console): we will know which machine was setup, the time of start, etc.. and we will be able to list that info from the machines page.

docs-sources/expand/bring-your-own-cloud/launch-machines-on-gcp/sections/section4.md
```diff
@@ -0,0 +1,58 @@
...
+## Current Limitations (Alpha Version)
+
+Please be aware of the following limitations in the current alpha release:
+
+### Machine Scaling
+- **Single machine per group**: BYOC machine groups only support one VM per group (unlike regular machine groups where you can specify `num_machines`)
+- **Workaround**: Create multiple machine groups in a loop to scale up (see [Python Client Usage](section2.md#machine-scaling-workaround))
+
+### Cost Calculation
+- **Cost reporting**: Currently always reports 0 in cost calculations
```
sarmento, line 12: @luispcunha: there is going to be a special task orquestration fee here, computed at the END of the task. Please talk with @ClaudiaRSilva

